PR #97, opened 2023-05-22
Setting typescript to version 4.9.5 and adding npm packages

craigbreaden:


.devcontainer/Dockerfile
```diff
@@ -5,8 +5,14 @@ RUN apt-get update && apt-get install -y rustc cargo nodejs npm python3-sphinx p
 # Install required Sphinx extensions for the documentation
 RUN pip3 install furo sphinx-inline-tabs
 
+# Install required Sphinx extensions for the documentation
```
garretfick, line 8: I think there is a duplicated line here.

package.json
```diff
@@ -0,0 +1,10 @@
+{
```
garretfick, line 1: Do you know why this is showing in the root of the project? I think this should only be in `integrations/vscode` or perhas the `.devcontainer` folder.

garretfick (changes requested):

.devcontainer/Dockerfile
```diff
@@ -5,8 +5,14 @@ RUN apt-get update && apt-get install -y rustc cargo nodejs npm python3-sphinx p
...
+RUN pip3 install furo sphinx-inline-tabs
+
 # Install just, a task runner
-RUN cargo install --root /usr/local just cargo-release llvm-conv
+RUN cargo install --root /usr/local just cargo-release 
+
+# Install llmv-conv
+RUN cargo install cargo-llvm-cov --locked
```
garretfick, line 15: Locked is not necessary here. I had the wrong name, but there is no cargo.toml file so it won't find the version in any case.

garretfick (changes requested):

.devcontainer/Dockerfile
craigbreaden: Resolved.

package.json
craigbreaden: During my debugging I had to manually install these packages due to missing deps; not sure how it got to the root folder but I moved it into integrations/vscode and re-tested. Looks to be working.

.devcontainer/Dockerfile
craigbreaden: Removed; blindly following online instructions to get round an error during the build. could probably have moved it back to line 12.

craigbreaden: In fact, I'll do that now.

garretfick (approved):

Issue #255, opened 2024-06-27
Error on check of structured text file with use of enum members


bitkeeper:
**Describe the bug**
When using a enum type member it isn't an allowed value.

```ST
TYPE
    MyColors: (Red, Green);
END_TYPE

FUNCTION_BLOCK FB_EXAMPLE
    VAR
        Color: MyColors := Red;
    END_VAR
    Color := Green; (* This isn't allowed; undefined variable *)
END_FUNCTION_BLOCK
```

**To Reproduce**

1. Save the above example as test.st:
2. Run:
`ironplc check ./test.st`
3. Output:
```
error[P0015]: Variable not defined before used (variable=Green)
   ┌─ \\?\C:\data\work\staruml\statemachinegenerator\out\test.st:10:14
   │
10 │     Color := Green;
   │              ^^^^^ Undefined variable

Error: "Error during analysis"
```
**Expected behavior**
That the check passes like:
```
OK
```
**System**
 - Version: 0.66.0
 - OS: Windows 11
 - Distribution: Installer

garretfick:
Thank you for reporting the issue.

The issue does look genuine given the data provided although I haven't yet tried to reproduce the issue. I will look into the cause and if it is genuine, I'll fix in an upcoming release. Releases normally happen every Monday.

Lastly, a word of caution. IronPLC's semantic analysis is nascent. You are likely to find additional issues. Further reports are very much appreciated.
________________________________
From: Bitlab ***@***.***>
Sent: Thursday, June 27, 2024 10:30:22 AM
To: ironplc/ironplc ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [ironplc/ironplc] Error on check of structured text file with use of enum members (Issue #255)


Describe the bug
When using a enum type member it isn't an allowed value.

TYPE
    MyColors: (Red, Green);
END_TYPE

FUNCTION_BLOCK FB_EXAMPLE
    VAR
        Color: MyColors := Red;
    END_VAR
    Color := Green; (* This isn't allowed; undefined variable *)
END_FUNCTION_BLOCK

To Reproduce
Steps to reproduce the behavior:

  1.  Save the above example as test.st:
  2.  Run:
ironplc check ./test.st
  3.  Output:

error[P0015]: Variable not defined before used (variable=Green)
   ┌─ \\?\C:\data\work\staruml\statemachinegenerator\out\test.st:10:14
   │
10 │     Color := Green;
   │              ^^^^^ Undefined variable

[...]

PR #256, opened 2024-07-03
Parse ambiguous enum and vars into a temporary structure

garretfick:


garretfick:
Addresses #255 (at least temporarily). Not a long term solution.

PR #885, opened 2026-04-12
Add implementation plan for MCP server scaffold and list_options tool

garretfick:
First incremental slice of the MCP server design: scaffold the ironplc-mcp
crate as a runnable MCP server with stdio transport, then implement the
list_options tool that returns dialect presets and feature flags.

https://claude.ai/code/session_01NyY3vt6KEiRsf3pRWUATef

### DIFF
--- a/compiler/mcp/src/tools/list_options.rs
+++ b/compiler/mcp/src/tools/list_options.rs
@@ -46,19 +46,7 @@ pub fn build_response() -> ListOptionsResponse {
         })
         .collect();
 
-    let mut flags = Vec::with_capacity(15);
-
-    // The special non-vendor flag (not in FEATURE_DESCRIPTORS but a real
-    // CompilerOptions field that callers can override).
-    flags.push(FlagInfo {
-        id: "allow_iec_61131_3_2013".into(),
-        flag_type: "bool".into(),
-        default: serde_json::Value::Bool(false),
-        description: "Recognise IEC 61131-3:2013 keywords (LTIME, LDATE, LTOD, LDT, REF_TO, \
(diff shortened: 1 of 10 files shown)


compiler/mcp/src/tools/list_options.rs
```diff
@@ -0,0 +1,159 @@
...
+            display_name: d.display_name().to_string(),
+            description: d.description().to_string(),
+        })
+        .collect();
+
+    let mut flags = Vec::with_capacity(15);
+
+    // The special non-vendor flag (not in FEATURE_DESCRIPTORS but a real
+    // CompilerOptions field that callers can override).
+    flags.push(FlagInfo {
```
garretfick: What if we didn't add this flag and relied entirely on dialect? I don't like how the CLI and MCP are diverging here.

compiler/parser/src/options.rs
```diff
@@ -31,6 +31,28 @@ impl Dialect {
...
+
+    /// A one-line description of what this dialect enables.
+    pub fn description(&self) -> &'static str {
+        match self {
+            Dialect::Iec61131_3Ed2 => {
+                "Strict IEC 61131-3:2003 (Edition 2). No vendor extensions. [default]"
+            }
+            Dialect::Iec61131_3Ed3 => "Strict IEC 61131-3:2013 (Edition 3). No vendor extensions.",
+            Dialect::Rusty => {
+                "RuSTy-compatible: Edition 2 base with REF_TO and all vendor extensions."
```
garretfick, line 52: Let's not say "all vendor extensions". That is temporary and won't be true forever

PR #892, opened 2026-04-12
Implement parse and check MCP tools with shared validation

garretfick:
## Summary
This PR implements the `parse` and `check` MCP tools for IronPLC, along with shared validation and diagnostic handling infrastructure. These tools enable syntax checking and semantic analysis via the Model Context Protocol.

## Key Changes

- **New `tools/common.rs` module**: Provides shared types and utilities for all source-accepting tools:
  - `SourceInput` and `ParseCheckInput` types for tool inputs
  - `McpDiagnostic` and `StructureEntry` types for tool outputs
  - `validate_sources()` function enforcing REQ-STL-004 constraints (non-empty names ≤256 bytes, no NUL/`/`/`\`, no duplicates)
  - `parse_options()` function for parsing and validating compiler options with dialect and feature-flag support
  - `map_diagnostic()` and `map_diagnostics()` functions converting compiler diagnostics to MCP format with 1-indexed line/column numbers
  - Comprehensive test coverage for all validation and conversion logic

- **New `tools/parse.rs` module**: Implements the `parse` tool (syntax checking only):
  - `ParseResponse` with `ok`, `structure`, and `diagnostics` fields
  - `build_response()` function that validates inputs, parses each source file, and extracts top-level declarations
  - Best-effort `structure` array containing programs, functions, function blocks, types, and configurations
  - Proper error handling with validation diagnostics returned before parsing

- **New `tools/check.rs` module**: Implements the `check` tool (parse + semantic analysis):
  - `CheckResponse` with `ok` and `diagnostics` fields
  - `build_response()` function using `MemoryBackedProject` for in-memory compilation
  - Full semantic analysis via `project.semantic()` call
  - Consistent diagnostic format with `parse` tool

- **Updated `server.rs`**: Added MCP tool handlers for `parse` and `check` tools with proper parameter binding and response serialization

- **Updated `tools/mod.rs`**: Exported new `parse`, `check`, and `common` modules

- **Updated `Cargo.toml`**: Added dependencies on `ironplc-dsl` and `ironplc-project` crates

## Notable Implementation Details

- Source validation happens before parsing/checking to provide clear validation errors
- Options parsing supports both dialect selection and per-dialect feature-flag overrides
- Byte-to-line/column conversion handles Unicode correctly (counts scalar values, not bytes)
- Diagnostic mapping preserves file context and uses 1-indexed line/column numbers per REQ-TOL-023
- Both tools return `ok: false` on any error (parse, semantic, or validation) but never throw MCP-level errors
- The `parse` tool does not run semantic analysis, allowing it to succeed on programs with type errors
- Comprehensive test coverage validates all requirements including edge cases (empty names, long names, special characters, duplicates, unknown dialects, etc.)

https://claude.ai/code/session_01HYHeRcVHDPDZUTNCtVrAzy


compiler/mcp/src/tools/common.rs
```diff
@@ -0,0 +1,503 @@
...
+            errors.push(validation_error("Source name must not be empty."));
+            continue;
+        }
+        if src.name.len() > 256 {
+            errors.push(validation_error(&format!(
+                "Source name '{}' exceeds 256-byte limit.",
+                truncate(&src.name, 40)
+            )));
+        }
+        if src.name.contains('\0') {
```
garretfick: Instead of rejecting specific characters, can we define what's allowed? We are excluding null but not bell and other special characters.

compiler/mcp/src/tools/common.rs
```diff
@@ -0,0 +1,503 @@
...
+    pub end_line: u32,
+}
+
+// ───────────────────────────────────────────────────────────────────
+// Source-name validation (REQ-STL-004)
+// ───────────────────────────────────────────────────────────────────
+
+/// Validates the `sources` array. Returns validation-error diagnostics
+/// (empty on success).
+pub fn validate_sources(sources: &[SourceInput]) -> Vec<McpDiagnostic> {
```
garretfick: These diagnostics would not have an associated P-code so the don't end up getting documentation. I think these should have a P-code so that we write appropriate documentation for the issues. I'd like to have a common error  that is for validation errors on the inputs - essentially malformed JSON.

compiler/mcp/src/tools/common.rs
```diff
@@ -0,0 +1,503 @@
...
+/// Converts a compiler `Diagnostic` to an `McpDiagnostic`.
+///
+/// `source_map` maps `FileId` to the source text used for byte→line/col
+/// conversion. When the file ID is not found, line 1 / col 1 is used.
+pub fn map_diagnostic(diag: &Diagnostic, source_map: &HashMap<FileId, &str>) -> McpDiagnostic {
+    let file_id = &diag.primary.file_id;
+    let file_name = file_id.to_string();
+    let source = source_map.get(file_id).copied();
+
+    let (start_line, start_col) = byte_offset_to_line_col(source, diag.primary.location.start);
```
garretfick: Why don't we just report the byte offset? Would that be easier for an agent to work with?

compiler/mcp/src/tools/common.rs
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+    (line, col)
+}
+
+// ───────────────────────────────────────────────────────────────────
+// Helpers
+// ───────────────────────────────────────────────────────────────────
+
+/// Creates a validation-level MCP diagnostic (no file / position).
```
garretfick: This assumption that P-codes must have a file/position is wrong. I really don't think we need to introduce a new diagnostic concept. If we really do, then I would rather refactor the existing diagnostics than create a new concept.

compiler/mcp/src/tools/parse.rs
```diff
@@ -0,0 +1,323 @@
...
+        .map(|s| (FileId::from_string(&s.name), s.content.as_str()))
+        .collect();
+
+    for src in sources {
+        let file_id = FileId::from_string(&src.name);
+        match parse_program(&src.content, &file_id, &options) {
+            Ok(library) => {
+                for element in &library.elements {
+                    if let Some(entry) =
+                        element_to_structure_entry(element, &src.name, &src.content)
```
garretfick: Can we get this from the type environment? I'd rather not be iterating over the AST/DSL here. The MCP should not need to know details about the DSL so that we can add new DSLs without changing the MCP server

compiler/mcp/src/tools/parse.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    ParseResponse {
+        ok,
+        structure,
+        diagnostics,
+    }
+}
+
+/// Converts a `LibraryElementKind` into a `StructureEntry`.
+fn element_to_structure_entry(
```
garretfick: See above - I want this to come from the type environment. If that doesn't work now, we can add something into the parse crate the answer the question.

garretfick (commented):
See comments

PR #920, opened 2026-04-16
Add agent triage pipeline for compatibility-gap issues

garretfick:
When a maintainer adds the `status/triage` label to an issue, a GitHub
Actions workflow runs a Python script that validates the issue has enough
information (ST program, actual/expected behavior) and generates a
structured requirements document posted as a comment.

Components:
- Issue template for compatibility-gap reports
- GitHub Actions workflow triggered by label addition
- Python script using Claude to validate and generate requirements
- Unit tests (14 passing, all network calls mocked)

https://claude.ai/code/session_011XRW3ZGJj31CjRLGaneQBA


agents/compatibility_resolver/test_requirements.py
```diff
@@ -0,0 +1,294 @@
+"""Unit tests for agents/compatibility_resolver/requirements.py.
```
garretfick, line 1: These tests should run as part of the CI build so that we don't break them. This isn't happening. I think we need a new "partial" workflow for the "agents" folder that runs those tests. Those should following the same justfile recipe pattern we have for other parts of the code.

garretfick: Added in d2c4b2d:

- `agents/justfile` with `ci` (aliased from `default`), `setup`, and `test` recipes — follows the same pattern as `compiler/justfile` and `docs/justfile`
- `.github/workflows/partial_agents.yaml` — reusable partial workflow (supports both `workflow_dispatch` and `workflow_call`) matching the existing partial workflow pattern
- Wired into `integration.yaml` so agent tests run on every PR, with the same collaborator-association guard as the other jobs

---

PR #928, opened 2026-04-17
Add MCP `symbols` tool for full symbol table extraction

garretfick:
Extends SymbolInfo with variable_type and address fields so the MCP
server can report variable direction and hardware addresses without
walking the AST. Adds insert_variable() to SymbolEnvironment and
updates visit_var_decl to populate the new fields, including handling
direct variables (previously a TODO).

The symbols tool reads from SymbolEnvironment, FunctionEnvironment,
and TypeEnvironment to build the response. Supports optional pou
filter (REQ-TOL-054) and 256 KiB response cap (REQ-TOL-055).

https://claude.ai/code/session_01PkWHgrG9bLPhPKTLa7Jqwb


compiler/mcp/src/tools/symbols.rs
```diff
@@ -0,0 +1,566 @@
...
+            ProgramSymbol {
+                name: name.to_string(),
+                variables,
+            }
+        })
+        .collect()
+}
+
+fn extract_functions(context: &SemanticContext) -> Vec<FunctionSymbol> {
+    context
```
garretfick, line 173: Can this filter logic move into the symbol environment? Here and throughout? That would let the lsp server also use this.

garretfick (commented):
Lets move some logic

PR #930, opened 2026-04-17
Add Linux/macOS install script hosted at ironplc.com

garretfick:
Introduces a one-liner installer matching the rustup/deno/bun pattern so
Linux users (and AI agents on Linux) can install IronPLC without a manual
tarball download:

    curl -fsSL https://www.ironplc.com/install.sh | sh

The script lives at compiler/install.sh and is published to the docs site
via docs/justfile. Because publish-website now depends on a new install
script smoke test, a broken installer cannot reach users — it blocks the
site publish first. Matches the model used by the existing Windows smoke
test.

- compiler/install.sh: POSIX sh, shellcheck clean; detects platform,
  verifies SHA-256, installs to $HOME/.ironplc/bin, idempotent PATH
  setup across bash/zsh/fish. ironplcvm and ironplcmcp are optional
  so the script keeps working against older releases.
- .github/workflows/partial_install_script_test.yaml: reusable smoke
  test matrix over ubuntu-latest and macos-latest, takes a release
  version as input (empty = latest).
- deployment.yaml: runs the smoke test after release; publish-website
  now depends on it.
- integration.yaml: PR CI gains a shellcheck lint job and a smoke job
  against the last published release.
- docs: Linux tab gets curl|sh instructions; macOS tab adds it as an
  alternative to Homebrew; README quick-start leads with the one-liner.


.github/workflows/integration.yaml
```diff
@@ -55,3 +55,25 @@ jobs:
     with:
       commit-ref: ""
       publish: false
+
+  install-script-lint:
+    name: Lint Install Script
+    if: >-
+      github.event_name != 'pull_request' ||
+      contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.pull_request.author_association)
```
garretfick, line 63: Why are we adding this check?

garretfick (commented):
See comment

PR #946, opened 2026-04-21
Add issue-resolver webhook orchestrator + infrastructure

garretfick:
Replaces the GitHub-Actions batch script `agents/compatibility_resolver/`
with a long-running FastAPI orchestrator (`agents/issue-resolver/`) that
drives compatibility-gap issues through
Requirements → Design → Plan → Code → PR stages. Only the Requirements
stage is wired up in this commit; later stages are stubbed in the
orchestrator router.

Key pieces:
- FastAPI `/webhook` with HMAC signature verification and background
  dispatch so event delivery returns 200 immediately.
- State machine keyed by GitHub labels plus a SQLite append-only event
  ledger (prompt hashes only — never raw prompts).
- RequirementsAgent does two LLM calls (validate → generate) so the
  "is this issue complete?" decision is AI-driven rather than a header
  parser. Requirement IDs use the `REQ-TBD-NNN` placeholder prefix so
  the future Design stage can reassign them.
- Terraform in `infrastructure/` provisions 14 workflow labels and the
  repository webhook.
- 24 unit tests cover signature verification, context packaging,
  ledger roundtrip, and orchestrator routing (happy path,
  incomplete issue, agent error, revision-limit cap).

The condensed implementation plan lives at
`specs/plans/2026-04-20-agent-orchestrator.md`.

https://claude.ai/code/session_01RQPZJSgxtFhPt3K4622ydk


agents/issue-resolver/agents/requirements.py
```diff
@@ -0,0 +1,208 @@
...
+Respond with a JSON object and nothing else:
+{"sufficient": true}
+or
+{"sufficient": false, "missing": "Brief explanation of what is missing"}
+"""
+
+
+_REQUIREMENTS_SYSTEM = """\
+You are a requirements engineer for the IronPLC compiler project. IronPLC
+is an open-source IEC 61131-3 PLC runtime. Given a compatibility-gap
```
garretfick: Not just a compatibility gap - can be any kind of issue.

agents/issue-resolver/agents/requirements.py
```diff
@@ -0,0 +1,208 @@
...
+`REQ-TBD-` prefix is a placeholder — the Design stage will reassign it
+to match the target design document (for example REQ-CF-, REQ-TH-, or
+REQ-CG-). Do not invent a different prefix.
+
+Guidelines:
+- Use SHALL for mandatory behavior. Keep each requirement specific,
+  testable, and traceable to what the issue describes.
+- Stay inside the scope of the reported gap. Do not expand into
+  unrelated features.
+- Ground requirements in IEC 61131-3 clauses or tables when the issue
```
garretfick: Make sure you reference the version of the IEC 61131-3 specification where appropriate.

agents/issue-resolver/agents/requirements.py
```diff
@@ -0,0 +1,208 @@
...
+  cites one; otherwise leave spec references out.
+- Where existing compiler problem codes (P####) are relevant, you may
```
garretfick: This is not generally true. We should omit this and follow the guidelines in the repo.

agents/issue-resolver/agents/requirements.py
```diff
@@ -0,0 +1,208 @@
...
+
+Then a blank line, then the first requirement. No other preamble.
+"""
+
+
+@dataclass
+class RequirementsAgent:
+    client: anthropic.Anthropic
+    ledger: Ledger
+    model: str = "claude-sonnet-4-20250514"
```
garretfick: Why this model?

agents/issue-resolver/agents/requirements.py
```diff
@@ -0,0 +1,208 @@
...
+
+    PLACEHOLDER_REQ_PREFIX = "REQ-TBD"
+    STAGE = "requirements"
+
+    async def run(
+        self, context: dict[str, Any], _work_item: Any = None
```
garretfick: Can we use strongly typed instead of `Any`?

agents/issue-resolver/agents/requirements.py
```diff
@@ -0,0 +1,208 @@
...
+                max_tokens=500,
+                system=_VALIDATION_SYSTEM,
+                messages=[{"role": "user", "content": user_prompt}],
+            )
+        except anthropic.AnthropicError as exc:
+            raise AgentError(self.STAGE, f"validation failed: {exc}") from exc
+        except Exception as exc:  # network, unexpected
+            raise AgentError(self.STAGE, f"validation failed: {exc}") from exc
+
+        text = resp.content[0].text if resp.content else ""
```
garretfick: I think we want to use JSON validation of the output. This should be part of the API we are using. I suggest we use Pydantic.

agents/issue-resolver/agents/requirements.py
```diff
@@ -0,0 +1,208 @@
...
+            input_tokens=getattr(resp.usage, "input_tokens", 0),
+            output_tokens=getattr(resp.usage, "output_tokens", 0),
+            phase="validate",
+        )
+        return _parse_validation(text)
+
+    def _generate(self, context: dict[str, Any]) -> str:
+        user_prompt = _format_user_prompt(context)
+        try:
+            resp = self.client.messages.create(
```
garretfick, line 200: Same here - we should be validating the output fits the schema. define the object type upfront.

agents/issue-resolver/agents/requirements.py
```diff
@@ -0,0 +1,208 @@
...
+            prompt=user_prompt,
+            model=self.model,
+            input_tokens=getattr(resp.usage, "input_tokens", 0),
+            output_tokens=getattr(resp.usage, "output_tokens", 0),
+            phase="generate",
+        )
+        return text.strip()
+
+
+def _format_user_prompt(context: dict[str, Any]) -> str:
```
garretfick: I would like to replace `Any` with an object. What would that take to parse this upfront so that formatting doesn't need to worry here?

agents/issue-resolver/context.py
```diff
@@ -0,0 +1,65 @@
...
+
+from typing import Any
+
+
+def build_requirements_context(
+    issue: dict[str, Any],
+    comments: list[dict[str, Any]],
+    *,
+    bot_login: str | None = None,
+) -> dict[str, Any]:
```
garretfick: I think this is where we can return a known object instead of loose dictionary. We are parsing the items here.

PR #955, opened 2026-04-22
Show line and column in playground diagnostics

garretfick:
Replace the raw byte-offset display in playground diagnostics with
1-based line and column numbers that match the editor's line number
gutter.

Add a single `offset_to_line_column` helper in `ironplc_dsl::diagnostic`
and use it from both the LSP server (`span_to_range`, `map_label`) and
the playground WASM entry point so one routine produces line/column
for every surface.

https://claude.ai/code/session_01LvCy4YV73e2phbs56VD18B


compiler/dsl/src/diagnostic.rs
```diff
@@ -28,6 +28,37 @@ pub struct Location {
...
+    pub column: u32,
+}
+
+/// Converts a byte offset in `source` to a 0-indexed line and column.
+///
+/// The column is measured in code points, matching LSP's default
+/// `PositionEncodingKind::Utf16` close enough for single-codepoint
+/// characters and mirroring the convention used elsewhere in the compiler.
+///
+/// Offsets beyond the end of `source` are clamped to the end of `source`.
```
garretfick: Should this be part of an impl block?

compiler/playground/src/lib.rs
```diff
@@ -90,6 +95,28 @@ struct DiagnosticInfo {
     label: String,
     start: usize,
```
garretfick: We don't need the start and end anymore. Let's remove

garretfick (commented):
See comments

PR #966, opened 2026-04-23
MCP run tool (container execution)

garretfick:
Adds the implementation plan for the MCP `run` tool, the final piece
of Milestone 2 (execution surface). MVP ships `every_cycle` trace,
name resolution, and resource limits; stimuli and full value codec
deferred to a follow-up.

https://claude.ai/code/session_017t3qvYa2aJbfHtGP53Ugk1


compiler/mcp/src/tools/run.rs
```diff
@@ -0,0 +1,815 @@
...
+                "Trace sample cap reached; emitted trace is truncated.".to_string()
+            }
+            TerminatedReason::Error => outcome
+                .error_message
+                .clone()
+                .unwrap_or_else(|| "VM trap during execution.".to_string()),
+            TerminatedReason::Completed => unreachable!(),
+        };
+        diagnostics.push(serialize_diagnostic(&Diagnostic::problem(
+            Problem::McpInputValidation,
```
garretfick, line 273: This is not an input validation issue. I think we should not be using diagnostic here and have a custom type that mirrors diagnostic in the shape but works with the VM's "V-codes"

compiler/mcp/src/tools/run.rs
```diff
@@ -0,0 +1,815 @@
...
+    // `max_wall_clock_ms` and must not block other MCP calls).
+    let cached_snapshot = (*cached).clone_for_run();
+    drop(guard);
+
+    // --- Execute ---
+    let outcome = match runner::execute(&cached_snapshot, &trace_set, effective_limits) {
+        Ok(o) => o,
+        Err(msg) => {
+            let diag = Diagnostic::problem(
+                Problem::InternalError,
```
garretfick, line 217: Should this be using a V code?

compiler/mcp/src/tools/run.rs
```diff
@@ -0,0 +1,815 @@
...
+
+/// Top-level response for the `run` tool.
+#[derive(Debug, Serialize)]
+pub struct RunResponse {
+    pub ok: bool,
+    pub trace: Vec<TraceEntry>,
+    pub truncated: bool,
+    pub terminated_reason: String,
+    pub summary: RunSummary,
+    pub diagnostics: Vec<Value>,
```
garretfick, line 95: this looks like it is returning a string for diagnostics. Can we make these structured?

compiler/mcp/src/runner.rs
```diff
@@ -0,0 +1,501 @@
...
+/// 4. On trap: capture diagnostic, drain final values from `VmFaulted`.
+/// 5. On clean stop: drain final values from `VmStopped`.
+pub fn execute(
+    cached: &CachedContainer,
+    trace_set: &[ResolvedVar],
+    limits: EffectiveLimits,
+) -> Result<RunOutcome, String> {
+    let mut bytes = cached.iplc_bytes.as_slice();
+    let container =
+        Container::read_from(&mut bytes).map_err(|e| format!("container read error: {e}"))?;
```
garretfick, line 231: This should return a V-code

garretfick (commented):
Generally, I would like better structured errors, ones that return V-code to indicate what happened with the runtime.

PR #990, opened 2026-04-27
Refactor float clamp and BCD encoding functions using macros

garretfick:
## Summary
This PR refactors repetitive floating-point clamp and binary-coded decimal (BCD) encoding functions by introducing two macros that eliminate code duplication while maintaining identical functionality.

## Key Changes
- **Float clamp refactoring**: Replaced separate `float_clamp_f32` and `float_clamp_f64` functions with a `define_float_clamp!` macro that generates both implementations. The macro preserves the IEEE 754-safe behavior where NaN inputs propagate to the output.

- **BCD encoding refactoring**: Replaced three separate BCD encoding functions (`int_to_bcd_u16`, `int_to_bcd_u32`, `int_to_bcd_u64`) with a `define_int_to_bcd!` macro that generates all three implementations. The macro is parameterized by the integer type, number of digits, and modulus value.

## Implementation Details
- Both macros are defined using Rust's `macro_rules!` and maintain the `#[inline]` attribute for performance-critical code paths
- The float clamp macro accepts a function name and type parameter
- The BCD macro accepts a function name, type, digit count, and modulus to handle the different ranges (u16: 4 digits, u32: 8 digits, u64: 16 digits)
- No functional changes to the actual clamp or BCD encoding logic—this is purely a refactoring to reduce code duplication

https://claude.ai/code/session_01Wwf57pK8hQjuKyKqucfXw5


garretfick:
Rejecting. This change does not improve the code even if it reduces duplication

PR #1003, opened 2026-04-30
Elide TRUNC opcodes in FOR loops with constant, in-range bounds

garretfick:
## Summary

Implements an optimization that eliminates redundant `TRUNC` (truncation) opcodes in FOR loops over narrow integer types (`SINT`, `INT`, `USINT`, `UINT`) when the loop bounds are compile-time constants that provably keep all iteration values within the declared type's range.

## Key Changes

- **Added interval analysis helpers** in `compile_stmt.rs`:
  - `try_constant_i64()`: Extracts constant integer values from expressions, including unary-negated literals
  - `narrow_type_range()`: Returns the min/max range for narrow integer types (8-bit and 16-bit)
  - `for_loop_trunc_can_be_elided()`: Conservative check that determines if both TRUNC sites can be safely removed based on constant bounds analysis

- **Modified FOR loop compilation** in `compile_stmt.rs`:
  - Gated both `emit_truncation()` calls (after init and after increment) on the new elision check
  - Preserves TRUNC when bounds are non-constant or when wrap-around at type boundaries could occur

- **Added comprehensive test coverage**:
  - Bytecode tests in `compile_loops.rs` for elision cases (INT/SINT/UINT with safe bounds, negative step) and non-elision cases (boundary values, non-constant bounds)
  - End-to-end execution tests in `end_to_end_loops.rs` verifying runtime correctness across narrow types
  - Updated benchmark profile test to assert zero TRUNC_I16 in the optimized loop

- **Added specification document** `specs/plans/2026-04-30-elide-for-loop-trunc.md` documenting the soundness condition, approach, and scope

## Implementation Details

The optimization is **conservative and sound**:
- Only elides TRUNC when all three bounds (`from`, `to`, `step`) are compile-time constants
- For positive step: verifies `from` and `to` are in range, and `to + step ≤ T_max`
- For negative step: mirrors the check with `to + step ≥ T_min`
- Rejects zero step and any non-constant bounds
- Preserves existing wrap-around behavior for boundary cases (e.g., `FOR i:INT := 1 TO 32767`)

This eliminates one opcode per loop iteration for typical safe FOR loops, improving performance without changing observable behavior.

https://claude.ai/code/session_01GNZBzmEiEPDAdQPNFUJWzZ

### DIFF
--- a/compiler/codegen/tests/compile_loops.rs
+++ b/compiler/codegen/tests/compile_loops.rs
@@ -1,6 +1,7 @@
 //! Bytecode-level integration tests for WHILE, REPEAT, and FOR loop compilation.
 
 mod common;
+use ironplc_container::opcode;
 use ironplc_parser::options::CompilerOptions;
 
 use common::parse_and_compile;
@@ -210,11 +211,6 @@ END_PROGRAM
 // are constants that keep every visible value (init, body, and the
 // post-final-increment) within the declared narrow type's range.
 
-const TRUNC_I8: u8 = 0x20;
-const TRUNC_U8: u8 = 0x21;
-const TRUNC_I16: u8 = 0x22;
-const TRUNC_U16: u8 = 0x23;
-
 #[test]
 fn compile_when_for_int_with_safe_constant_bounds_then_omits_trunc() {
     // Body uses a DINT sink so the only candidate TRUNC opcodes are the two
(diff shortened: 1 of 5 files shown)


compiler/codegen/tests/compile_loops.rs
```diff
@@ -204,3 +204,194 @@ END_PROGRAM
     // Verify LT_I32 for negative step (instead of GT_I32)
     assert_eq!(bytecode[12], 0x6A); // LT_I32
 }
+
+// FOR-loop TRUNC elision (specs/plans/2026-04-30-elide-for-loop-trunc.md):
+// the per-iteration TRUNC opcode is elided when the control variable's bounds
+// are constants that keep every visible value (init, body, and the
+// post-final-increment) within the declared narrow type's range.
+
+const TRUNC_I8: u8 = 0x20;
```
garretfick: We should not copy these values. Use the constant instead.

garretfick (commented):
See comment

PR #1039, opened 2026-05-03
Display string values in the playground UI

garretfick:


compiler/playground/src/lib.rs
```diff
@@ -171,6 +156,39 @@ fn build_enum_value_map(container: &Container) -> EnumValueMap {
...
+    }
+    let cur_len = u16::from_le_bytes([data_region[off + 2], data_region[off + 3]]) as usize;
+    let start = off + STRING_HEADER_BYTES;
+    let end = start + cur_len;
+    if end > data_region.len() {
+        return "'<invalid>'".into();
+    }
+    let bytes = &data_region[start..end];
+    let text = String::from_utf8_lossy(bytes);
+    format!("'{text}'")
```
garretfick: What will happen if the text contains characters such as newline or null terminating. I think we need to substitute those with IEC defined escape sequences.

compiler/playground/src/lib.rs
```diff
@@ -171,6 +156,39 @@ fn build_enum_value_map(container: &Container) -> EnumValueMap {
...
+            map.insert(entry.var_index.raw(), entry.data_offset);
+        }
+    }
+    map
+}
+
+/// Reads a STRING value from the data region at the given offset and renders
+/// it as a single-quoted IEC literal. Returns `'<invalid>'` if the offset
+/// or length would read out of bounds.
+fn read_string_value(data_region: &[u8], data_offset: u32) -> String {
```
garretfick: Can we make the idea of 'invalid' a return type so that we can display this differently in the UI? Right now, the string invalid and not actually being valid would look the same.

PR #1050, opened 2026-05-05
Implement WSTRING support per ADR-0034 and ADR-0035

garretfick:
## Summary
- Implements the [Complete WSTRING Support plan](specs/plans/2026-05-05-complete-wstring-support.md) end-to-end across all 6 phases
- Bumps container `FORMAT_VERSION` 2 → 3 and `STRING_HEADER_BYTES` 4 → 6 (new `char_width: u16` field per ADR-0035) so STRING and WSTRING share one self-describing layout
- Adds `Trap::EncodingMismatch` (V9014) and runtime encoding-mismatch checks on every string opcode (defense-in-depth per ADR-0034)
- Threads `char_width` through analyzer (`IntermediateType::String`), codegen (`StringVarInfo` / `StringParamInfo` / `StringReturnInfo` / `ArraySpec`), constant pool (per-entry encoding tag, new `ConstType::WStr`), and the VM
- New `end_to_end_wstring.rs` (8 tests) covers WSTRING declarations, UTF-16LE byte verification, mixed STRING+WSTRING coexistence, assignment, `LEN`, and `CONCAT`

## Phase breakdown
1. Container format & VM layout helpers — version bump, header layout, `ConstType::WStr`, helper signatures
2. VM opcodes — STR_INIT takes `char_width`; STORE/LOAD/array variants verify encoding; string functions scale byte offsets by `char_width`
3. Analyzer — `IntermediateType::String { max_len, char_width }`, seeded from `StringDeclaration.width` / `StringInitializer.width`
4. Codegen — width flows from declared type to all info structs and the constant pool; `encode_string_literal` learns UTF-16LE
5. End-to-end tests — `compiler/codegen/tests/it/end_to_end_wstring.rs`
6. Docs / final CI — V9014 problem page, MCP types render `kind: \"wstring\"`, `cd compiler && just` green (compile + coverage @ 85% + clippy + fmt + dupes)

## Test plan
- [x] `cd compiler && just compile`
- [x] `cd compiler && just test` (all crates pass, including the new `end_to_end_wstring` suite)
- [x] `cd compiler && just coverage` (≥ 85%)
- [x] `cd compiler && just lint` (clippy + fmt clean)
- [x] `cd compiler && just dupes` (within thresholds)
- [x] Existing STRING tests still pass after the format-version and header-size bump


compiler/codegen/src/compile_fn.rs
```diff
@@ -230,9 +233,13 @@ pub(crate) fn compile_user_function(
     let return_string_info = match &func_decl.return_type {
         FunctionReturnType::String(spec) | FunctionReturnType::WString(spec) => {
             let max_length = resolve_string_spec_max_length(spec)?;
+            let char_width = match &func_decl.return_type {
+                FunctionReturnType::WString(_) => WIDE_CHAR_WIDTH,
+                _ => NARROW_CHAR_WIDTH,
```
garretfick: Let's match both variants directly so we can identify if we ever add a new enum variant because the build breaks.

compiler/analyzer/src/intermediates/string.rs
```diff
@@ -12,6 +21,7 @@ pub fn from(initializer: &StringInitializer) -> TypeAttributes {
                 .length
                 .as_ref()
                 .and_then(|len| len.as_integer().map(|i| i.value)),
+            char_width: char_width_for(&initializer.width),
```
garretfick: Can this also be an impl?

compiler/analyzer/src/intermediates/string.rs
```diff
@@ -1,8 +1,17 @@
 use crate::{intermediate_type::IntermediateType, type_environment::TypeAttributes};
 
-use ironplc_dsl::common::{StringDeclaration, StringInitializer};
+use ironplc_dsl::common::{StringDeclaration, StringInitializer, StringType};
 use ironplc_dsl::core::Located;
 
+/// Returns the per-code-unit byte width for a [`StringType`]:
+/// 1 for STRING (Latin-1), 2 for WSTRING (UTF-16LE) — per ADR-0016.
+fn char_width_for(width: &StringType) -> u8 {
```
garretfick: Instead of a detached function, can we instead add an impl for StringType?

garretfick:
Addressed all three review comments in 6022b0f3:

- **`compile_fn.rs:238`** (match both return-type variants directly): dropped the inner wildcard match. `char_width` now comes from `spec.width.char_width()`. The outer match on `FunctionReturnType` is exhaustive (`Named` is the third arm), so a future variant fails the build here.
- **`intermediates/string.rs` `char_width_for`**: moved to `impl StringType { fn char_width(&self) -> u8 { ... } }` on the DSL enum so callers use `width.char_width()`. The detached function in the analyzer and the parallel one in codegen (`char_width_for_string_type`) are both gone.

`cd compiler && just` (compile + coverage @ 85% + clippy + fmt + dupes) passes locally.

garretfick:
Follow-up in `0b9d8eb3`: introduce a typed `CharWidth { Narrow = 1, Wide = 2 }` enum and thread it through the container, codegen, and VM in place of the raw `u8` `char_width` tag.

The motivation is that the three runtime tag sites described in ADR-0034 — constant-pool entry, string-header byte, and `STR_INIT`'s bytecode operand — were each a `u8` with two valid values, so the code carried a lot of `as usize` casts and string ops couldn't exhaustively match on the encoding. Boundary reads now go through a single `CharWidth::from_u8` validator that produces:

- `ContainerError::InvalidCharWidth(u8)` at deserialization (constant pool)
- `Trap::InvalidCharWidth(u8)` (V9015) at runtime (header reads, `STR_INIT` operand)

Any byte other than `1` or `2` is rejected at the boundary instead of silently being multiplied into a byte offset. Documented as `V9015` and covered by unit tests in `string_ops`, container `constant_pool`, and the new `char_width` module.

`cd compiler && just` is green (compile + coverage @ 85% + clippy + fmt + dupes).

PR #1067, opened 2026-05-13
Add CODESYS dialect

garretfick:
Adds a `codesys` dialect preset to the parser, CLI, LSP, MCP, and docs.
The dialect uses an Edition 2 base with `REF_TO` and the vendor extensions
that CODESYS accepts (all of the RuSTy set except `allow_system_uptime_global`,
which is an IronPLC runtime convention rather than a CODESYS feature).

Also refactors the dialect plumbing so future dialects only need to touch
`parser/src/options.rs`:

* Implement `clap::ValueEnum` for a `ClapDialect` newtype that re-uses
  `Dialect::ALL` (via a `#[repr(transparent)]` slice cast). Drops the
  parallel `CliDialect` enum and its `to_dialect()` mapping.
* Add `FromStr for Dialect` and route the LSP through `.parse()`,
  replacing the hand-rolled per-variant match.
* Add `Dialect::cli_name() -> &'static str` so the CLI/MCP/LSP all share
  one source of the string form.

https://claude.ai/code/session_01SYnUHBNMawU1yUfTiTRdAM


compiler/ironplc-cli/bin/main.rs
```diff
@@ -25,32 +25,35 @@ struct Args {
...
-            CliDialect::Iec61131_3Ed3 => Dialect::Iec61131_3Ed3,
-            CliDialect::Rusty => Dialect::Rusty,
-        }
+impl clap::ValueEnum for ClapDialect {
+    fn value_variants<'a>() -> &'a [Self] {
+        // SAFETY: `ClapDialect` is `#[repr(transparent)]` over `Dialect`, so
+        // it has identical layout and validity.  A `&[Dialect]` therefore
+        // satisfies the layout requirements of `&[ClapDialect]`.
+        let dialects: &'static [Dialect] = Dialect::ALL;
+        unsafe { &*(dialects as *const [Dialect] as *const [ClapDialect]) }
```
garretfick: We don't allow unsafe in the project. If we have to duplicate code, I'd rather that over a single use of unsafe.

garretfick (commented):
See comment. No unsafe allowed

compiler/ironplc-cli/bin/main.rs
garretfick: Understood. Dropped the `unsafe` slice cast in d08b5c9 — `ClapDialect::value_variants` now returns a hand-written `&[ClapDialect]` that mirrors `Dialect::ALL`, and there is a unit test that compares the two so they can't drift.

---

PR #1069, opened 2026-05-21
Binary-search DebugSection source-map lookup

garretfick:
Sort line_map by (function_id, bytecode_offset) in Builder::build and DebugSection::read_from, and switch lookup_source_location from a linear scan to a binary search. This pre-factors the source-map data structure for the upcoming debugger that calls the lookup once per instruction. Documents the new sort invariant on the struct field.

### DIFF
--- a/compiler/container/src/debug_section.rs
+++ b/compiler/container/src/debug_section.rs
@@ -144,9 +144,9 @@ pub struct DebugSection {
     ///
     /// Invariant: entries are sorted by `(function_id, bytecode_offset)` so
     /// that [`lookup_source_location`](Self::lookup_source_location) can use
-    /// a binary search. [`Self::read_from`] and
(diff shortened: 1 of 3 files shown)


compiler/container/src/debug_section.rs
```diff
@@ -264,6 +276,10 @@ impl DebugSection {
             }
         }
 
+        // Restore the sort invariant in case the source bytes were written
```
garretfick: Why do we need this?

garretfick: Good point — we don't. Tag 1 isn't emitted by production codegen yet and the only in-tree producer is `Builder::build`, which now sorts. There's no shipped version of the format that could have unsorted entries to defend against. Dropped the sort-on-read (and the test that depended on it) in dba69d4, and updated the field doc.

---

PR #1074, opened 2026-05-22
Add debug sub-table tag 6 (source file table); switch hashes to BLAKE3

garretfick:
## Summary

- **Drop `header.source_hash`** and switch the four container hashes (content, debug, layout, per-file) to **BLAKE3** for consistency. `HEADER_SIZE` shrinks 256 → 224; debug/layout hashes shift up. ADR-0007 amended substantively. No SHA-256 code exists today (the header fields were zero-filled placeholders), so this is a spec-and-layout change only.
- **Implement debug sub-table tag 6 (SOURCE_FILE_TABLE)** with `SourceFileEntry { path: String, content_hash: [u8; 32] }`. Per-file BLAKE3 hashes replace the single combined `source_hash` and identify *which* file drifted, not just *that* something drifted.
- **`LineMapEntry` grows 8 → 10 bytes** by adding `file_id: u16` — an index into the new SOURCE_FILE_TABLE. The `(function_id, bytecode_offset)` sort invariant is unchanged, so `lookup_source_location` keeps its O(log N) binary search.
- **Widen `Emitter::set_source_position` to `(file_id, line, column)`** since the API has no external consumers yet — clean rewrite, not a backwards-compat shim. Dedup now keys on the full triple so a file change records a new entry at the same line/column.
- **No codegen integration yet.** Nothing in production calls `set_source_position`, so threading source-bytes lookup through `codegen::compile()` would be plumbing without payload. That wiring (and the BLAKE3 hashing of real source files) belongs in a follow-up PR alongside actual source-map integration.

Plan: `specs/plans/2026-05-22-debug-source-file-table.md` (committed in this PR).

## Test plan

- [x] `cd compiler && just` passes (compile + coverage + clippy + fmt + dupes + smoke)
- [x] Container crate tests: 154 → 160 (+6 new)
  - Round-trip a SOURCE_FILE_TABLE with two entries
  - Round-trip a line map mixing `file_id = 0` and `file_id = 1` per entry
  - Cross-check `SourceFileEntry.content_hash` against `blake3::hash(input)` so any silent algorithm swap fails loudly
  - Forward-compat: a reader with an unknown tag still loads tag 6 that follows
  - Two `ContainerBuilder` tests for insertion-order `file_id` assignment
- [x] Emitter: existing line-map tests updated for the widened API, plus a new test that a file_id change at the same `(line, column)` records a separate entry
- [x] Golden `steel_thread.iplc` regenerated under the new 224-byte header layout; vm-cli tests pass under the new layout

## Notes for review

- The plan is one of the commits — recommend reading it first to ground the rest.
- Spec changes: `specs/design/bytecode-container-format.md` (REQ-CF-001/005/006 updated to 224 bytes; tag 6 layout formalized; LineMapEntry table now 10 bytes), `specs/design/debugger-support.md` (gap #7 closed), `specs/adrs/0007-dual-signature-integrity-model.md` (BLAKE3 throughout; per-file hashes replace the single source hash).
- ADR-0007 is amended, not superseded — the dual-signature *structure* is unchanged; only the algorithm and the source-integrity granularity move.

https://claude.ai/code/session_017jnAenCns8PVjcU4rj66jN


compiler/codegen/src/emit.rs
```diff
@@ -52,9 +56,10 @@ pub struct Emitter {
     /// `Some`, the next opcode that adds bytes to `bytecode` records a
     /// line_map entry at its offset (subject to dedup against the last
     /// recorded position).
-    current_position: Option<(u16, u16)>,
-    /// Recorded `(bytecode_offset, source_line, source_column)` entries in
-    /// emission order — therefore monotonically non-decreasing in offset.
+    current_position: Option<(u16, u16, u16)>,
```
garretfick: Instead of bare u16, can we can we make types for these? That would:

1. make it easier to understand
2. ensure correct usage

compiler/container/src/spec_conformance.rs
```diff
@@ -38,11 +38,11 @@ fn all_spec_requirements_have_tests() {
 // Container Format — File Header (REQ-CF-001 through REQ-CF-007)
 // ---------------------------------------------------------------------------
 
-/// REQ-CF-001: The file header is exactly 256 bytes.
+/// REQ-CF-001: The file header is exactly 224 bytes.
 #[spec_test(REQ_CF_001)]
-fn container_spec_req_cf_001_header_size_is_256_bytes() {
+fn container_spec_req_cf_001_header_size_is_224_bytes() {
```
garretfick: It seems like you changed the length of the header. I think we decided it is 256 bytes with padding as necessary. Why are you changing this?

compiler/container/src/header.rs
```diff
@@ -171,59 +173,55 @@ impl FileHeader {
         let profile = buf[6];
         let flags = buf[7];
 
-        // Region 2: Hashes (bytes 8-135)
+        // Region 2: Hashes (bytes 8-103)
```
garretfick: Why are we changing the position of items in the header?

garretfick (commented):
See comments

compiler/container/src/spec_conformance.rs
garretfick: Reverted in 1e0f6b7. You're right — I shouldn't have shrunk the header. `HEADER_SIZE` is back to 256, the `source_hash` field is still gone from `FileHeader`'s public API, but the 32-byte slot at bytes 40-71 is preserved on disk as `reserved_hash_slot: [u8; 32]` (must be zero). All other fields back at their original offsets. The pre-PR `steel_thread.iplc` was restored from main and loads cleanly under the new reader — that's the backwards-compat check.

---

compiler/container/src/header.rs
garretfick: Reverted in 1e0f6b7 along with the size change. `debug_hash` is back at offset 72, `layout_hash` at 104, section directory at 136, runtime params at 192, trailing reserved at 218. The 32-byte hole left by `source_hash` becomes `reserved_hash_slot` (always zero) so the wire layout is unchanged.

---

compiler/codegen/src/emit.rs
garretfick: Done in 1e0f6b7. Added three newtypes to `container::id_types` alongside `FunctionId`/`VarIndex`/etc.:

- `SourceFileId(u16)` — index into the SOURCE_FILE_TABLE
- `SourceLine(u16)` — 1-based line, `0` = unknown
- `SourceColumn(u16)` — 1-based column, `0` = unknown

Replaced the bare `u16`s on `LineMapEntry`, `EmittedLineMapEntry`, and `Emitter::set_source_position`. Prevents argument-order swaps at call sites and matches the existing newtype convention.

Kept `bytecode_offset: u16` as-is — wrapping that one would touch a lot of jump-patch / label-resolution code in the emitter, which is out of scope for this PR. Happy to do it as a follow-up if you want.

---

PR #1075, opened 2026-05-23
Thread CharWidth through codegen string metadata

garretfick:
## Summary
- Replaces codegen's raw `u16` / `STRING_CHAR_WIDTH = 1` representation with a typed `CharWidth` on `StringVarInfo`, `StringParamInfo`, `StringReturnInfo`, and `ArraySpec`. Adds typed `NARROW_CHAR_WIDTH` / `WIDE_CHAR_WIDTH` constants and a `char_width_for_string_type(&StringType) -> CharWidth` helper.
- Construction sites in `compile_fn.rs`, `compile_setup.rs`, `compile_struct.rs`, and `compile_array.rs` now read the actual encoding from the declaration's `StringType` (or from `IntermediateType::String.char_width` for array elements typed via the analyzer) rather than hard-coding Latin-1.
- `string_region_size` and `encode_string_literal` take `CharWidth`. The encoder still panics on `Wide` — UTF-16LE literal encoding needs the constant-pool wire format to carry a per-entry encoding tag, which lands with the format bump.
- `ArraySpec` gains a `string_char_width: Option<CharWidth>` field next to the existing `string_max_len`, populated from `ArrayElementType::String` (Narrow) vs `WString` (Wide).

No behavioral change for STRING — Narrow yields the same byte layout as the prior `STRING_CHAR_WIDTH = 1`. WSTRING declarations now flow through with `Wide` and the data region grows for the doubled payload, but no test on `main` exercises WSTRING end-to-end. Carved out of #1050 to keep the WSTRING split-up reviewable. Builds on #1070 (CharWidth enum) and #1073 (analyzer width tracking).

## Test plan
- [x] `cd compiler && just compile`
- [x] `cd compiler && just test`
- [x] `cd compiler && just coverage` (≥ 85%)
- [x] `cd compiler && just lint` (clippy + fmt clean)
- [x] `cd compiler && just dupes` (within thresholds)
- [x] Existing STRING tests pass unchanged — STRING declarations resolve to `CharWidth::Narrow`, byte-identical to the prior `STRING_CHAR_WIDTH = 1`

---
_Generated by [Claude Code](https://claude.ai/code/session_01BG78PFCpLogkubyscPWEiQ)_

### DIFF
--- a/compiler/codegen/src/compile.rs
+++ b/compiler/codegen/src/compile.rs
@@ -45,12 +45,12 @@ use ironplc_container::debug_section::{
     EnumDefEntry, FuncNameEntry, StringLayoutEntry, VarNameEntry,
 };
 use ironplc_container::{
-    Container, ContainerBuilder, FbTypeId, FunctionId, UserFbDescriptor, VarIndex,
+    CharWidth, Container, ContainerBuilder, FbTypeId, FunctionId, UserFbDescriptor, VarIndex,
     STRING_HEADER_BYTES,
 };
 use ironplc_dsl::common::{
     FunctionBlockDeclaration, FunctionDeclaration, InitialValueAssignmentKind, Library,
-    LibraryElementKind, ProgramDeclaration, VarDecl, VariableType,
+    LibraryElementKind, ProgramDeclaration, StringType, VarDecl, VariableType,
 };
 use ironplc_dsl::configuration::ConfigurationDeclaration;
 use ironplc_dsl::core::{FileId, Id, Located};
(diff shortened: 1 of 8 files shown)


compiler/codegen/src/compile.rs
```diff
@@ -469,9 +481,13 @@ fn compile_program_with_functions(
...
         if ctx.num_temp_bufs > 0 {
-            builder = builder
-                .num_temp_bufs(ctx.num_temp_bufs)
-                .max_temp_buf_bytes(string_region_size(ctx.max_string_capacity));
+            builder =
+                builder
+                    .num_temp_bufs(ctx.num_temp_bufs)
+                    .max_temp_buf_bytes(string_region_size(
+                        ctx.max_string_capacity,
+                        NARROW_CHAR_WIDTH,
```
garretfick, line 489: How will this work when this is WIDE CHAR?

PR #1078, opened 2026-05-24
Instrument playground with PostHog event capture

garretfick:
## Summary

Adds structured product analytics to the playground so we can answer "which docs page is shipping a broken embed?", "which examples do people run?", and "what kinds of errors trip users up?" — questions the current Clicky goal pings (`embed_compile_error` / `playground_run` etc.) can't answer because they carry no properties.

Clicky pageviews are unchanged. The two `clicky.goal` calls are removed and replaced by PostHog events with rich super-properties. The docs Sphinx extension now passes the docname through the iframe URL so the playground can self-identify which docs page hosted each load.

## Event schema

**Super-properties** (registered on load, attached to every event):

| Property | Type | Notes |
|---|---|---|
| `embed` | bool | True when `?embed=true` |
| `dialect` | `"2003" \| "2013"` | Re-registered on dialect change |
| `allows` | string | Sorted CSV of `--allow-*` flags |
| `program_origin` | `"docs" \| "url_shared" \| "example" \| "user_defined"` | See taxonomy |
| `host_page` | string \| null | Docs page slug; only when origin=docs |
| `example_name` | string \| null | Only when origin=example |
| `program_modified` | bool | Flips true on first keystroke after load |

**Events**: `playground_loaded`, `compile_attempted`, `compile_finished` (with `success`, dedup'd `error_codes[]`, `error_count`, `program_lines`, `duration_ms`), `run_started`, `run_stopped` (with `reason` ∈ `user`/`error`/`reload`, `duration_ms`, `cycle_count`, plus `error_codes[]` when reason=error), `example_loaded`.

**`program_origin` taxonomy**:

| Load condition | `program_origin` | `host_page` |
|---|---|---|
| `?source=ironplc-docs&host=<slug>&code=…` | `docs` | `<slug>` |
| `?code=…` without `source=ironplc-docs` | `url_shared` | null |
| User picks from examples dropdown | `example` | null |
| Anything else (default Counter template, scratch) | `user_defined` | null |

## What changed

- **`playground/index.html`** — PostHog snippet added (US host, `person_profiles: identified_only`, no autocapture / pageview / session recording). Clicky script untouched.
- **`playground/app.js`** — analytics module (`capture`, `registerSuper`, `setProgramOrigin`, `markModified`, `captureRunStopped`, `extractErrorCodes`, `initAnalytics`). Events wired at: Start (compile_attempted, compile_finished both branches, run_started), Stop button (run_stopped/user), step-loop trap (run_stopped/error with diagnostic codes), worker error (run_stopped/error), cycle overrun (run_stopped/error with synthetic `CYCLE_OVERRUN` code), dialect/source/example interrupt (run_stopped/user), `pagehide` (run_stopped/reload), example dropdown (example_loaded), editor input (markModified). `trackGoal` and both call sites removed.
- **`docs/extensions/ironplc_playground.py`** — `host` threaded through `_build_playground_url` / `_build_iframe`; each `Directive.run()` derives it from `self.state.document.settings.env.docname`. URLs gain `source=ironplc-docs&host=<docname>` when called from a directive. Extension version bumped to 0.4.
- **`specs/plans/2026-05-24-playground-analytics-instrumentation.md`** — plan committed first per the project's planning requirement.

## Things explicitly NOT in scope

- No cookie/consent banner (site has none today; separate policy decision).
- No replacement of Clicky pageviews.
- No PII or source code in events; PostHog `distinct_id` stays anonymous.

## Test plan

- [x] `node --check playground/app.js` — parses
- [x] `python3 -m py_compile docs/extensions/ironplc_playground.py` — parses
- [x] `cd docs && sphinx-build -a -W -n -b html . _build` — builds with strict warnings; rendered iframes contain `source=ironplc-docs&host=explanation%2Freferences` (and equivalents for every page using `.. playground::`)
- [x] `python3 -m http.server` on `playground/` — index.html serves both Clicky + PostHog snippets; app.js serves new helpers; zero residual `trackGoal` references
- [x] `cd compiler && just` — compile + coverage (≥85%) + lint (clippy/fmt) + dupes all pass
- [ ] Post-deploy: verify in PostHog that `playground_loaded` lands with expected super-props (standalone → `program_origin=user_defined`; docs iframe → `program_origin=docs, host_page=<docname>`; `?code=...` link → `program_origin=url_shared`)
- [ ] Post-deploy: confirm `compile_finished{success:false}` events appear with non-empty `error_codes` when the program has compile errors
- [ ] Post-deploy: confirm `run_stopped{reason:"error"}` fires on a runtime trap (e.g. divide by zero) with diagnostic `error_codes`

https://claude.ai/code/session_01DCZpq9ziZokAHEzU2vuKxP

---
_Generated by [Claude Code](https://claude.ai/code/session_01DCZpq9ziZokAHEzU2vuKxP)_


docs/extensions/ironplc_playground.py
```diff
@@ -139,6 +150,10 @@ def _build_playground_url(
     if allows:
         params.append("allows=" + quote(allows, safe=","))
 
+    if host:
```
garretfick: I think these values are always required so why not make this an error if these parameters are not present?

playground/index.html
```diff
@@ -14,6 +14,16 @@
...
...cript")[0]).parentNode.insertBefore(p,r);var u=e;for(void 0!==a?u=e[a]=[]:a="posthog",u.people=u.people||[],u.toString=function(t){var e="posthog";return"posthog"!==a&&(e+="."+a),t||(e+=" (stub)"),e},u.people.toString=function(){return u.toString(1)+".people (stub)"},o="init capture register register_once register_for_session unregister unregister_for_session getFeatureFlag getFeatureFlagPayload isFeatureEnabled reloadFeatureFlags updateEarlyAccessFeatureEnrollment getEarlyAccessFeatures on onFeatureFlags onSessionId getSurveys getActiveMatchingSurveys renderSurvey canRenderSurvey identify setPersonProperties group resetGroups setPersonPropertiesForFlags resetPersonPropertiesForFlags setGroupPropertiesForFlags resetGroupPropertiesForFlags reset get_distinct_id getGroups get_session_id get_session_replay_url alias set_config startSessionRecording stopSessionRecording sessionRecordingStarted captureException loadToolbar get_property getSessionProperty createPersonProfile opt_in_capturing opt_out_capturing has_opted_in_capturing has_opted_out_capturing clear_opt_in_out_capturing debug".split(" "),n=0;n<o.length;n++)g(u,o[n]);e._i.push([i,s,a])},e.__SV=1)}(document,window.posthog||[]);
```
garretfick: What is this code? Can you just load a script from posthotg so that we are using the right pg script?

garretfick (commented):
See inline comments.

docs/extensions/ironplc_playground.py
garretfick: Agreed — `host` is now a required arg on `_build_playground_url` and `_build_iframe` (no default). Every directive already supplies it from `env.docname`, and any future caller that forgets will get a `TypeError` instead of an un-attributed URL. Fixed in 452e541.

---

playground/index.html
garretfick: That blob is PostHog's official browser snippet from https://posthog.com/docs/libraries/js — it's their script, just copy-pasted inline. It does two things: dynamically inserts a `<script>` tag for the real SDK (`/static/array.js`) and sets up a stub `window.posthog` so calls made before the SDK finishes downloading get queued and replayed.

A plain `<script async src=".../array.js">` on its own would drop our very first event (`playground_loaded` fires immediately when `app.js` runs, which is typically before the async script lands), so the snippet is doing real work.

To address the "what is this code?" concern I extracted it to its own file `playground/posthog-init.js` with a header comment explaining the above, and `index.html` now just references it via `<script src="posthog-init.js"></script>`. Fixed in 452e541.

---

PR #1094, opened 2026-06-02
Convert VM dispatch from recursive to iterative

garretfick:
Frame stack lives in caller-allocated VmBuffers (matches the existing
OperandStack/VariableTable pattern). CALL pushes a frame; RET pops one;
FB_CALL preserves its copy-in/copy-out semantics by tagging the frame
with FbCallReturn. Safety relies on the existing Trap::CallStackOverflow
fired from FrameStack::push, so no verifier or header.max_call_depth
plumbing is required for this PR. Existing CALL/RET and FB_CALL tests
are the regression net.


compiler/vm/src/vm.rs
```diff
@@ -2113,12 +2117,56 @@ pub(crate) fn execute_with_hook<H: DebugHook>(
...
+    variables: &mut VariableTable,
+) -> Result<(), Trap> {
+    let popped = frame_stack
+        .pop()
+        .expect("caller must hold the loop invariant: non-empty before return");
+    temp_alloc.rewind_to(popped.temp_alloc_mark);
+
+    if let Some(fbr) = popped.fb_return {
+        for i in 0..fbr.num_fields as usize {
+            let offset = fbr.instance_start + i * 8;
```
garretfick: Can this offset be computed in the fbr? Maybe inline it, but I don't like having the knowledge of "8" here.

PR #1095, opened 2026-06-03
Add WSTRING prep that doesn't touch the container format

garretfick:
Three purely-additive moves toward WSTRING that don't depend on the upcoming `STRING_HEADER_BYTES` 4 -> 6 / `FORMAT_VERSION` 2 -> 3 bump:

1. `StringType::char_width()` on `ironplc-dsl`, returning `CharWidth` from `ironplc-container`. The analyzer drops its local `char_width_for` helper and uses the method instead. Adds an `ironplc-container` dep on `ironplc-dsl`.
2. `ConstType::WStr = 7` enum variant plus `ConstEntry::wstring()` constructor, `ConstType::char_width()` accessor, and read/write/ `get_str` support that treats `WStr` the same as `Str` on the wire. No codegen path emits `WStr` yet, so no v2 file ever contains one -- the format version does not move.
3. The `Wide` arm of `encode_string_literal` is implemented as UTF-16LE (`(ch as u16).to_le_bytes()` per char, BMP only per ADR-0016) instead of `unreachable!`. No caller passes `Wide` yet.

Each move lands shape, not behavior, so the upcoming format-bump PR can focus exclusively on the wire-format break.

https://claude.ai/code/session_016gH2cPtk793rzYjAJcYftX


compiler/dsl/Cargo.toml
```diff
@@ -12,6 +12,7 @@ maintenance = { status = "experimental" }
 
 [dependencies]
 time = "0.3.47"
+ironplc-container = { path = "../container", version = "0.217.0" }
```
garretfick: This dependency isn't right. The dsl shouldn't depend on the container format.

garretfick: Good call — fixed in 79bba4d. Dropped the `ironplc-container` dependency from `ironplc-dsl` entirely and removed the `StringType::char_width()` method (its only purpose was to return `CharWidth`). The `StringType → CharWidth` mapping moves back to the analyzer-local `char_width_for` helper, where both `StringType` (dsl) and `CharWidth` (container) are already in scope, so the syntax tree no longer depends on the wire-format crate. The other two additive moves (`ConstType::WStr` and the UTF-16LE `encode_string_literal` arm) are unchanged. `cd compiler && just` is green.

---

compiler/container/src/constant_pool.rs
```diff
@@ -139,13 +148,15 @@ impl ConstantPool {
...
+    /// Gets a string value (raw bytes) from the constant pool at the given
+    /// index. Accepts both [`ConstType::Str`] (Latin-1) and
+    /// [`ConstType::WStr`] (UTF-16LE) entries.
     pub fn get_str(&self, index: ConstantIndex) -> Result<&[u8], ContainerError> {
         let entry = self
             .entries
             .get(index.raw() as usize)
             .ok_or(ContainerError::InvalidConstantIndex(index))?;
-        if entry.const_type != ConstType::Str {
+        if !matches!(entry.const_type, ConstType::Str | ConstType::WStr) {
```
garretfick: This idea of `matches` shows up in two places. Can we replace this with a function on ConstType that is `is_string_like`?

garretfick: Done in fbaaef8. Added `ConstType::is_string_like()` (returns true for `Str`/`WStr`) and used it to replace the duplicated `matches!(.., ConstType::Str | ConstType::WStr)` checks — three call sites collapse onto it: `get_str`, `read_from`, and the `primitive_le` debug-assert. The one remaining `Str | WStr` pattern is the `bytes()` match arm, which is a value-dispatch (returns `&str_value`) rather than a boolean test, so it stays a match arm. Added a unit test covering all eight variants. `cd compiler && just` is green.

---

PR #1099, opened 2026-06-08
Codegen: emit max_call_depth; VM rejects over-deep containers

garretfick:
Closes the loose end the iterative-dispatch PR (#1094) explicitly deferred: the call-frame buffer's artificial `MAX_CALL_DEPTH = 32` is now backed by a real per-program declaration that the VM checks before any init code runs.

**Depends on #1094** — the FrameStack infrastructure must land first, so this PR targets that branch as its base. Once #1094 merges to main, I'll retarget this PR.

## Summary

- **Codegen**: records one edge per `CALL` / user `FB_CALL` during emission. After all bodies are compiled, an iterative 3-color DFS from `FunctionId::SCAN` returns the longest path (depth 1 = entry only). Result is stamped on the already-reserved `FileHeader.max_call_depth` field via a new `ContainerBuilder::max_call_depth(u16)` setter.
- **VM**: `VmReady::start` rejects containers whose declared depth exceeds the frame buffer with `Trap::ProgramExceedsCallDepth { required, capacity }` (V9016). The check is skipped when `declared == 0` so hand-built test containers keep working.
- **Bug fix** the call-graph DFS surfaced: FB types and user functions both allocated function IDs starting at 2, so a function and an FB could share a runtime id. User functions now start at `2 + N_fb` so each POU has a unique id.
- **Cycle backstop**: if the DFS ever finds a cycle, codegen returns `Problem::InternalError` rather than looping forever. Semantic analysis (`Problem::RecursiveCycle`) should reject these earlier; this is defense-in-depth.

## Out of scope

Sizing `VmBuffers.frames` from `header.max_call_depth`. The buffer stays at fixed `MAX_CALL_DEPTH = 32`; tightening it requires threading an embedder-supplied frame budget through `VmBuffers::from_container` and is a follow-up.

## Test plan

- [x] `compiler/codegen/src/call_graph.rs` — 8 unit tests for the DFS (empty graph, chain, diamond, shared subtree, self-cycle, indirect cycle).
- [x] `compiler/codegen/tests/it/codegen_max_call_depth.rs` — 7 end-to-end tests compiling small IEC 61131-3 programs and asserting `header.max_call_depth`.
- [x] `compiler/vm/tests/it/load_max_call_depth.rs` — 4 tests covering the boundary cases (exceeds, equal, fits, zero-skip).
- [x] `compiler/vm/src/error.rs` — Display / `v_code` / `exit_code` tests for the new trap variant.
- [x] `compiler/container/src/builder.rs` — setter propagation tests.
- [x] `cd compiler && just` exits 0 (compile + coverage + clippy + fmt + dupes).
- [x] Full workspace `cargo test --workspace` passes including the previously failing `end_to_end_when_user_fb_body_calls_user_function_then_compiles` (now passes because the ID collision is fixed).

https://claude.ai/code/session_01MYbAUdSrAJyFz7BwKpZWaT

---
_Generated by [Claude Code](https://claude.ai/code/session_01MYbAUdSrAJyFz7BwKpZWaT)_

### DIFF
--- a/compiler/vm/src/vm.rs
+++ b/compiler/vm/src/vm.rs
@@ -147,6 +147,25 @@ impl<'a> VmReady<'a> {
     pub fn start(mut self) -> Result<VmRunning<'a>, FaultContext> {
         let shared_globals_size = self.container.task_table.shared_globals_size;
 
+        // Reject containers whose declared call depth would not fit in
+        // the embedder's frame buffer. Codegen populates
+        // `max_call_depth` from the static call graph; a value of 0
+        // means "not computed" (legacy or hand-built test containers)
+        // and disables the check, preserving the prior behavior where
+        // the runtime `Trap::CallStackOverflow` is the only signal.
+        let declared = self.container.header.max_call_depth;
+        let capacity = self.frames.len();
+        if declared != 0 && declared as usize > capacity {
+            return Err(FaultContext {
+                trap: Trap::ProgramExceedsCallDepth {
+                    required: declared,
+                    capacity: capacity.min(u16::MAX as usize) as u16,
+                },
+                task_id: TaskId::DEFAULT,
+                instance_id: InstanceId::DEFAULT,
+            });
+        }
+
         // Execute init functions once before entering scan mode.
         for pi in 0..self.program_instances.len() {
             let init_fn = self.program_instances[pi].init_function_id;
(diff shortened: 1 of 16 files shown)


compiler/vm/src/vm.rs
```diff
@@ -147,6 +147,25 @@ impl<'a> VmReady<'a> {
     pub fn start(mut self) -> Result<VmRunning<'a>, FaultContext> {
         let shared_globals_size = self.container.task_table.shared_globals_size;
 
+        // Reject containers whose declared call depth would not fit in
+        // the embedder's frame buffer. Codegen populates
+        // `max_call_depth` from the static call graph; a value of 0
+        // means "not computed" (legacy or hand-built test containers)
```
garretfick, line 153: Can we just reject anything with a value of 0?

PR #1105, opened 2026-06-11
Finish end-to-end WSTRING support

garretfick:


compiler/analyzer/src/rule_string_encoding_compat.rs
```diff
@@ -0,0 +1,316 @@
...
+    match &expr.kind {
+        ExprKind::Variable(var) => named_variable_encoding(var, string_vars),
+        _ => None,
+    }
+}
+
+/// Human-readable type name for an encoding, for the diagnostic context.
+fn encoding_type_name(encoding: &StringType) -> &'static str {
+    match encoding {
+        StringType::String => "STRING",
```
garretfick: Why not put this with the StringType enum itself?

garretfick (commented):
See comment

compiler/analyzer/src/rule_string_encoding_compat.rs
garretfick: Good call — moved it onto the enum as `StringType::keyword()` in `dsl/src/common.rs` and dropped the local helper. Done in 98fe462.

---

PR #1106, opened 2026-06-12
Codegen: emit VAR_NAME debug entries for function-local variables

garretfick:
## Summary

Layer 1 debugger work (`specs/design/debugger-support.md`). Until now, debug-section `VAR_NAME` entries (Tag 2) were emitted **only** for program/global-scope variables — every parameter and local inside a user-defined `FUNCTION` or `FUNCTION_BLOCK` body was invisible to the debug section. This closes Gap #2 (variable scope info) for the non-global case, which is what the DAP `scopes` / `variables` requests need to populate the Variables pane for a stack frame.

Each emitted entry carries:
- the **owning `function_id`** (not `GLOBAL_SCOPE`), so a debugger can attribute a slot to the frame that owns it;
- the real IEC `var_section` (VAR_INPUT / VAR_OUTPUT / VAR_IN_OUT / VAR / VAR_TEMP);
- the `iec_type_tag`, name, and source-level `type_name`.

The container format already had the `function_id` / `var_section` fields and round-trips them, so this is **codegen-only** — no format, VM, or VS Code changes.

## Details

- `compile_fn.rs`: a small `push_local_var_name` helper records an entry at each slot-assignment site — input params, locals, and FB fields in `compile_user_function` / `compile_user_function_block`.
- The function **return variable** is modeled as `VAR_OUTPUT` so a debugger surfaces it in the "Outputs" scope (per the section→DAP-scope table in the design doc). FB bodies have no return variable.
- `compile_setup.rs`: `map_var_section` is now `pub(crate)`, plus two new helpers `debug_type_for_decl` / `debug_type_for_return` mirror the global path's best-effort type resolution (composite types fall back to `OTHER`, exactly as globals do). No change to the existing global emission.

The variable table is globally partitioned (each POU's `var_offset` advances past the previous), so indices are already globally unique; the `function_id` tag is what lets a debugger group a frame's variables.

## Plan

`specs/plans/2026-06-11-debug-info-function-local-variables.md` (committed first).

## Out of scope (deferred)

- FB type/field name tables (Tags 4/5, design Gaps #5/#6) — Phase 5.
- DAP server, stepping, breakpoints, VS Code adapter — later phases.

## Test plan

- [x] `cd compiler && just` exits 0 (compile, coverage ≥ 85%, clippy, fmt, dupes).
- [x] New `end_to_end_debug_var_names.rs` — 4 tests through the real `compile()` path:
  - function params/local/return carry the function's id and correct sections/tags;
  - FB body input/output/local entries carry the FB body's id;
  - globals keep `GLOBAL_SCOPE` while function locals are owned by the function (no regression / leak);
  - two functions each own their own parameter with distinct ids and indices.

https://claude.ai/code/session_014FrPMDCLhcTZ7tTWEte2LF

---
_Generated by [Claude Code](https://claude.ai/code/session_014FrPMDCLhcTZ7tTWEte2LF)_


compiler/codegen/src/compile_fn.rs
```diff
@@ -22,12 +23,38 @@ use super::compile::{
...
 
+/// Records a debug [`VarNameEntry`] for a function- or FB-local variable
+/// (parameter or local) owned by `function_id`. Mirrors the program/global
+/// emission in `compile_setup::assign_variables`, but tags the entry with
+/// the owning function so a debugger can filter a frame's variables to the
+/// current stack frame (plus globals).
+fn push_local_var_name(
+    ctx: &mut CompileContext,
+    var_index: VarIndex,
+    function_id: u16,
```
garretfick: I thought function IDs here have a specific type. What's missing?

PR #1107, opened 2026-06-12
Add OpenCode integration end-to-end test

garretfick:
Validate that IronPLC's MCP server (ironplcmcp) works with the OpenCode
terminal agent, and guard the integration against regression.

Building the test surfaced a real bug: schemars emits a boolean JSON
schema ("options": true) for the pou_lineage, pou_scope, and symbols
tools because their `options` field lacked a description. OpenCode
rejects the entire tools/list when any tool uses a boolean property
schema, so no IronPLC tools loaded in OpenCode at all. Adding the field
description makes schemars emit an object schema, which OpenCode accepts
(verified: `opencode mcp list` now reports the server `connected`).

Test layers, cheapest first:
- Rust schema guard in compiler/mcp/tests/cli.rs: asserts no tool
  advertises a boolean `properties` schema. Runs in `cargo test`; no
  OpenCode required.
- Connectivity smoke (integrations/opencode): runs the real
  `opencode mcp list` and asserts the server is `connected`.
  Deterministic, no model, no API key.
- Agent end-to-end: runs `opencode run` against a local, key-free Ollama
  model (llama3.2:3b) and asserts the agent invokes `check` and the
  compiler returns diagnostics, observed via a JSON-RPC recording wrapper.

Wired into integration.yaml via partial_opencode_e2e.yaml (uses
ai-action/setup-ollama). Also documents the OpenCode config in the
AI-agents how-to guide.

https://claude.ai/code/session_01FTCe182FVr9JjFNt9rUgyW


.github/workflows/partial_opencode_e2e.yaml
```diff
@@ -0,0 +1,74 @@
...
+      # Configure the execution environment
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: '20'
+      - uses: dtolnay/rust-toolchain@631a55b12751854ce901bb631d5902ceb48146f7 # stable
+
+      # Build the MCP server under test
+      - name: Build ironplcmcp
+        working-directory: ./compiler
+        run: cargo build --release -p ironplc-mcp --bin ironplcmcp
```
garretfick: This should use what was actually built and what we posted to publish, similar to the VS Code smoke test, rather than rebuild. You therefore don't need the rust toolchain here or need to build.

.github/workflows/partial_opencode_e2e.yaml
```diff
@@ -0,0 +1,74 @@
...
+          IRONPLCMCP_BIN: ${{ github.workspace }}/compiler/target/release/ironplcmcp
+
+      # Layer 3 — real-agent end-to-end with a local, key-free Ollama model.
+      # NOTE: pin ai-action/setup-ollama to a commit SHA per the repo's policy
+      # once the SHA for v2 is known.
+      - name: Set up Ollama
+        uses: ai-action/setup-ollama@v2
+
+      - name: Start Ollama and pull the model
+        run: |
```
garretfick: The test steps should be written as a justfile recipe

.github/workflows/partial_opencode_e2e.yaml
garretfick: Done. Moved the OpenCode E2E out of per-PR `integration.yaml` and into the deployment pipeline (`deployment.yaml`, after `publish-prerelease`), consuming the just-published release via `install.sh` like the VS Code / install-script smoke tests. Dropped the Rust toolchain and the cargo build step. The per-PR regression guard for the underlying schema bug stays as the `cargo test` in `compiler/mcp/tests/cli.rs`, so PRs are still protected without a rebuild here.

---

.github/workflows/partial_opencode_e2e.yaml
garretfick: Done — the steps are now a `just opencode-e2e <version> <model>` recipe (installs the published compiler, runs the connectivity smoke, then the Ollama agent run). The workflow step just calls the recipe.

---

PR #1115, opened 2026-06-17
Design: unified variable inspection model

garretfick:
## Summary

Adds a design doc, **`specs/design/variable-inspection-model.md`**, defining how IronPLC presents PLC variables for inspection — in the `ironplcvm` CLI dump today and in the future DAP `variables` view. This PR is **docs-only** (one new file).

It addresses why the earlier `--dump-vars --group-by-scope` cut was the wrong unit: it grouped by **FUNCTION_BLOCK type**, but a FB type has no state — its *instances* do, and a program may have many of them.

## What the doc decides

- **Two orthogonal hierarchies**, which `--group-by-scope` conflated:
  - the **instance/data tree** (static, always inspectable): program → globals → FB-instance fields → nested instances → struct fields → array elements;
  - the **call-stack view** (only meaningful while paused): a FUNCTION's transient params/locals.
- **One shared `VarNode` model** that both the CLI dump and the DAP `variables`/`variablesReference` tree build, so they can't diverge.
- Grounded in the real data model: a composite variable's var-table slot holds a *data-region offset*; per-instance fields live in the data region; FB-body slots are shared scratch reused across instances via copy-in/out. A faithful tree must walk the data region using field layout we don't emit yet.
- **New debug-section tables** to carry that layout — `COMPOSITE_TYPE` (FB + struct fields), `VAR_TYPE_REF` (additive, doesn't touch the #1106 `VarNameEntry` format), and `ARRAY_TYPE`. This subsumes the deferred Phase 5 work from `debugger-support.md`.
- **Tag-registry fix:** `debugger-support.md` lists tag 4 as `FB_TYPE_NAME`, but the *implemented* tag 4 is `STRING_LAYOUT`. The doc keeps the implemented meaning, assigns fresh tags (10/11/12), and flags the stale registry.
- **`--group-by-scope` is demoted** to a flat fallback; the instance tree becomes the real view via `--dump-vars --tree`.

## Open questions (in the doc)

1. Nested composite base offset — inline bytes vs. offset pointer for composite *fields*.
2. Stdlib FB field tables — source from `builtin.rs` vs. hand-author.
3. `type_id` namespace — confirm FB and struct ids don't collide.

## Test plan

- [ ] Design review of `specs/design/variable-inspection-model.md`. Docs-only; no code change.

https://claude.ai/code/session_014FrPMDCLhcTZ7tTWEte2LF

---
_Generated by [Claude Code](https://claude.ai/code/session_014FrPMDCLhcTZ7tTWEte2LF)_


specs/design/variable-inspection-model.md
```diff
@@ -0,0 +1,384 @@
...
+This spec defines how IronPLC presents PLC variables for inspection — in
+the `ironplcvm` CLI dump today and in the DAP debugger's `variables` view
+later. It exists because the first cut (`ironplcvm run --dump-vars
+--group-by-scope`) grouped variables by **FUNCTION_BLOCK type**, which is
+the wrong unit: a FB type has no state; its **instances** do, and a program
+may have many of them.
+
+The model here is a single source of truth that both consumers render:
+
+- The **CLI dump** renders it as an indented text tree (with a flat list
```
garretfick, line 14: Indentation isn't the right model here. I understand the idea of keep it similar but we should separate display from data structure. Intended would waste space. I think we want a context header followed by the values

specs/design/variable-inspection-model.md
```diff
@@ -0,0 +1,384 @@
...
+
+Keeping one model prevents the CLI and the debugger from drifting into two
+incompatible notions of "what a variable is."
+
+This spec builds on and amends:
+
+- **[Debugger Support](debugger-support.md)** — Layer 1 debug info and the
+  DAP `variables` request. This spec supersedes that document's Gaps #5/#6
+  (FB type/field tables) and corrects its debug-section tag registry (see
+  [Debug-section tag reconciliation](#debug-section-tag-reconciliation)).
```
garretfick, line 43: Can we have this plan update that doc so that it is correct. We want to be spec anchored and that means updating specs as we go.

specs/design/variable-inspection-model.md
```diff
@@ -0,0 +1,384 @@
...
+
+## Problem statement
+
+### Two things called "scope" were conflated
+
+`--group-by-scope` produced output like:
+
+```
+[ACCUMULATOR]
+  step : DINT = 3  (VAR_INPUT)
```
garretfick, line 57: Nit: we want to use 61131 syntax. This isn't that

specs/design/variable-inspection-model.md
```diff
@@ -0,0 +1,384 @@
...
+
+To expand a composite node the inspector needs a **base offset** into the
+data region and a **field layout**:
+
+- **Base offset.** For a top-level composite variable, read the
+  variable-table slot — it already holds the `data_offset`
+  (`compile_setup.rs:474`). For a nested field, the base is computed from
+  the parent. *Open item:* confirm during implementation whether a nested
+  composite field stores an inline region (structs lay fields out
+  contiguously) or an offset pointer (as top-level FB instances do); the
```
garretfick: I'm pretty sure we did contiguously

specs/design/variable-inspection-model.md
```diff
@@ -0,0 +1,384 @@
...
+composite layout, avoiding the collision. `debugger-support.md`'s registry
+must be updated to match (action item, not a format change).
+
+| Tag | Name | This spec |
+|-----|------|-----------|
+| 10 | COMPOSITE_TYPE | FB + struct type descriptors (name + fields) |
+| 11 | VAR_TYPE_REF | var_index → type reference (for composite/array vars) |
+| 12 | ARRAY_TYPE | array layout for debug (element type + dims) |
+
+Tags 5, 7, 8 (the stale FB_*/LD/FBD reservations) are left unused and
```
garretfick, line 235: I'm not clear why we don't use those tags? No one really depends on this now.

garretfick:
## Design feedback: simplify var→type linkage, and make the instance tree inspectable without a live VM

Two design-level changes to the debug-info model in `variable-inspection-model.md`.

### 1. Drop the separate `VAR_TYPE_REF` table — fold the type reference into `VarNameEntry`

`VAR_TYPE_REF` (tag 11, REQ-VI-040..044) exists only to avoid changing the #1106 `VarNameEntry` wire format. We don't need that backward compatibility, so the indirection isn't worth its cost: a second table keyed on `(function_id, var_index)` forces a correlation step on every composite render and introduces failure modes the doc never specifies (a `VarNameEntry` whose `iec_type_tag == OTHER` but with no matching `VAR_TYPE_REF`, or an orphaned ref).

**Proposal:** add the `TypeRef` directly to `VarNameEntry`, populated for composite/array variables (scalars keep relying on `iec_type_tag`). One entry per variable, no join, no cardinality invariant to maintain. Update the #1106 conformance tests to the new format — that's acceptable now. Tag 11 goes away.

*Optional follow-on:* once `COMPOSITE_TYPE.name` carries the type name, the `type_name` string on `VarNameEntry` becomes redundant for composites (follow the ref to get the name). Can stay for now.

### 2. Make the instance tree resolvable from the container alone (no live VM for structure)

REQ-VI-005 reads the composite base offset from the live VM. But `data_offset` is a **compile-time constant** — `compile_setup.rs:474` already materializes it via `add_i32_constant`. Reading it live needlessly couples *structure* resolution to a running VM and forecloses container-only inspection (relevant to `iplc-file-viewer.md`).

**Proposal:** record the static `data_offset` in debug info (alongside the `TypeRef` on `VarNameEntry`, per item 1). Split the model cleanly:
- **Structure** (names, types, nesting, offsets) = static debug info, resolvable offline from the container.
- **Values** = overlaid when a VM is attached (or from static initial values).

Reword REQ-VI-005 accordingly.

These compose: a composite `VarNameEntry` carries both its `TypeRef` and its static base offset, so the whole instance tree's *shape* is described by the debug section with no VM in the loop.

---

garretfick:
## Design feedback: drop the `inline` field flag; keep `ARRAY_TYPE` self-contained with a consistency guard

### 3. Remove the `inline` flag from `FieldEntry` (REQ-VI-037) — derive indirection from the type kind

Indirection is **type-determined** for the in-scope (by-value) field set, so the flag is redundant:
- **Struct field** → laid out contiguously inside the parent region (`compile_struct.rs:522`), i.e. *inline*.
- **FB-instance field** → parent slot holds a `data_offset` pointer into the FB's own region (`compile_setup.rs:474`), i.e. *pointer*.

Both cases are exactly `COMPOSITE_TYPE.kind` (0 = struct → inline, 1 = function_block → pointer). The consumer dispatches on the referenced type's kind it already has to load, so storing `inline` on every field buys nothing. **Drop REQ-VI-037.**

This also reframes **open question 1**: it's no longer "store a flag for both cases" but "verify at codegen that no in-scope composite field is laid out against its kind's default." If a counterexample ever appears, that's the signal to introduce an explicit *binding* axis (by-value vs by-reference) rather than re-adding a bool.

Note for later: reference-bound fields (`REF_TO`, `VAR_IN_OUT`) are pointers because of how they're *bound*, not what they point at. When they come into scope they get a binding axis orthogonal to `TypeRef` — they do not resurrect `inline`.

### 4. `ARRAY_TYPE` (tag 12) is justified — keep it self-contained, add a consistency test

The new table is needed: the runtime `ArrayDescriptor` (`container/type_section.rs:80`) carries only a **flat** `total_elements` count and a `FieldType`-encoded scalar `element_type` — it has no IEC `lower..upper` bounds and no way to reference a `COMPOSITE_TYPE` for arrays of structs/FBs. So `ARRAY_TYPE`'s `dims` and composite `element_ref` are real additions, not duplication of what's already there.

Keep `ARRAY_TYPE` self-contained rather than referencing `ArrayDescriptor` — note the two even use **different element encodings** (`ArrayDescriptor.element_type` is a `FieldType` tag; `TypeRef` scalar id is an `iec_type_tag`), so a cross-reference wouldn't be clean anyway.

The one real risk is drift: the debug `element_ref` and the runtime `ArrayDescriptor.element_type` describing the *same* array could disagree. **Add a round-trip consistency test** asserting they map to the same element type (bridging the `FieldType` ↔ `iec_type_tag` encodings), so VM indexing and debugger display can't diverge silently.

---

garretfick:
## Resolving the remaining open questions (OQ2, OQ3) + the enum gap

The codebase answers all three concretely, so these don't need to stay "open."

### OQ2 (stdlib FB field tables) — source them from the existing codegen field maps, not builtin internals or a hand-authored table

The field maps **already exist and already carry names + offsets + the visible-field decision**. `compile_call.rs:1205+` defines `timer_fb_fields()`, `ctu_fb_fields()`, `ctd_fb_fields()`, etc., mapping field name → byte index:

```rust
// timer_fb_fields(): "Fields 4-5 are hidden (start_time, running) and not included."
"in"→0, "pt"→1, "q"→2, "et"→3
```

These are the **same maps that drive intrinsic FB lowering**, so they're the runtime source of truth, and they already exclude hidden implementation fields (e.g. timer `start_time`/`running`) — which is exactly the user-facing field set we want to display.

**Decision for REQ-VI-063:** generate stdlib `COMPOSITE_TYPE` entries from the `*_fb_fields()` maps. Pull per-field IEC type/section from the analyzer stdlib defs (`analyzer/src/intermediates/stdlib_function_block.rs`, which has `("CU", bool_type(), Input)` etc.) if needed. Do **not** hand-author a parallel table and do **not** read `vm/src/builtin.rs` — either would be a third place to drift from `*_fb_fields()`.

### OQ3 + my #10 (type_id namespace) — it's a **three-space** problem; partition the u16

The doc's OQ3 only worries about FB-vs-struct, but there are actually three id spaces that must coexist in one `COMPOSITE_TYPE.type_id` (and resolve unambiguously from a `TypeRef(kind=composite, id)`, which carries no struct/fb discriminator):

| Space | Current allocation | Source |
|---|---|---|
| Stdlib FB type ids | `0x0010`–`~0x00FF` (TON=0x0010, CTU=0x0020, R_TRIG=0x0040) | `opcode.rs:1279+` |
| User FB type ids | `0x1000`+ | `compile.rs:944` `next_user_fb_type_id` |
| Struct type ids | **don't exist yet** (structs keyed by name/`IntermediateType`, `compile_struct.rs:333`) | — must be invented |

So "structs start at 0" would collide with stdlib FBs. **Decision:** make `COMPOSITE_TYPE.type_id` a single globally-unique u16 with reserved ranges — e.g. stdlib FB `0x0000–0x0FFF`, user FB `0x1000–0x7FFF`, struct `0x8000–0xFFFF` — stated as a normative REQ. This corrects REQ-VI-031 ("type_id matches `FbInstanceInfo.type_id` for FBs"), which silently assumes the FB space is the whole space. (Alternative — widening `TypeRef.kind` to distinguish composite-struct from composite-fb — also works but is more invasive than partitioning an already-mostly-partitioned space.)

### My #12 (enum `TypeRef` path) — confirmed gap, needs a `kind = enum`

`iec_type_tag::OTHER = 255` and there is **no** enum tag; enums are tagged `OTHER` + `type_name` and resolved against `ENUM_DEF` (tag 9), which is keyed by `type_name`. Top-level enum vars work this way. But a struct/FB **field** of enum type goes through `FieldEntry.type_ref`, whose `kind` is only scalar/composite/array — so it would encode as `kind=scalar, id=255` and **lose the name table, rendering the raw ordinal**. That's a regression vs. how a top-level enum var displays.

**Decision:** add `TypeRef kind = enum`, with `id` referencing an `ENUM_DEF` entry. `ENUM_DEF` has no numeric id today (name-keyed), so give its entries a stable index and point `TypeRef.id` at it. For consistency, top-level enum vars' (now-merged, per item 1) `VarNameEntry` can carry the same `TypeRef(kind=enum)` instead of the `OTHER`+string path — unifying enum resolution for vars and fields.

### OQ1
Already addressed by dropping the `inline` flag (item 3 above): it's no longer "store a flag" but "verify at codegen that layout is type-determined."

---

PR #1176, opened 2026-07-12
Add one-click P9999 "Submit Code" reporting to playground

garretfick:
## Summary

Adds a low-friction, consent-forward "Submit Code" affordance to the playground diagnostics panel that appears when P9999 ("capability not implemented yet") diagnostics are shown. Users can submit their program source directly to the IronPLC team via PostHog without requiring a GitHub account, or open a prefilled GitHub issue as an alternative.

## Key Changes

- **Capture compiled source** (`app.ts`): Store `lastCompiledSource` when compilation starts so reports reflect what actually produced the diagnostics, even if the user continues typing.

- **Report panel rendering** (`app.ts`): When P9999 diagnostics appear, render a report panel with:
  - Clear consent language stating the program may be published publicly
  - A "Submit Code" button with explicit action verb
  - A secondary "open a GitHub issue" link with prefilled details

- **PostHog event capture** (`app.ts`): New `todo_report_submitted` event carries:
  - Program source (truncated to 50KB to prevent pathological pastes)
  - Error codes and diagnostic labels (including compiler `file#Lline` locations)
  - Program metadata (size, line count, dialect, allows, compiler version)
  - Transmission only on explicit user click (never automatic)

- **GitHub issue builder** (`app.ts`): Generates prefilled GitHub issue URLs with:
  - P9999 label and title
  - Diagnostic locations and compiler metadata
  - Program source inline when it fits under browser URL limits (~7KB)
  - "Attach the file" note for larger programs

- **Styling** (`style.css`): Added styles for report panel, consent text, button, and confirmation message with theme variable support.

- **E2E tests** (`e2e.spec.ts`): Four new tests verify:
  - Panel appears on P9999 (using `%QX0.0 := TRUE;` as reliable P9999 trigger)
  - Panel does not appear on ordinary syntax errors
  - Submit Code button captures event with program source
  - GitHub link is properly prefilled

- **Documentation** (`P9999.rst`): Updated to surface the playground one-click path as the easiest reporting method, with privacy notice.

- **Analytics** (`posthog.tf`): Added "P9999 code submissions" dashboard tile and clarified privacy comment to document that `todo_report_submitted` intentionally carries source on explicit, consented user action.

## Implementation Details

- **Consent-first UX**: Button text says exactly what happens ("Submit Code"), and a plain-language consent line directly above makes clear the code is shared and may become public.
- **Idempotent submission**: After successful submit, the panel is replaced with a confirmation and the button is removed, preventing double-submission.
- **Reliable P9999 trigger**: Direct hardware-address writes (`%QX0.0 := TRUE;`) reliably produce P9999 from code generation and are genuinely out of scope for a software playground VM.
- **URL length management**: GitHub issue URLs are kept under 7000 characters by omitting source when necessary.
- **Privacy preservation**: The existing automatic `compile_finished` event remains source-free; only the explicit `todo_report_submitted` event carries program code.

https://claude.ai/code/session_01WwpkRmJB3CJVFLRkmbHX8e

### DIFF
--- a/docs/reference/compiler/problems/P9999.rst
+++ b/docs/reference/compiler/problems/P9999.rst
@@ -18,8 +18,8 @@ Reporting the program
 ---------------------
 
 We can only add support for an unsupported capability once we can see the
-program that needs it, so please send it to us. There are two ways, and no
-GitHub account is required:
+program that needs it. Sending us the program gives the best results, so please
+do — there are two ways, and no GitHub account is required:
 
 * **From the** `IronPLC Playground`_ **(easiest).** When a program produces
   P9999, a **Submit Code** button appears next to the diagnostic. Clicking it
--- a/docs/reference/compiler/problems/P9999.rst
+++ b/docs/reference/compiler/problems/P9999.rst
@@ -12,7 +12,26 @@ requires adding support for the elements to the IronPLC compiler.
 
 * avoid the particular elements (this is usually not feasible)
 * verify that you have the latest version of the IronPLC compiler
-* report the issue to `IronPLC GitHub issues`_.
+* report the program so support can be added (see below).
+
+Reporting the program
+---------------------
+
+We can only add support for an unsupported capability once we can see the
+program that needs it, so please send it to us. There are two ways, and no
+GitHub account is required:
+
+* **From the** `IronPLC Playground`_ **(easiest).** When a program produces
+  P9999, a **Submit Code** button appears next to the diagnostic. Clicking it
+  sends the program directly to the IronPLC team.
+* **As a** `GitHub issue`_. Attach the source file (or paste it) so we can
+  reproduce the problem.
+
+.. note::
+
+   Whichever route you choose, the program you submit is shared with the
+   IronPLC team and **may become public** — for example in a GitHub issue.
+   Do not submit anything confidential.
 
 Known Limitations
 -----------------
@@ -27,4 +46,5 @@ produce this error:
 * Direct variables (hardware-mapped I/O such as ``%IX0.0``)
 * Enumerated values in expressions and ``CASE`` selectors
 
-.. _IronPLC GitHub issues: https://github.com/ironplc/ironplc/issues/new?template=todo_report.md
+.. _IronPLC Playground: https://playground.ironplc.com/
+.. _GitHub issue: https://github.com/ironplc/ironplc/issues/new?template=todo_report.md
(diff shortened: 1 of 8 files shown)


docs/reference/compiler/problems/P9999.rst
```diff
@@ -12,7 +12,26 @@ requires adding support for the elements to the IronPLC compiler.
 
 * avoid the particular elements (this is usually not feasible)
 * verify that you have the latest version of the IronPLC compiler
-* report the issue to `IronPLC GitHub issues`_.
+* report the program so support can be added (see below).
+
+Reporting the program
```
garretfick, line 17: Actually, when I think about it, the P9999 error code should also have a compiler file and line number. Can we have that part automatically reported. Then we can say we get better results with the real code.

We should then have a dashboard that shows most common file/line numbers. I don't know how difficult it is to get just this information and not the program itself.

garretfick (commented):
See comment

playground/src/app.ts
```diff
@@ -134,6 +134,22 @@ let stepInFlight = false;
...
+// The exact source last handed to the compiler. Captured on Start so a report
+// reflects what actually produced the diagnostics, even if the user keeps
+// typing afterwards.
+let lastCompiledSource = "";
+
+// Diagnostic codes for which we surface the "Submit Code" affordance. P9999
+// ("capability not implemented yet") can only be fixed once we can see the
+// program that triggered it, so we ask the user to send it. Extend this set to
+// offer submission for other codes.
+const REPORTABLE_CODES = new Set<string>(["P9999"]);
```
garretfick: Anything that starts with "P9" should be reportable. Those are compiler errors.

garretfick (commented):
Additional comments

docs/reference/compiler/problems/P9999.rst
garretfick: Done. The compiler `file#Lline` is now exposed as structured fields (`compiler_file`/`compiler_line`) from the WASM diagnostic and attached automatically to the existing `compile_finished` event as `error_locations` — this is the compiler's own location, never the program, so no consent is needed and it costs nothing extra. Added a "Top P9 compiler error locations" dashboard tile that ranks these. The program itself is still only sent on an explicit Submit Code click, which is where "better results with the real code" applies.

---

playground/src/app.ts
garretfick: Done — replaced the fixed `{P9999}` set with an `isReportable()` predicate matching `/^P9\d{3}$/`, so every P9xxx compiler error now gets the Submit Code affordance. The panel wording and the prefilled GitHub issue (label + title) generalize to whichever P9 code(s) fired.

---

PR #1185, opened 2026-07-12
Add help notes to diagnostics and expand playground dialect selection

garretfick:
## Summary

This change makes C-style comment errors actionable by adding a general-purpose help note system to diagnostics, and expands the playground to support all four compiler dialects (ed2, ed3, rusty, codesys) instead of just two. Together, these changes help users discover solutions when they encounter unsupported syntax.

## Key Changes

**Diagnostic Help Notes**
- Added `help: Vec<String>` field to `Diagnostic` with builder method `with_help()` and accessor `help()`
- Rendered in all three consumers:
  - CLI: mapped to codespan `.with_notes()`
  - LSP: appended to the diagnostic message
  - Playground: new `help` field on `DiagnosticInfo` JSON, styled as muted trailing notes
- Attached static, dialect-agnostic help note to P0004 (C-style comments): "Convert the comment to IEC 61131-3 syntax using `(*` and `*)`, or select a dialect that supports C-style comments."

**Playground Dialect Selection**
- Extracted dialect resolution logic into new `dialect_from()` function that:
  - Accepts canonical `Dialect::cli_name` values (`"iec61131-3-ed2"`, `"iec61131-3-ed3"`, `"rusty"`, `"codesys"`)
  - Supports legacy aliases (`"2003"` → ed2, `"2013"` → ed3) for backward compatibility
  - Defaults to `rusty` for empty string and unrecognized values (preserves lenient behavior for ~127 existing doc embeds)
- Updated HTML dropdown to list all four dialects with canonical names
- Updated TypeScript to handle new dialect type and legacy URL parameter mapping
- Updated e2e tests to reflect new dialect option values and display names

**Documentation**
- Added comprehensive plan document (`specs/plans/2026-07-12-c-style-comment-guidance-and-playground-dialects.md`) explaining the problem, approach, and design decisions
- Updated P0004 problem documentation to reference dialect selection as a solution path

## Implementation Details

- The help note system is intentionally kept separate from labels (which describe *what* is at a location) to maintain semantic clarity
- Backward compatibility is preserved: empty dialect string and legacy year-based values continue to work as before, ensuring no regression in existing documentation embeds
- The playground's interactive default is now strict `iec61131-3-ed2` (standards-first), while doc embeds remain lenient via the empty string default
- All four dialects are now equally discoverable in the playground UI, making it clear that C-style comments are a dialect choice, not an IronPLC feature

https://claude.ai/code/session_01TESyhhVuN9cWEVavGtJMBz

### DIFF
--- a/compiler/dsl/src/diagnostic.rs
+++ b/compiler/dsl/src/diagnostic.rs
@@ -124,6 +124,12 @@ pub struct Diagnostic {
     /// Additional descriptions to the constant description.
     pub described: Vec<String>,
 
+    /// Guidance on how to resolve the problem. Unlike labels (which describe
+    /// *what* is at a location), help notes explain *how to fix* the problem.
+    /// Rendered as trailing notes by the CLI and appended to the message by
+    /// the language server.
+    pub help: Vec<String>,
+
     /// Additional information about the diagnostic.
     pub secondary: Vec<Label>,
 
@@ -145,6 +151,7 @@ impl Diagnostic {
             description: problem.message().to_string(),
             primary,
             described: vec![],
+            help: vec![],
             secondary: vec![],
             source_file: None,
             source_line: None,
@@ -264,6 +271,21 @@ impl Diagnostic {
     pub fn with_secondary(mut self, label: Label) -> Self {
         self.secondary.push(label);
         self
+    }
+
+    /// Adds a help note describing how to resolve the problem.
+    ///
+    /// Help notes are for fix guidance (e.g. "use `(* *)` comments"), which is
+    /// intentionally kept out of labels. They are rendered as trailing notes by
+    /// the command line and appended to the message by the language server.
+    pub fn with_help(mut self, help: impl Into<String>) -> Self {
+        self.help.push(help.into());
+        self
+    }
+
+    /// Returns the help notes describing how to resolve the problem.
+    pub fn help(&self) -> &[String] {
+        &self.help
     }
 
     /// Sets the Rust source file and line that produced this diagnostic.
--- a/playground/index.html
+++ b/playground/index.html
@@ -30,8 +30,10 @@
             <option value="" disabled selected>Examples</option>
           </select>
           <select id="dialect-select" data-testid="dialect-select" title="IEC 61131-3 dialect">
-            <option value="2003" selected>IEC 61131-3:2003</option>
-            <option value="2013">IEC 61131-3:2013</option>
+            <option value="iec61131-3-ed2" selected>IEC 61131-3 Ed. 2</option>
+            <option value="iec61131-3-ed3">IEC 61131-3 Ed. 3</option>
+            <option value="rusty">RuSTy-compatible</option>
+            <option value="codesys">CODESYS-compatible</option>
           </select>
           <span id="dialect-badge" data-testid="dialect-badge" class="dialect-badge"></span>
           <label title="Milliseconds between scan cycles">
--- a/playground/index.html
+++ b/playground/index.html
@@ -29,12 +29,9 @@
           <select id="examples-select" data-testid="examples-select" title="Load an example program">
             <option value="" disabled selected>Examples</option>
           </select>
-          <select id="dialect-select" data-testid="dialect-select" title="IEC 61131-3 dialect">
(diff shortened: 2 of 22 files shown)


playground/src/types/messages.d.ts
```diff
@@ -4,7 +4,17 @@
...
-export type Dialect = "" | "2003" | "2013";
+// Canonical dialect names (matching the compiler's `Dialect::cli_name`), plus
+// the legacy year-based aliases and the unset default ("") for backwards
+// compatibility with older embeds and Sphinx directives.
+export type Dialect =
+  | "iec61131-3-ed2"
+  | "iec61131-3-ed3"
+  | "rusty"
+  | "codesys"
+  | "2003"
```
garretfick: Why do we need the year based dialects?  Can we just remove these entirely?

compiler/dsl/src/diagnostic.rs
```diff
@@ -124,6 +124,12 @@ pub struct Diagnostic {
     /// Additional descriptions to the constant description.
     pub described: Vec<String>,
 
+    /// Guidance on how to resolve the problem. Unlike labels (which describe
```
garretfick, line 127: I want you to also write an ADR that says we don't have an IronPLC dialect.

garretfick (commented):
See the comments

playground/src/types/messages.d.ts
garretfick: Removed entirely. Dialect strings are now exclusively the canonical `cli_name` values (`iec61131-3-ed2`, `iec61131-3-ed3`, `rusty`, `codesys`), resolved through `Dialect::from_str`. I also migrated the 6 doc embeds and the Sphinx directive example from `:dialect: 2013` to `:dialect: iec61131-3-ed3`, dropped the `LEGACY_DIALECTS` map and the unused `edition` URL alias, and updated the tests. (840d208)

---

compiler/dsl/src/diagnostic.rs
garretfick: Added as `specs/adrs/0036-no-ironplc-dialect.md` (840d208). It states that every accepted configuration maps to a real IEC edition or vendor dialect, the default stays strict, and rejected constructs are resolved by fixing the code or selecting an existing dialect — never by adding IronPLC-specific leniency. It ties into ADR-0012 (no dialect that "exists nowhere else") and explains why this PR keeps C-style comments gated rather than default-enabling them.

---

playground/index.html
```diff
@@ -30,8 +30,10 @@ <h1>IronPLC Playground</h1>
             <option value="" disabled selected>Examples</option>
           </select>
           <select id="dialect-select" data-testid="dialect-select" title="IEC 61131-3 dialect">
-            <option value="2003" selected>IEC 61131-3:2003</option>
-            <option value="2013">IEC 61131-3:2013</option>
+            <option value="iec61131-3-ed2" selected>IEC 61131-3 Ed. 2</option>
```
garretfick: Can we auto-generate this code or have this come from JS? This will again get out of date.

playground/src/types/messages.d.ts
```diff
@@ -4,7 +4,14 @@
 // dispatches into the WASM crate and posts back WorkerResponse. The WASM crate
 // returns JSON strings that app.ts parses into RunResult / LoadResult.
 
-export type Dialect = "" | "2003" | "2013";
+// Canonical dialect names (matching the compiler's `Dialect::cli_name`), plus
+// the unset default ("") used by embeds that do not force a dialect.
+export type Dialect =
+  | "iec61131-3-ed2"
```
garretfick: I want this to be defined by the compiler. Not repeated here

garretfick (commented):
More comments

playground/index.html
garretfick: Good call — it now comes from JS, sourced from the compiler. Added a WASM `dialects()` export that returns the list (`value`, `label`, `is_default`) straight from `Dialect::ALL`; the worker forwards it in the `ready` message and the UI builds the `<select>` at load. `index.html` no longer enumerates any dialects, so the picker can't drift from the compiler. I also dropped the duplicated `Dialect` string-literal union in `messages.d.ts` (now a documented `string` alias) for the same reason. (e3ddd14)

---

PR #1190, opened 2026-07-15
fix(cli): prevent compile output from overwriting source files

ShiroKSH:
## Summary

- prevent `ironplcc compile` from writing a bytecode container over any loaded source file
- compare file identity so relative paths, case differences, symlinks, hard links, and directory-discovered sources are covered
- add the P6009 diagnostic, regression coverage, and user documentation

## Root cause

`compile` loaded all sources into memory, then called `File::create(output)` without checking whether the output referred to an input. The command could therefore exit successfully after replacing source text with container bytes.

## Validation

- `cd compiler && just`
- `cargo test -p ironplc-cli --test cli compile_when_output_is_`
- `sphinx-build -a -W -n -b html . _build`


garretfick:
This appears to be an automated code change without a specific person behind the change. Unfortunately, this repo does not accept commits from bots that cannot be attributed to a person.

I'm closing this, but you can reopen if I have mis-characterized the author.

PR #1200, opened 2026-07-19
Keep trap v-code structured through the playground WASM boundary

garretfick:
The playground fault sites flattened a VM trap into a single
"VM trap: <Display> ..." message and discarded the structured v-code.
Preserve the code instead:

- Add a `VmFault` carrier that keeps the `Trap` itself (not a
  pre-stringified code), so the v-code is derived from `Trap::v_code`
  only when the JSON result is built.
- Change the `run_vm_scans`/`run_vm_step` helpers to return
  `Option<VmFault>` instead of `Option<String>`.
- Add `error_code: Option<String>` to `RunResult` and `StepResult`
  (serde default + skip when None); the run/step callers now set
  `error` from the message and `error_code` from the trap's v-code.
- messages.d.ts: add `error_code?` to `RunResultErr`.
- app.ts: in the VM-trap branch, report the trap code to analytics
  via `captureRunStopped` instead of an empty list.

Adds a Rust test asserting a divide-by-zero program surfaces
`RunResult.error_code == "V4001"` in the JSON.

Claude-Session: https://claude.ai/code/session_014HYrYD9UzyK7KqYFruWfiF

### DIFF
--- a/compiler/playground/src/lib.rs
+++ b/compiler/playground/src/lib.rs
@@ -23,7 +23,6 @@ use ironplc_dsl::core::FileId;
 use ironplc_dsl::diagnostic::{Diagnostic, LineColumn};
 use ironplc_parser::options::{CompilerOptions, Dialect, FeatureDescriptor};
 use ironplc_sources::{parse_source, FileType};
-use ironplc_vm::error::Trap;
 use ironplc_vm::{Slot, Vm, VmBuffers};
 use serde::{Deserialize, Serialize};
 use wasm_bindgen::prelude::*;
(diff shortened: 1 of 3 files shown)


compiler/playground/src/lib.rs
```diff
@@ -198,6 +214,11 @@ struct RunResult {
     scans_completed: u64,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     error: Option<String>,
+    /// The faulting trap's stable v-code (e.g. `"V4001"`) when execution
+    /// stopped on a VM trap. Kept separate from `error` so callers can act on
+    /// the code without re-parsing the message. Absent when there was no trap.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
garretfick: I would rather have an error object. The object has two members - message and code. Eventually we will likely move to Json rpc.

playground/src/types/messages.d.ts
```diff
@@ -122,6 +122,8 @@ export interface RunResultErr {
   ok: false;
   diagnostics?: Diagnostic[];
   error?: string;
+  /** The faulting trap's stable v-code (e.g. "V4001") on a VM trap. */
```
garretfick: This should be an error object with message and code members

playground/src/app.ts
```diff
@@ -564,7 +564,15 @@ function startStepLoop(): void {
     const result = JSON.parse(msg.json) as RunResult;
     if (!result.ok) {
       const diagnostics = result.diagnostics || [];
-      captureRunStopped("error", extractErrorCodes(diagnostics));
+      // Prefer compiler diagnostic codes; otherwise a VM trap carries its
+      // structured v-code in error_code (e.g. "V4001").
+      const errorCodes =
```
garretfick: Can we make the runtime also produce diagnostics when it fails. Then we don't need two representations.

garretfick (commented):
See comments

compiler/playground/src/lib.rs
garretfick: Done in e838380. Replaced the sibling `error`/`error_code` fields with a single `error: Option<RunError>` object, where `RunError { message, code }` carries the trap's v-code in `code` (absent for non-trap errors). One object internally, matching where JSON-RPC is heading.

---

playground/src/types/messages.d.ts
garretfick: Done in e838380. Added a `RunError { message: string; code?: string }` interface and pointed `RunResultErr.error` at it — same message/code shape as `Diagnostic`.

---

playground/src/app.ts
garretfick: Done in e838380. A runtime `RunError` is now converted to a `Diagnostic` (`runErrorToDiagnostic`) and rendered through the existing `renderDiagnostics` path, so compiler diagnostics and runtime faults share one representation — no separate error-message rendering. The v-code still reaches analytics because it rides in the folded diagnostic that `extractErrorCodes` already reads. (`renderDiagnostics` now skips the code chip for codeless infrastructure errors.)

I kept the internal representation simple — a `{ message, code }` object — and did the diagnostic conversion at the front end, per your note about a simple internal representation that's diagnostic-compatible once it reaches the UI. Happy to move the conversion into the Rust runtime instead if you'd rather it emit diagnostics directly.

---

PR #1202, opened 2026-07-19
Ask users to submit their program on non-user runtime stops

garretfick:
When a playground run stops for a reason other than the user pressing Stop —
a cycle overrun or a VM trap — invite the user to submit their program via the
existing consent-gated "Submit Code" flow, so we can reproduce and understand
failures (cycle overrun is common even on seemingly simple programs).

- Route cycle overrun through renderDiagnostics as a synthetic CYCLE_OVERRUN
  diagnostic, so it shares the trap/compiler rendering path and can offer the
  report panel. Record the run context (interval, measured execution, cycles)
  so a later Submit click can attach it.
- Broaden the report-panel gate: a V#### trap or CYCLE_OVERRUN now qualifies
  alongside P9xxx (isRuntimeReportable).
- Make the panel title and prefilled GitHub issue kind-aware (compiler vs
  runtime); runtime issues use a playground-runtime label and include the step
  interval and measured execution time.
- Rename the PostHog event todo_report_submitted -> report_submitted and add
  report_kind plus, for runtime stops, interval_ms/elapsed_ms/cycle_count. Point
  the PostHog dashboard tile at the renamed event and break it down by
  report_kind.
- Add e2e tests: cycle overrun shows the panel, Submit captures a runtime
  report with interval_ms and program, the GitHub link is prefilled, and a
  user-initiated stop shows no panel.

Claude-Session: https://claude.ai/code/session_01MP2AYvmMD2NdChREgD4dTi


playground/src/app.ts
```diff
@@ -148,6 +154,21 @@ function isReportable(code: string): boolean {
   return /^P9\d{3}$/.test(code);
 }
 
+// A runtime stop for a reason other than the user pressing Stop — a VM trap
+// (V#### code) or a cycle overrun — also gets the "Submit Code" affordance: we
+// can only understand why a program stopped (especially overruns on seemingly
+// simple programs) once we can see the program and the interval it ran with.
+function isRuntimeReportable(code: string): boolean {
+  return /^V\d{4}$/.test(code) || code === "CYCLE_OVERRUN";
```
garretfick: Why do we have the special value here? Let's make this a V code

garretfick (commented):
See comments

playground/src/app.ts
garretfick: Done in 1161479. A cycle overrun is the same condition as the VM's watchdog trap (a task exceeding its time budget), so it now reports as **V4003** instead of the synthetic `"CYCLE_OVERRUN"` string. `isRuntimeReportable` is now just the `/^V\d{4}$/` pattern, the overrun diagnostic renders a V4003 chip that links to the runtime docs, and I noted the playground's interval-based overrun in `V4003.rst`.

---

Issue #1199, opened 2026-07-18
TwinCAT vendor dialect support (tracking interest re: ADR 0012)


thusser:
Hi,

first, thanks for this project! I'm evaluating IronPLC as a diagnostics backend for two things:

1. An IntelliJ plugin for TwinCAT Structured Text (via `ironplcc lsp`), instead of writing an ST parser + type checker from scratch.
2. Linting TwinCAT projects in CI without needing a full TcXaeShell/Windows build agent — TcXaeShell has been painful to keep running reliably on self-hosted runners...

I found `specs/adrs/0012-accept-vendor-dialect-files-as-is.md` already proposes exactly this (Beckhoff TwinCAT as a dialect), so this is more a data point + expression of interest than a new request.

I let Claude ran `ironplcc check --dialect codesys` against 158 real `.TcPOU`/ `.TcGVL`/`.TcDUT` files across 8 of our TwinCAT projects. **17/158 (~11%)  parse clean today.** Ranked by how many files each blocks:

1. **`{attribute 'qualified_only'}` / `{attribute 'strict'}` pragma headers** — by far the biggest blocker. Emitted on nearly every Declaration by modern TwinCAT project templates; fails even trivially simple files (e.g. a plain 6-value enum) since there's no pragma support in the lexer at all.
2. `EXTENDS`/`IMPLEMENTS` and interface-typed variables (OOP) — matches ADR 0012's list.
3. `REFERENCE TO` — parsed as an identifier, then chokes on the following `TO`; distinct from the `REF_TO` that `codesys`/`rusty` already support.
4. `AT%I*` — Beckhoff's auto-assigned I/O address shorthand.
5.  `PI` used as a constant — not TwinCAT-specific, just a missing stdlib constant.

Given that ranking, pragma-skipping alone (treat `{...}` like a comment, no need to interpret contents yet) looks like it'd unblock a disproportionate number of otherwise-trivial files for very little grammar work.

Is TwinCAT support something already planned, or would a scoped PR starting with pragma-skipping be welcome? Happy to share the full file-by-file breakdown if useful.

I'd probably use Claude for most of the code (honestly: other priorities), but check everything myself and test it extensively.

Cheers,
Tim

garretfick:
I am very interested in extending IronPLC to better support syntax from other vendors. Even just using for "is this valid syntax" is a use case I'd like to have working across many syntax variants. This makes IronPLC useful for your use case of 'in CI' and also for MCP.

[...]

PR #1207, opened 2026-07-20
feat(twincat): CODESYS/TwinCAT pragma-skipping (--allow-pragmas)

thusser:
Part of the TwinCAT dialect work tracked in #1199.

Treats `{attribute '...'}` pragmas as opaque trivia (parsed and discarded, like a comment), gated behind a new `--allow-pragmas` flag (enabled by `--dialect=rusty` and `--dialect=codesys`).

Confirmed this is genuine CODESYS-core syntax (documented by 3S/CODESYS itself and by Schneider Electric's Machine Expert), not a TwinCAT-only invention -- so it's placed on the existing `Codesys`/`Rusty` dialects rather than a new TwinCAT-specific dialect.

See `specs/plans/2026-07-18-twincat-pragma-skipping.md` for the full design notes.

## Test plan

- [x] `cd compiler && just` passes (compile, coverage, clippy, fmt, dupes)
- [x] New unit tests for pragma-skipping across pragma variants
- [x] Verified against real TwinCAT project files


compiler/ironplc-cli/src/lsp_project.rs
```diff
@@ -474,6 +474,7 @@ impl From<LspTokenType> for Option<SemanticToken> {
             TokenType::RightParen => None,
             TokenType::LeftBrace => None,
             TokenType::RightBrace => None,
+            TokenType::Pragma => Some(COMMENT_INDEX),
```
garretfick: Wouldn't this be a keyword so then uses KEYWORD_INDEX

compiler/parser/src/parser.rs
```diff
@@ -199,7 +199,8 @@ parser! {
     rule whitespace() -> () = tok(TokenType::Whitespace) {} / tok(TokenType::Newline) {}
 
     rule comment() -> () = tok(TokenType::Comment) ()
-    rule _ = (whitespace() / comment())*
+    rule pragma() -> () = tok(TokenType::Pragma) ()
+    rule _ = (whitespace() / comment() / pragma())*
```
garretfick, line 203: This is close but not quite right. From the twincat docs, pragmas are not allowed within expressions but this would allow inside an expression. Can you look at what would be required so that this doesn't appear in a rule. One idea I have is to define `__` two underscores to mark where pragmas are allowed while not expanding too much (or use `__` where pragmas are not allowed).

compiler/ironplc-cli/src/lsp_project.rs
thusser: Good question — this was a deliberate choice, not an oversight. `TokenType::Pragma` maps to `COMMENT_INDEX` because pragmas are parsed and discarded like a comment in this implementation — their contents aren't interpreted yet (see the plan doc for this feature). Since the semantic-highlighting classification is meant to reflect what the compiler actually does with a token, "comment" felt more accurate than "keyword": nothing about the pragma's contents is understood or acted on, the same as a `(* ... *)` comment.

That said, I can see the argument for `KEYWORD_INDEX` too — TwinCAT/CODESYS editors do typically highlight `{attribute '...'}` distinctly from an ordinary comment, since it's a recognized syntactic construct even if we don't act on its contents. Happy to switch it if you'd still prefer that — just let me know.

garretfick (dismissed):
Overall this is an improvement. Primary ask is to make sure this doesn't allow pragmas where they are not permitted. I don't have CODESYS currently available so ask that you figure out where they are not allowed. If the CODESYS docs are wrong, then let me know.

thusser:
Trying this in Beckhoff's TcXaeShell right now. Will come back with results...

garretfick:
> Trying this in Beckhoff's TcXaeShell right now. Will come back with results...

Thank you. I may be OK with this as is if fixing that results in a significantly larger PR. I would then track that as an open issue.

compiler/parser/src/parser.rs
thusser: Good catch to double-check this — I tested it directly against real TcXaeShell rather than relying on the docs. Tried pragmas in exactly the positions you flagged as concerning: inside an arithmetic expression (`a {attribute 'hide'} + b`), inside a function call's argument list (`MAX(a, {attribute 'hide'} b)`), inside an array subscript on both the RHS and the assignment target (`arr[{attribute 'hide'} i]`), and inside a struct field-access chain (`s.{attribute 'hide'}x`). All six compiled without error.

So it looks like the docs describe a recommended/conventional placement, but the actual TwinCAT compiler is more permissive than that — it doesn't enforce "instruction position only." Given that, the current implementation (pragma as universal trivia, consumable anywhere whitespace/comments are) actually matches real compiler behavior rather than being a bug. Happy to dig into a specific file/example if you have one that this implementation incorrectly accepts — but based on this test, I don't think a placement restriction is warranted.

thusser:
Our TwinCAT projects are here:
https://github.com/brotlib
It's TwinCAT code for telescopes based on Beckhoff hardware.
Since those might not cover everything, I let Claude search for a handful of more popular projects and check against them as well.

compiler/ironplc-cli/src/lsp_project.rs
garretfick: I think keyword is the right one here

thusser: Done.

PR #1214, opened 2026-07-21
fix(twincat): allow a CASE branch with no statements

thusser:
## Summary

- A `CASE` branch whose body has zero statements (a label that falls through to nothing -- just a comment, or a genuine placeholder) previously failed to parse, with the error confusingly located at the *next* case label rather than the actual empty branch.
- Root cause: `case_element()`'s statement portion is `statement_list()`, which requires one-or-more statements. An empty branch has nothing for that to match, so `case_element()` fails entirely and backtracks; the parser then tries (and fails) to interpret the next label's bare integer as a statement for the empty previous branch.
- Verified as a real gap against actual TcXaeShell before implementing: a normal populated `CASE`, an empty branch followed by another label, and an empty branch as the last one before `END_CASE` all compile clean in real TwinCAT.
- New `statement_list_or_empty()` fallback rule, used only by `case_element()`. Deliberately not extended to `CASE`'s own `ELSE` clause or to `IF`/`FOR`/`WHILE`/`REPEAT` bodies, which share the same underlying gap but aren't part of the verified case.

## Test plan
- [x] Parser tests: empty branch followed by another label; empty branch as the last one before `END_CASE`; regression test for an ordinary populated branch
- [x] plc2plc round-trip test

### DIFF
--- a/compiler/parser/src/parser.rs
+++ b/compiler/parser/src/parser.rs
@@ -1520,11 +1520,6 @@ parser! {
     pub rule statement_list() -> Vec<StmtKind> = items:statements_or_empty()+ {
       flatten_statements(items)
     }
-    // A CASE branch can have zero statements (a label that falls through
-    // to nothing -- confirmed against real TwinCAT). statement_list() on
-    // its own requires at least one, so this falls back to an empty Vec,
-    // matching the semisep_or_empty pattern used elsewhere in this file.
-    rule statement_list_or_empty() -> Vec<StmtKind> = statement_list() / { vec![] }
     rule statements_or_empty() -> StatementsOrEmpty = _ tok(TokenType::Semicolon) _ { StatementsOrEmpty::Empty() } / s:semisep(<statement()>) { StatementsOrEmpty::Statements(s)}
     rule statement() -> StmtKind = assignment_statement() / selection_statement() / iteration_statement() / subprogram_control_statement()
 
@@ -1595,7 +1590,7 @@ parser! {
         span: SourceSpan::join(&start.span, &end.span),
       })
     }
-    rule case_element() -> CaseStatementGroup = selectors:case_list() _ tok(TokenType::Colon) _ statements:statement_list_or_empty() {
+    rule case_element() -> CaseStatementGroup = selectors:case_list() _ tok(TokenType::Colon) _ statements:statement_list() {
       CaseStatementGroup {
         selectors,
         statements,
--- a/compiler/parser/src/parser.rs
+++ b/compiler/parser/src/parser.rs
@@ -1520,6 +1520,11 @@ parser! {
     pub rule statement_list() -> Vec<StmtKind> = items:statements_or_empty()+ {
       flatten_statements(items)
     }
+    // A CASE branch can have zero statements (a label that falls through
+    // to nothing -- confirmed against real TwinCAT). statement_list() on
+    // its own requires at least one, so this falls back to an empty Vec,
+    // matching the semisep_or_empty pattern used elsewhere in this file.
+    rule statement_list_or_empty() -> Vec<StmtKind> = statement_list() / { vec![] }
     rule statements_or_empty() -> StatementsOrEmpty = _ tok(TokenType::Semicolon) _ { StatementsOrEmpty::Empty() } / s:semisep(<statement()>) { StatementsOrEmpty::Statements(s)}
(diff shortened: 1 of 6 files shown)


garretfick (commented):
**Bottom line:** This is a real, well-verified gap and the fix is close, but I think the change needs to be gated behind `--allow-missing-semicolon` rather than applied unconditionally in the grammar. As written it relaxes the strict default dialect too.

**Why it needs gating**

An empty `CASE` branch with no terminating semicolon isn't strictly valid IEC 61131-3. The grammar (Annex B) is:

```
case_element   ::= case_list ':' statement_list
statement_list ::= statement ';' { statement ';' }
statement      ::= NIL | assignment_statement | selection_statement | ...
```

`statement_list` is one-or-more `statement ';'`. The `NIL` alternative makes an empty statement legal, but only as `statement ';'`, i.e. a bare `;`. So strict IEC accepts `5: ;` and rejects `5:` falling straight through to the next label. The current grammar already mirrors this exactly: `statement_list() = statements_or_empty()+`, and `statements_or_empty()` already matches a bare `;`, so `5: ;` parses on the strict default today. What this PR adds is tolerance for the dropped `;`, which is precisely what `--allow-missing-semicolon` exists for.

Because of that, per the syntax-support-guide rule ("Anything not in the IEC 61131-3 standard must be gated behind an `--allow-x` flag"), this should ride on a flag. And since `parse_library(tokens)` has no access to `CompilerOptions`, a raw grammar rule like `statement_list_or_empty()` can't be made conditional, so it's on for every dialect including the default strict `iec61131-3-ed2`.

**Suggested approach**

Route it through the existing token-insertion path instead of the grammar:

- Revert `case_element()` to `statement_list()` and drop `statement_list_or_empty()`.
- Extend `insert_keyword_statement_terminators()` in `xform_tokens.rs` so that, when `allow_missing_semicolon` is set, an empty branch (a `:` reaching the next `case_list` / `ELSE` / `END_CASE` with no statement) gets a `;` injected, turning `5:` into the already-legal `5: ;`. Strict mode then still rejects it, exactly as IEC requires.

This keeps the default dialect faithfully strict, reuses the flag you already have (grouping guidance favors folding a variation into an existing flag), and needs no AST or renderer changes since the renderer already handles an empty `CaseStatementGroup.statements`.

**One follow-up on the tests**

The new tests parse with `CompilerOptions::default()` and assert success. Under gating those should flip to asserting an error on the default, with new tests using `allow_missing_semicolon: true` for the success cases.

Really nice verification work against real TcXaeShell, and good call scoping out `ELSE` and the loop bodies. Thanks for this.

---

garretfick (changes requested):
See the comment to reuse the allow missing semicolon flag

thusser:
Good catch, and the CS2001-style framing from the other PR review clarifies this well too — makes sense that a strict-grammar relaxation needs to ride on a flag instead. Reworked as suggested:

- Reverted `case_element()` to plain `statement_list()` and dropped `statement_list_or_empty()` — the strict default dialect is back to only accepting an explicit empty statement (`5: ;`).
- Extended `insert_keyword_statement_terminators()` in `xform_tokens.rs` (already gated behind `allow_missing_semicolon`) instead: while inside `CASE...END_CASE`, tokens right after a case label's `:` are buffered until we see either an unambiguous statement token (`:=`, `(`, `.`, `[`, `^`, or a statement keyword) or the next branch terminator (another `:`, `ELSE`, `END_CASE`). Reaching a terminator while still buffering means the branch was empty, so a `;` gets inserted there, turning `5:` into the already-legal `5: ;`.

I went with buffering rather than a single-token lookahead because of the "reaches the next `case_list`" part of your description — a plain integer/identifier right after the colon can't be classified on its own (it might be a real case label like `6:`, or the start of an assignment like `6 := ...` isn't valid but `CONST1 := ...` is), so it has to stay ambiguous until whatever follows resolves it. Otherwise consecutive empty numeric labels (`5:` immediately followed by `6:`) would get misread as `6` being a statement and the fix would silently not apply to that pattern — which is exactly the shape in the original repro.

Also updated the two success-case tests to use `allow_missing_semicolon`, added a test confirming the default dialect still rejects the unflagged case, and narrowed the plc2plc round-trip test's flag requirement to just the first parse (the renderer already emits an explicit `;` for an empty branch, so the re-parse of rendered output needs no flag).

garretfick (approved):

PR #1215, opened 2026-07-21
fix(sources): don't abort project discovery on one unresolvable .plcproj entry

thusser:
## Summary

- `ironplcc` used to treat *any* single unreadable `<Compile Include="...">` entry in a `.plcproj` file as fatal for the whole project, aborting before checking anything else -- even if every other referenced file was perfectly valid.
- Found via two real cases: a case-sensitivity mismatch between the `.plcproj` entry and the actual filename on a case-sensitive filesystem, and a genuinely missing referenced file. Both cases meant an entire project went completely unchecked, not because its files failed analysis, but because of one bad reference.
- Fixed by recording each unresolvable entry as a warning and skipping it, rather than aborting -- not a case-insensitive filesystem fallback (that would silently paper over a real project/file mismatch instead of surfacing it). The CLI's existing `handle_diagnostics` helper surfaces these warnings non-fatally.
- Also fixes a related bug along the way: when multiple `.plcproj` files exist in the same directory (e.g. a stale rename artifact), discovery now picks deterministically instead of erroring or picking non-deterministically.

## Test plan
- [x] `.plcproj` referencing a missing file returns a warning, not an error; the rest of the project still resolves
- [x] All `.plcproj` entries unresolvable returns empty with warnings (not a hard failure)
- [x] Multiple `.plcproj` candidates in one directory picks deterministically
- [x] Deeply nested `.plcproj` layouts still resolve correctly (regression)
- [x] End-to-end: a project with a valid file, a file with a real syntax error, and a `.plcproj` entry referencing a missing file correctly reports both the missing-file warning and the real syntax error, while the valid file passes silently


garretfick (changes requested):
Thanks for this, and for the thorough writeup and tests. You found a real bug: one bad reference shouldn't stop every other valid file in a project from being checked. That half of this PR is a genuine improvement and I'd love to get it in.

**The one thing I want to change: a missing referenced file should stay an error, not become a warning.** A `.plcproj` that names a file which doesn't exist is a broken project, and `ironplcc check` should fail on it. So I'm asking for changes, but the fix is small: keep the "don't abort the whole project" behavior, just leave the missing reference fatal.

There are two separable changes here:

**(A) Don't abort the entire project when one `<Compile Include>` entry is unresolvable.** This is the real fix. Keep it.

**(B) Demote an unresolvable reference from a fatal error to a non-fatal warning (exit 0).** This is the part I'd like to drop.

**Why it stays an error:**

This isn't a vendor-dialect question. [ADR-0012 (Accept Vendor Dialect Files As-Is)](https://github.com/ironplc/ironplc/blob/main/specs/adrs/0012-accept-vendor-dialect-files-as-is.md) covers parsing syntax the vendor toolchain accepts, and it explicitly **rejected the "Best-Effort with Warnings" option** because warnings erode trust and a noisy best-effort result is "worse than useless." A file the manifest names but that is absent isn't syntax we don't understand yet. It's a project-integrity failure. TwinCAT itself won't build that project.

[ADR-0005 (Safety-First)](https://github.com/ironplc/ironplc/blob/main/specs/adrs/0005-safety-first-design-principle.md) frames the domain cost: the thing we most want to avoid in a PLC toolchain is silently reporting success when something is actually wrong. A green `check` on a project that is missing a file it declares is exactly that silent pass. If a file the project says should be analyzed never gets analyzed, "success" is misreporting.

Other compilers back this up. A missing `<Compile>` in an MSBuild/C# project is **`CS2001: Source file could not be found`**, an error with a non-zero exit, yet `csc` still compiles every other file and reports their diagnostics too. TypeScript's `tsconfig` `files` array errors when a listed file is missing, while its `include` globs just match what exists. In both cases the answer to one bad entry is: collect the error and keep going, never downgrade it so the build passes. For true source-level includes (`#include`, `import`) missing files are universally fatal (MSVC `C1083`, GCC `fatal error`). Nobody warns-and-continues.

The precedent is consistent: keep checking the other files, but the missing reference stays fatal.

**On the "warning" itself:** the CLI has no warning tier. `handle_diagnostics` renders every diagnostic at `Severity::Error` (`cli.rs:361`), and there's no promotion flag. So this change doesn't produce a warning the user can act on. It produces a `P6004` that still prints as `error[P6004]: Unable to read file` in red but no longer sets `had_error`, so the command exits 0 (`cli.rs:214-228`). That's a silently non-fatal error, which is more confusing than a plain error.

**What I'd suggest:** keep change (A). Collect unresolvable entries and continue discovery so the other files are still checked, but route them through the normal error path so `had_error = true` and `check` still fails. That's the `CS2001` model exactly, and it's a small change from where you already are.

The case-sensitivity mismatch you hit is a legitimate separate problem, and I agree the fix there is not a case-insensitive fallback (your plan reasons through that well). The resolution is to fix or flag the casing as an error the user corrects.

Really appreciate the contribution. The discovery bug you surfaced is worth fixing and I'm glad you did.

---

thusser:
Thanks for the detailed review — the ADR-0012/ADR-0005 framing makes sense, and dropping the "warning" tier is clearly right given the CLI has no way to render one that isn't just a quieter error.

Pushed a fix that keeps (A) and drops (B) exactly as described:

- `DiscoveredProject::warnings` renamed to `errors` (same "don't abort discovery of the rest of the project" behavior, but now documented as something callers must still treat as a failure).
- `enumerate_files`/`create_project` in `cli.rs` now keep whatever files DID resolve (so they still get loaded and checked, and still report their own diagnostics) while also setting `had_error = true` from the unresolved reference — so `ironplcc check` still fails overall, matching the `CS2001` model you described.
- Added two tests at the `check()` level: one confirming a missing entry alone still fails the command, and one confirming a valid entry alongside a missing one both gets checked and still fails the command overall.

The plan doc has a new "Revision" section recording this. Let me know if I've missed anything.

garretfick (approved):

PR #1216, opened 2026-07-21
feat(lsp): merge multiple workspace folders into one compilation unit

thusser:
## Summary

- When an LSP client (e.g. VS Code) opens multiple workspace folders, the LSP previously only initialized from `workspace_folders.first()`, silently ignoring the rest -- so cross-folder type resolution (a common layout for multi-project TwinCAT solutions) never worked in the editor, even though the CLI's own multi-argument `ironplcc check dir1 dir2 ...` already merges multiple directories into one project correctly.
- Adds `initialize_many` (discovers and merges multiple directories without clearing between them, unlike naively looping over the existing single-directory `initialize`), a matching `Project::initialize_many` trait method, and wires `lsp.rs` to pass every `workspace_folders` entry instead of just the first.
- Mirrors the existing, working CLI pattern rather than inventing a new merge strategy.

## Test plan
- [x] Two workspace folders: a type declared in one folder resolves correctly when referenced from the other
- [x] A single folder still works (regression)
- [x] `Project::initialize_many` unit tests (two directories merge; single directory is unaffected)

### DIFF
--- a/compiler/ironplc-cli/src/lsp.rs
+++ b/compiler/ironplc-cli/src/lsp.rs
@@ -119,9 +119,7 @@ fn start_with_connection(
     match initialize_params.workspace_folders {
         Some(folders) => {
             debug!("Initialize server with workspace folders {folders:?}");
-            if let Some(folder) = folders.first() {
-                server.project.initialize(folder);
-            }
+            server.project.initialize_many(&folders);
         }
         None => {
             debug!("Initialize server without a workspace folder");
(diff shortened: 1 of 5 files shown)


garretfick (changes requested):
There are two tests failing. Can you fix those tests. Otherwise I'm ok with the PR

thusser:
Already caught and fixed this — the two failing tests were `initialize_many_when_single_folder_then_still_works` and `initialize_many_when_two_folders_then_cross_folder_type_resolves`, both Windows-only failures.

Root cause: the test helper `workspace_folder()` built its `file://` URI with a naive `format!("file://{}", dir.display())`, which breaks on Windows temp paths (backslash separators + a `C:`-style drive letter, neither valid directly in a URI). Fixed it to reuse the same separator-normalization + drive-letter-prefix logic already used by `UriKey::from_file_id` elsewhere in this file.

Pushed in de13ab28. CI passed locally before pushing; should go green here too once the workflow re-runs.

PR #1217, opened 2026-07-21
feat(twincat): register LTRUNC/LMOD as Beckhoff Tc2_Math stdlib functions

thusser:
## Summary

- `LTRUNC`/`LMOD` were assumed to be plain missing entries in the generic `TRUNC`/`MOD` table, but Beckhoff's own `Tc2_Math` docs show they're `LREAL`-only, library-specific functions: `LTRUNC` truncates the fractional part like `TRUNC` but returns `LREAL` instead of `ANY_INT` (not clamped to an integer type's range), and `LMOD` is a floating-point modulo (`LMOD(400.56, 360) = 40.56`), unlike the integer-oriented `MOD`.
- Registered conditionally under a new `--allow-extended-math-functions` flag, following the same conditional-registration pattern already used for `SIZEOF` -- not joined to the always-on core arithmetic tables, since a real TwinCAT project must reference the `Tc2_Math` library to use them.

## Test plan
- [x] `get_extended_math_functions()` returns both signatures with correct parameter/return types
- [x] `LTRUNC`/`LMOD` resolve cleanly with the flag enabled
- [x] `LTRUNC` produces diagnostics (undeclared function) with the flag disabled (regression/default-off)

### DIFF
--- a/compiler/analyzer/src/intermediates/stdlib_function.rs
+++ b/compiler/analyzer/src/intermediates/stdlib_function.rs
@@ -1045,6 +1045,40 @@ pub fn get_sizeof_function() -> FunctionSignature {
     )
 }
 
+// =============================================================================
+// Extended math functions (Beckhoff Tc2_Math library, Vendor Extension)
+// =============================================================================
+
+/// Returns the `LTRUNC`/`LMOD` function definitions.
+///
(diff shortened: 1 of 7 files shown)


garretfick (commented):
See my comments on PI. I would like this to be an external library. I think this will need access to the byte codes to truncate. I can provide guidance there if you want me to sketch out how that would work.

PR #1218, opened 2026-07-21
feat(twincat): register MODABS as a Beckhoff Tc2_Math stdlib function

thusser:
## Summary

**Stacked on #1217 (`ltrunc-lmod`)** -- this PR's branch is built directly on top of that one's, so the diff below currently includes both PRs' commits combined; it will automatically shrink to just this PR's own commit once #1217 merges into `main` (GitHub recomputes the diff against the new `main` at that point). This is a genuine dependency, not just a shared-file collision: `MODABS` is added to the same `get_extended_math_functions()` table and reuses the same `allow_extended_math_functions` flag that #1217 introduces.

- Same family as `LTRUNC`/`LMOD` (#1217): a Beckhoff `Tc2_Math` library function, `LREAL`-only, not core IEC 61131-3.
- Unlike `LMOD`, `MODABS` always returns an unsigned/non-negative modulo result (`MODABS(-400.56, 360) = 319.44`, where `LMOD` would return `-40.56`) -- used in NC-axis contexts where modulo values are conventionally unsigned.
- Added to the same `get_extended_math_functions()` table, gated by the existing `allow_extended_math_functions` flag (no new flag needed).

## Test plan
- [x] `get_extended_math_functions()` now returns `MODABS` alongside `LTRUNC`/`LMOD` with correct parameter/return types
- [x] `MODABS` resolves cleanly with the flag enabled
- [x] `MODABS` produces diagnostics (undeclared function) with the flag disabled (regression/default-off)

### DIFF
--- a/compiler/analyzer/src/intermediates/stdlib_function.rs
+++ b/compiler/analyzer/src/intermediates/stdlib_function.rs
@@ -1049,14 +1049,14 @@ pub fn get_sizeof_function() -> FunctionSignature {
 // Extended math functions (Beckhoff Tc2_Math library, Vendor Extension)
 // =============================================================================
 
-/// Returns the `LTRUNC`/`LMOD` function definitions.
+/// Returns the `LTRUNC`/`LMOD`/`MODABS` function definitions.
 ///
-/// Both are Beckhoff `Tc2_Math` library functions, not core IEC 61131-3
-/// (unlike `TRUNC`/`MOD` above) -- a real TwinCAT project must reference
-/// that library to use them. Registered conditionally based on the
-/// `allow_extended_math_functions` compiler option.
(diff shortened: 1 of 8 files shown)


garretfick (changes requested):
See my comments on the PI changes. I would like this to be a standard library external to ironplc because these are not part of the standard. New OP codes are OK in the compiler if needed, but the functions should be externalized.

thusser:
Superseded by the Tc2_Math compatibility library (merged), which now provides MODABS end to end. Verified locally: `MODABS(-400.56, 360.0)` matches this PR's own worked example, correctly gated behind `--library Tc2_Math` (undeclared-function diagnostic without it). Same disposition as #1219 -- closing in favor of the library mechanism.

PR #1219, opened 2026-07-21
feat(twincat): register PI as implicit LREAL global constant

thusser:
## Summary

**First in a stacked series of new-flag PRs** (`pi-constant` -> `var-initializer-expressions` -> `mixed-located-var-declarations` -> `sized-string-and-inline-fb-ctor` -> `explicit-enum-values` -> `and-then-operator`) opened this way purely to avoid each one colliding with the others in `options.rs`'s shared feature-count table if merged out of order -- not because they depend on each other's functionality. This one has no functional dependency on any other PR and can be reviewed/merged independently.

- Adds `--allow-math-constants` (Rusty/Codesys), which injects a `PI` `VarDecl` as a `GlobalVarDeclarations` element during `resolve_types`, so ordinary variable resolution, type checking, and codegen's generic top-level `VAR_GLOBAL` collection all just work with no codegen changes.
- `PI` resolves in *statement* context today (e.g. `x := PI/180.0;`, matching real usage in `ATAN2.TcPOU`), but not yet as a `VAR` initializer (`d2r : LREAL := PI/180.0;`, the dominant real-world usage pattern) -- `VAR` initializers currently only accept a bare literal, not an expression. A regression test pins down this known limitation explicitly.
- Fixes the end-to-end tests whose hardcoded global variable slot indices shifted once `PI` became an extra implicit global under Rusty/Codesys.

## Test plan
- [x] `PI` is not injected when the flag is disabled; injected as an `LREAL` constant when enabled
- [x] `PI` resolves cleanly in statement context with the flag enabled; produces diagnostics when disabled
- [x] Regression test documents that `PI` as a `VAR` initializer still fails to parse (tracked separately)
- [x] Full end-to-end / codegen test suite updated for the new global slot indices

### DIFF
--- a/compiler/analyzer/src/stages.rs
+++ b/compiler/analyzer/src/stages.rs
@@ -3,6 +3,10 @@
 //! The compiler as individual stages (to enable testing).
 
 use ironplc_dsl::{
+    common::{
+        next_block_id, ConstantKind, DeclarationQualifier, InitialValueAssignmentKind,
+        LibraryElementKind, RealLiteral, VarDecl, VariableIdentifier, VariableType,
+    },
     core::{FileId, Id, SourceSpan},
     diagnostic::{Diagnostic, Label},
 };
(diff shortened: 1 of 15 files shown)


compiler/analyzer/src/stages.rs
```diff
@@ -3,6 +3,10 @@
 //! The compiler as individual stages (to enable testing).
```
garretfick, line 3: Thanks for this. The implementation follows the pattern of `__SYSTEM_UP_TIME`. That said, I don't want to land PI as a per-constant flag. The key problem with the flag-per-constant approach is

* Coupling. `allow_math_constants` is tagged [Rusty, Codesys], welding which constants resolve to which dialect parses. Those are orthogonal: someone may want CODESYS syntax without its math library, or the constants on strict IEC.
* Extensibility.  `allow_math_constants` doesn't follow what the project declares.

What I'd like instead is constants come from referenced libraries like Tc2_Standard rather than compiler built-ins. This doesn't exist yet.

* A LibraryProvider + the standard library as data — hand-reproduced library .st files that live alongside the compiler and are parsed like any other source (filesystem for native, embedded for the WASM playground). PI becomes a one-line entry in a math library file, not Rust.
* An explicit include flag rather than a dialect tag — e.g. `--library`. This library is orthogonal to dialects.

This will mean shipping additional files via the various installation paths and tests to make sure those ship. Files go in `lib\<NAME>`. I'm OK for a steel thread (minimal) first pass that plums everything together and later extends.

The good news is your analysis work carries straight over: the VarDecl-in-resolve_types → generic GlobalVarDeclarations collection path you've validated is precisely what the library loader would reuse — so this becomes "route PI through a LibraryProvider + a math library file" rather than a from-scratch rewrite. I'm happy to sketch the LibraryProvider/library-selection design so we can align on the interface before you reshape this, but in general, I'd like this to follow what GCC does.

garretfick (changes requested):
See comments

thusser:
Superseded by the library work in #1307, which now handles this end to end (including the VAR-initializer case this PR couldn't). See comment on #1199.

PR #1220, opened 2026-07-21
feat(twincat): allow constant expressions in VAR initializers

thusser:
## Summary

**Second in a stacked series of new-flag PRs** (`pi-constant` #1219 -> **this** -> `mixed-located-var-declarations` -> `sized-string-and-inline-fb-ctor` -> `explicit-enum-values` -> `and-then-operator`), stacked purely to avoid colliding with each other in `options.rs`'s shared feature-count table if merged out of order -- not a functional dependency (this PR's diff currently includes #1219's commit too; it'll shrink to just this PR's own commits once #1219 merges). Combined with #1219's `PI` constant, this fully unblocks the dominant real-world TwinCAT pattern from issue #1199's original survey (`d2r : LREAL := PI/180.0;`).

- Adds `--allow-constant-initializer-expressions` (Rusty/Codesys): a `VAR` initializer can now be a constant *expression* (arithmetic between literals and/or `CONSTANT` references), not just a bare literal -- matching real CODESYS/TwinCAT code. The IEC 61131-3 standard's `constant()` grammar production is literal-only, so this is a vendor extension.
- Design: the parser now accepts any `expression()` in initializer position, producing a new `InitialValueAssignmentKind::SimpleExpr` placeholder when it isn't a bare literal. A new `xform_fold_initializer_expressions` pass runs early in analysis and always normalizes `SimpleExpr` back to the ordinary `Simple`/`ConstantKind` shape (substituting known `CONSTANT` references, then folding arithmetic) or emits `P4036`/`P4037` -- so the many existing call sites that pattern-match on `Simple` initializers need no changes.
- Two regressions surfaced only through the full test suite during original implementation and are covered by regression tests here: (1) a naive "try `constant()`, fall back to `expression()`" doesn't work in PEG, since `constant()` greedily matches a prefix and "succeeds" -- fixed by always parsing via `expression()` and dispatching on the resulting shape; (2) that change initially broke every negative-literal initializer in the codebase, since `expression()` wraps a leading `-` as a unary operator rather than using `constant()`'s built-in signed-literal parsing -- fixed by collapsing that shape back to a directly-negative `Const` unconditionally.

## Test plan
- [x] Negative integer/real literal initializers still parse as plain `Simple` (not the new `SimpleExpr` shape) with the flag disabled -- regression
- [x] `identifier := identifier`-shaped enum-default initializers still resolve as `EnumeratedType`, not `SimpleExpr` -- disambiguation regression
- [x] Arithmetic initializer parses unconditionally (grammar is permissive; only the fold pass gates on the flag)
- [x] Arithmetic initializer resolves to `SimpleExpr` with the flag enabled
- [x] End-to-end codegen test combining `PI` + arithmetic initializer
- [x] plc2plc round-trip test
- [x] Full workspace test suite (1000+ tests) passes


garretfick:
This adds into the AST indicators of the parse path. I'd like to take a different approach. A function on the AST node is ok, but we don't need to maintain a dedicated value on the AST. This is a design decision so I've added ADR 0040 to document the decision.

thusser:
Thanks for writing this up — ADR-0040 is a great reference for the whole series. One thing I want to flag before making changes here: the ADR's final revision (b179816d, "distinguish faithful-node folding from persistent provenance markers") specifically walks through `InitialValueAssignmentKind::SimpleExpr` from this PR under rule 4 and endorses it as-is:

> `InitialValueAssignmentKind::SimpleExpr` holds the *actual* initializer expression the user wrote — it is honest to the source, not a flag recording a parse path. A single normalization pass (`xform_fold_initializer_expressions`) makes the policy decision with that faithful expression in hand, then folds the node to the ordinary `Simple` literal shape before semantic analysis or code generation ever sees it. No downstream stage branches on it... This *is* the "derive the property each stage cares about" shape.

That matches this PR's actual implementation exactly (`SimpleExpr` → `xform_fold_initializer_expressions` → folded to `Simple` before analysis/codegen ever see it, nothing downstream branches on it). So I want to check: does the revised ADR already resolve this comment in favor of the current design, or is there something else here you'd still like changed? Happy to adjust either way, just want to make sure I'm not redesigning something the ADR already settled.

(For contrast, I agree completely on #1221's `in_mixed_var_block` — that one *is* the anti-pattern the ADR calls out, since it's a pure boolean with no source content that persists into the analyzer. I'll propose a concrete accessor-based redesign over there.)

compiler/analyzer/src/xform_fold_initializer_expressions.rs
```diff
@@ -0,0 +1,442 @@
...
+
+    if !folder.diagnostics.is_empty() {
+        return Err(folder.diagnostics);
+    }
+
+    result
+}
+
+/// Scan the library for global constant declarations with literal values.
+fn collect_constants(lib: &Library) -> HashMap<String, ConstantKind> {
```
garretfick: Identifier has a case-insensitive match. We have a `HashMap<Id, ConstantKind>`, not to a `String`. Then we don't need the `to_uppwercase` that shows later on. This also enables special case sensitive matching if that ever shows up in some grammar.

compiler/analyzer/src/xform_fold_initializer_expressions.rs
```diff
@@ -0,0 +1,442 @@
...
+
+    for element in &lib.elements {
+        match element {
+            LibraryElementKind::GlobalVarDeclarations(decls) => {
+                collect_from_var_decls(decls, &mut constants);
+            }
+            LibraryElementKind::ConfigurationDeclaration(config) => {
+                collect_from_var_decls(&config.global_var, &mut constants);
+                for resource in &config.resource_decl {
+                    collect_from_var_decls(&resource.global_vars, &mut constants);
```
garretfick: Global vars in configuration and resource are specific to particular to the configuration or resource declarations - they are only "half global" - and this doesn't yet account for that.

Taking the safety first approach, I'd rather have compilation fail than potentially have incorrect behavior.

My strong suggestions here is to address the "true" global vars first (I put "true global" because 61131-3 doesn't really have bar VAR_GLOBAL). Then do the configuration and resource scoped global vars as a follow-up PR.

compiler/analyzer/src/xform_fold_initializer_expressions.rs
```diff
@@ -0,0 +1,442 @@
...
+            VariableIdentifier::Symbol(id) => id.clone(),
+            VariableIdentifier::Direct(d) => match &d.name {
+                Some(name) => name.clone(),
+                None => continue,
+            },
+        };
+
+        if let InitialValueAssignmentKind::Simple(simple) = &decl.initializer {
+            if let Some(value) = &simple.initial_value {
+                constants.insert(name.to_string().to_uppercase(), value.clone());
```
garretfick: Inserting here is dangerous - it could overwrite an existing values. In worse case, different compilation order could result in different behavior. If the value exists, then this should be an error.

compiler/analyzer/src/xform_fold_initializer_expressions.rs
```diff
@@ -0,0 +1,442 @@
...
+            }
+        }
+    }
+}
+
+/// Recursively substitutes known constant references and folds arithmetic
+/// within an initializer's expression tree. Reuses the same binary/unary
+/// folding rules as `xform_fold_constant_expressions`.
+fn substitute_and_fold(expr: Expr, constants: &HashMap<String, ConstantKind>) -> Expr {
+    let kind = match expr.kind {
```
garretfick, line 140: What happens if this is a cycle. Would that terminate? Please add a test.

compiler/analyzer/src/xform_fold_initializer_expressions.rs
```diff
@@ -0,0 +1,442 @@
...
+            InitialValueAssignmentKind::SimpleExpr(se) => Ok(self.normalize(se)),
+            other => InitialValueAssignmentKind::recurse_fold(other, self),
+        }
+    }
+
+    fn fold_function_block_declaration(
+        &mut self,
+        node: FunctionBlockDeclaration,
+    ) -> Result<FunctionBlockDeclaration, Diagnostic> {
+        let saved = self.constants.clone();
```
garretfick: I think this could use `ScopedTable` but this works equally well.

garretfick:
## Architectural note: coupling between the two folding passes

This PR makes `xform_fold_initializer_expressions` import arithmetic helpers from its sibling `xform_fold_constant_expressions` (the helpers are promoted from private to `pub(crate)` to expose them): `integer_value`, `make_integer_constant`, `make_real_constant`, `fold_integer_binary`, `fold_real_binary`, `const_as_f64`.

Reusing the folding logic is the right call — the concern is only the *direction*: one `xform_*` pass now depends on another `xform_*` module's internals. That creates an implicit "pass B knows about pass A's guts" dependency with no real reason to exist, and an `xform_*` module (a transformation) isn't a natural home for primitives another pass consumes.

### Suggestion: extract the shared primitives into their own module

Move the six helpers into a new analyzer-internal module — e.g. `compiler/analyzer/src/constant_folding.rs` (or `fold_arithmetic.rs`) — as `pub(crate)`, and have **both** `xform_fold_*` passes depend on it. Neither transform reaches into the other; both depend on a shared leaf module.

This fits the existing conventions:
- **Naming**: `xform_*` = transformations, `intermediate_*` = data structures, `*_environment` = symbol tables. A pure-helper module is none of those, so a plain descriptive name is the natural slot.
- **Single responsibility**: the module's one job is "evaluate arithmetic over constant literals," well under the 1000-line limit.
- The helpers are pure (no environment, no diagnostics — they return `Option`), so they extract without dragging state along.

### Why not push it down into the `dsl` crate

Tempting, since these operate purely on DSL types (`ConstantKind`, `ExprKind`, `BinaryExpr`, `Operator`) — but I'd keep them in the analyzer. `dsl` is deliberately the data-model + generic-traversal layer (`fold.rs` is the `Fold`/`Visitor` machinery, not evaluation). These helpers encode compile-time evaluation *policy* — checked-overflow → don't fold, division-by-zero → don't fold, int→real promotion — which is compiler policy, not AST structure. If a consumer outside the analyzer ever needs the same arithmetic, that's the trigger to reconsider `dsl`; for two analyzer passes, an analyzer-internal shared module is the right altitude.

### One nuance

The higher-level dispatchers `try_fold_binary` / `try_fold_unary` stayed private, so the new pass reimplements the substitution/recursion around them. If you want to share the *whole* folding core rather than just the leaf arithmetic, move those into the shared module too — otherwise the dispatch logic will tend to drift and duplicate.

This is a clean follow-up refactor and doesn't need to block the PR.

---

garretfick (changes requested):
See comments

thusser:
Addressed all the review feedback:

**Correctness/safety comments:**
- Replaced `HashMap<String, ConstantKind>` (keyed by manual `.to_uppercase()`) with `ScopedTable<Id, ConstantKind>` — `Id`'s own `Hash`/`Eq` is already case-insensitive per the spec, so this drops the manual uppercasing entirely and gets real lexical scoping (`enter()`/`exit()`) in place of the hand-rolled clone/restore.
- Narrowed constant collection to true `VAR_GLOBAL` declarations only. `CONFIGURATION`/`RESOURCE`-scoped global vars are "half global" (visible only within their own configuration/resource) and this pass doesn't model that yet — deferred to a follow-up rather than silently treating them as unconditionally global.
- Registering a constant now uses `try_add` and diagnoses a same-scope duplicate (`P4011`) instead of silently overwriting the earlier value.
- Added a regression test for a mutual-reference cycle (`A := B; B := A`) — confirmed by construction (not just the test) that this can't actually recurse: `constants` only ever admits already-literal values collected before folding starts, so a substituted value is always terminal.

**Architectural note:** extracted the shared arithmetic primitives (`integer_value`, `make_integer_constant`, `make_real_constant`, `fold_integer_binary`, `fold_real_binary`, `const_as_f64`, `try_fold_binary`, `try_fold_unary`) into a new `analyzer/src/constant_folding.rs`. Both `xform_fold_constant_expressions` and `xform_fold_initializer_expressions` now depend on that shared module instead of one reaching into the other's internals.

Also rebased onto current `main` (picking up #1239/#1241/#1228) and found `--allow-constant-initializer-expressions` had no `feature_flag_conformance` fixture — added one.

Full CI green throughout.

garretfick (approved):

garretfick:
Merged in https://github.com/ironplc/ironplc/pull/1253

PR #1221, opened 2026-07-21
feat(twincat): allow AT-located variables mixed with plain vars in one VAR block

thusser:
## Summary

**Third in a stacked series of new-flag PRs** (`pi-constant` #1219 -> `var-initializer-expressions` #1220 -> **this** -> `sized-string-and-inline-fb-ctor` -> `explicit-enum-values` -> `and-then-operator`), stacked purely to avoid colliding with each other in `options.rs`'s shared feature-count table if merged out of order -- not a functional dependency (this PR's diff currently includes both prior PRs' commits too; it'll shrink to just this PR's own commits as #1219/#1220 merge). No functional dependency on either.

- Adds `--allow-mixed-located-var-declarations` (Rusty/Codesys): an `AT`-located variable (e.g. `tempSensor AT%I*: INT;`) can now appear inside an otherwise plain `VAR`/`VAR_INPUT`/`VAR_OUTPUT` block, instead of requiring its own dedicated block -- matching real TwinCAT code (28 of 158 surveyed files, ~240 `AT %I*`/`AT %Q*` occurrences). The IEC 61131-3 standard requires located variables in their own dedicated block; IronPLC's grammar already parsed the address syntax unconditionally but only inside that dedicated, all-located block type.
- Design: a new `located_var1_init_decl()` grammar rule lets a single declaration carry an `AT` clause within the ordinary `var_declarations()`/`input_declarations()`/`output_declarations()` machinery, producing the same `VariableIdentifier::Direct` shape the pre-existing dedicated-block rules already use unconditionally. Since that shape is otherwise indistinguishable from the always-allowed dedicated-block case, a new `in_mixed_var_block` marker on `DirectVariableIdentifier` records which path produced it, computed at the whole-block level (not per-declaration -- an all-located block using the new per-declaration syntax must not be flagged, since it's semantically identical to the dedicated block type). A new `rule_mixed_located_var_declarations` semantic rule gates on the marker, emitting `P4038` when the flag is off.
- Two bugs surfaced only by running tests, not by reading the grammar: `VarDeclarations::flat_map` (the function `var_declarations()` actually calls) ignored the location field entirely, always producing `Symbol`; and the mixed/not-mixed decision needed to move from per-declaration to whole-block scope once an all-located block was caught by the wrong grammar path and incorrectly flagged.

## Test plan
- [x] Mixed block (located + plain declarations) parses with the flag enabled, rejected (`P4038`) with it disabled
- [x] All-located block using the new per-declaration syntax is *not* incorrectly flagged (regression)
- [x] `VAR_INPUT`'s separate aggregation path also correctly gates on the whole-block marker
- [x] End-to-end codegen test
- [x] plc2plc round-trip test


thusser:
Per ADR-0040 rule 4 (and your comment on #1220), I want to redesign this away from `in_mixed_var_block: bool` before merging. Proposing the concrete shape here rather than just implementing it, since the exact representation isn't fully spelled out in the ADR and it touches shared DSL types.

**The constraint I ran into**: `variables: Vec<VarDecl>` on `FunctionBlockDeclaration`/`ProgramDeclaration`/etc. already flattens every `VAR`/`VAR_INPUT`/`VAR_OUTPUT` block into one list — block boundaries aren't preserved anywhere in the tree today. So an accessor computed purely from the *current* AST can't tell "these two declarations came from the same source block" at all; that information has to be captured faithfully somewhere, or it's unrecoverable once flattened.

**Proposed fix**: give each `VarDecl` a small, source-derived block identity — e.g. `block: BlockId` (a plain `usize`, assigned once per `VAR`/`VAR_INPUT`/`VAR_OUTPUT` occurrence as the parser builds it) — recording *which block this declaration came from*, not any policy conclusion about it. That's a faithful structural fact (analogous to `address_assignment` already being the faithful location the user wrote), not an inference like "is mixed."

Then replace the precomputed boolean with an accessor, something like:

```rust
/// Declarations that are `AT`-located but share a block with at least one
/// plain (symbolic) declaration -- derived from block identity, not stored.
pub fn mixed_located_var_decls(variables: &[VarDecl]) -> impl Iterator<Item = &VarDecl> { ... }
```

`rule_mixed_located_var_declarations.rs` calls this accessor directly instead of reading `direct.in_mixed_var_block`. `mark_mixed_located_var_block`/`mark_mixed_located_var_block_in_place` in `vars.rs` go away entirely — the parser just assigns `block` ids as it builds each block's `Vec<VarDecl>`, no cross-declaration bookkeeping needed at parse time at all.

This keeps `variables: Vec<VarDecl>` itself flat and unchanged for every other consumer (codegen, other analyzer rules, XML transform) — only the location-relevant path gains the `block` field and the new accessor.

Let me know if this matches what you had in mind, or if you'd rather see a different shape (e.g. keying off block source spans instead of a synthetic id, or something else) before I implement it.

garretfick:
My thinking here was that flattening is the root of the problem and that flattening is actually unnecessary. In short, my thinking was to remove the flattening entirely accepting that this would introduce new nodes throughout. However that's quite a bit of work and I want to find a way to unblock. I'll accept the block ID concept if that leads to resolving this.

thusser:
Implemented the `BlockId` approach. Each `VarDecl` now carries a `block: BlockId` (a plain, process-globally-unique `usize` assigned by `next_block_id()`), and the old precomputed `in_mixed_var_block` boolean is gone entirely — `DirectVariableIdentifier` no longer carries any policy conclusion at all.

`mixed_located_var_decls(variables: &[VarDecl]) -> impl Iterator<Item = &VarDecl>` (in `dsl::common`) derives the mixed set by grouping on `block`: a block is "mixed" only if it contains both a `Direct` and a `Symbol` declaration. `rule_mixed_located_var_declarations.rs` now walks at the POU level (`FunctionBlockDeclaration`/`FunctionDeclaration`/`ProgramDeclaration`, via the existing `HasVariables` trait) and calls the accessor directly, instead of reading a per-node flag.

Declarations built anywhere other than the parser's own block-grouping logic (test helpers, the XML transform, etc.) each get their own fresh id from `next_block_id()`, so they can never accidentally collide with an unrelated declaration's block.

One design note: `block` is deliberately excluded from `VarDecl`'s `PartialEq` (now a manual impl) since it's parse-order provenance, not semantic content — otherwise tests comparing a manually-built expected AST against a real parse, or comparing a re-parsed rendered AST against the original, would fail on non-deterministic id mismatches whenever the renderer regroups declarations differently than the source did.

Also caught and fixed a real, unrelated gap while touching this: `--allow-mixed-located-var-declarations` had never been wired into the CLI's clap flags at all (only into `FEATURE_DESCRIPTORS`), caught by the existing `FEATURE_DESCRIPTORS`-vs-CLI-args consistency test.

garretfick:
Merged in another PR that resolved merge conflicts

PR #1222, opened 2026-07-21
feat(twincat): STRING(n)/WSTRING(n) parentheses + FB call-style instance init

thusser:
## Summary

**Fourth in a stacked series of new-flag PRs** (`pi-constant` #1219 -> `var-initializer-expressions` #1220 -> `mixed-located-var-declarations` #1221 -> **this** -> `explicit-enum-values` -> `and-then-operator`), stacked purely to avoid colliding with each other in `options.rs`'s shared feature-count table if merged out of order (this one adds no new flag itself, but sits in the stack between others that do). This PR's diff currently includes all three prior PRs' commits too; it'll shrink to just this PR's own commits as they merge. No functional dependency on any of them -- bundles two unrelated sub-fixes that happened to be surveyed together:

**1. `STRING(n)`/`WSTRING(n)` parenthesis form** -- adds parenthesis-delimited `STRING`/`WSTRING` length as an alternate to the bracket form (`STRING(255)` alongside `STRING[255]`) in `VAR` declarations and function return types, matching the existing unconditional `string_type_declaration__parenthesis()` precedent already used for `TYPE` aliases. Unconditional, no dialect flag needed.

**2. CODESYS/TwinCAT call-style FB instance initialization** -- `comm : FB_Comm(retries := 3, THIS);`, reusing the same positional-or-named parameter shape as an ordinary FB call. Uncovered two pre-existing, unrelated bugs along the way:
- `fb_name_decl()` was dead code due to a spurious trailing-comma requirement in `commasep_oneplus()`.
- `xform_toposort_declarations.rs` ordered FunctionBlock-initializer dependency edges backwards relative to `Structure`/`LateResolvedType`, causing a spurious `P2011` once eager `FunctionBlock` construction became reachable for a real, user-declared FB type.

## Test plan
- [x] `STRING(n)`/`WSTRING(n)` parens parse in `VAR` declarations and function return types; standard bracket form still parses unchanged (regression)
- [x] FB call-style init parses with named args, positional args, and both combined; empty parens; bare declaration (no initializer) still resolves via the existing late-bound path (regression); standard `:= (member := value)` struct-init form still parses (regression)
- [x] plc2plc round-trip tests for both features (string parens normalize to brackets on render, matching the existing `STRING[n]`-only DSL shape)

### DIFF
--- a/compiler/analyzer/src/stages.rs
+++ b/compiler/analyzer/src/stages.rs
@@ -359,4 +359,41 @@ END_FUNCTION_BLOCK";
         let src = read_shared_resource(name);
         parse_program(&src, &FileId::default(), &CompilerOptions::default()).unwrap()
     }
+
+    // ---------------------------------------------------------------------
+    // FB-instance call-style initializer.
+    // See specs/plans/2026-07-19-twincat-sized-string-and-inline-fb-ctor.md.
+    // ---------------------------------------------------------------------
+
+    #[test]
+    fn analyze_when_fb_call_style_init_references_earlier_declared_fb_then_no_diagnostics() {
+        // End-to-end regression for the toposort edge-direction bug found
+        // while adding this feature: `comm : FB_Comm(retries := 3);`
+        // constructs InitialValueAssignmentKind::FunctionBlock directly at
+        // parse time (unlike a bare declaration, which defers to
+        // LateResolvedType). Before the xform_toposort_declarations.rs fix,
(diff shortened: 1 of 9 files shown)


garretfick:
@thusser can you split this into two different PRs so that I can review the string and the FB call separately (unless there is a reason that these are the same)

garretfick:
I will implement this in https://github.com/ironplc/ironplc/pull/1273

PR #1223, opened 2026-07-21
feat(twincat): explicit enum-value assignment + base-type suffix

thusser:
## Summary

**Fifth in a stacked series of PRs** (`pi-constant` #1219 -> `var-initializer-expressions` #1220 -> `mixed-located-var-declarations` #1221 -> `sized-string-and-inline-fb-ctor` #1222 -> **this** -> `and-then-operator`). This one adds no new dialect flag itself (there's no keyword to gate a grammar change against, matching the `STRING(n)`/qualified-call precedent), so it doesn't strictly need to be stacked for count-table reasons -- it's simply next in the original branch order and has no functional dependency on the others. This PR's diff currently includes all four prior PRs' commits too; it'll shrink to just this PR's own commits as they merge.

- Adds CODESYS/TwinCAT explicit per-member enum values (`E_Mode : (Deutsch := 1, English := 2);`) and the enum base-type suffix (`(A, B) BYTE;`), using ordinary C-style continuation semantics for unlabeled members after an explicit one.
- Sizing and codegen's ordinal map now resolve actual member values via a shared `resolve_ordinal_values()` helper instead of raw declaration position, so an explicit value can both exceed the member count (sizing) and be reflected at runtime (codegen).
- Found and fixed two pre-existing, unrelated plc2plc renderer bugs while adding the new value's rendering: no `visit_enumerated_value` override existed at all, so a qualified enum value reference (`COLOR#RED`) silently rendered as `"COLOR RED"` (dropped the `#`); this branch's new `explicit_value` field would have been silently dropped too without a dedicated override.

## Test plan
- [x] Explicit per-member values parse and resolve correctly, including values that skip/exceed ordinary declaration-position ordinals
- [x] Enum base-type suffix (`(A, B) BYTE;`) parses and sizes correctly
- [x] Codegen reflects explicit values at runtime, not raw declaration position
- [x] plc2plc round-trip tests, including the two pre-existing renderer bugs found and fixed


specs/plans/2026-07-19-twincat-explicit-enum-values.md
```diff
@@ -0,0 +1,269 @@
...
+
+Survey item 1 from `twincat-status.md`'s "Next" list (11 files). Support
+CODESYS/TwinCAT enum declarations where members carry explicit integer
+values, and an optional underlying integer/bit-string base type after the
+value list:
+
+```
+TYPE E_ModeLanguage : (Deutsch := 1, English := 2); END_TYPE
+
+TYPE E_AssertionType :
```
garretfick, line 16: This is also valid for edition 3 and later

specs/plans/2026-07-19-twincat-explicit-enum-values.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Standard-vs-extension status (per the open question in `twincat-status.md`)
+
+Beckhoff's own docs describe per-member explicit values
+(`Red := 2, Green, Blue := 10`) as ordinary syntax, and separately call
+out *custom enum base types* (e.g. `BYTE` instead of the implicit
+default) as "extension beyond IEC 61131-3." This matches IEC 61131-3:2013
+(Edition 3), which added explicit enum values as standard syntax. In
+practice this distinction doesn't change the implementation approach
+here: **the parser has no mechanism to gate a non-keyword grammar change
```
garretfick, line 67: The parser doesn't have such a gate but this is becoming difficult to achieve

specs/plans/2026-07-19-twincat-explicit-enum-values.md
thusser: Good catch — reworded the Goal section (2863cfe6) to say explicit enum-value assignment is standard IEC 61131-3:2013 (Edition 3) syntax, not a CODESYS/TwinCAT extension. Only the trailing base-type suffix (`) BYTE;`) is the genuine vendor extension, per Beckhoff's own docs (see the "Standard-vs-extension status" section below in the plan).

specs/plans/2026-07-19-twincat-explicit-enum-values.md
thusser: Agreed this is worth flagging explicitly. Confirmed directly: parser::parse_library never receives CompilerOptions today — only the lexer stage does, through the keyword-demotion transforms. Non-keyword grammar changes (`:=` and integer literals here, and `STRING(n)` parens / qualified method calls in two earlier PRs) currently have no gate at all and parse unconditionally under every dialect.

For this PR specifically I don't think it's a practical problem: explicit enum values are Edition 3 standard, and the base-type suffix (the one actual extension) is a strict grammar superset with no real files exercising a conflicting reading, so parsing it permissively everywhere doesn't misparse anything today.

But I agree the pattern doesn't scale — each new permissive-superset addition makes it harder to reason about what's "really" enabled under a given dialect, and eventually something will need to be edition/dialect-gated at the grammar level, not just the lexer. That's a real architecture piece (threading CompilerOptions into the peg grammar, or a post-parse gating pass) that's bigger than any one of these feature PRs. Happy to write it up as a separate plan/issue if you'd like it tracked — just let me know.

garretfick:
Merged in https://github.com/ironplc/ironplc/pull/1252 due to merge conflct

PR #1224, opened 2026-07-21
feat(twincat): recognize AND_THEN short-circuit boolean operator

thusser:
## Summary

**Last in a stacked series of new-flag PRs** (`pi-constant` #1219 -> `var-initializer-expressions` #1220 -> `mixed-located-var-declarations` #1221 -> `sized-string-and-inline-fb-ctor` #1222 -> `explicit-enum-values` #1223 -> **this**), stacked purely to avoid colliding with the earlier ones in `options.rs`'s shared feature-count table if merged out of order -- not a functional dependency. This PR's diff currently includes all five prior PRs' commits too; it'll shrink to just this PR's own commits as they merge.

- Adds `--allow-short-circuit-operators`, recognizing `AND_THEN`: a genuine Beckhoff/CODESYS extension (verified against Beckhoff's docs) -- a short-circuit `AND` that only evaluates its right operand when the left is `TRUE`, commonly used to guard a dereference (`ptr <> 0 AND_THEN ptr^ = 99`).
- Kept as a distinct `CompareOp::AndThen` variant rather than folding into `CompareOp::And` -- unlike `REFERENCE TO`/`POINTER TO` -> `REF_TO` (behaviorally identical spellings), `AND_THEN`'s short-circuit vs. eager evaluation is a real, externally-visible difference in TwinCAT/CODESYS itself, so normalizing it away on render would not be behavior-preserving for a real downstream toolchain.
- `ironplcc check` fully supports `AND_THEN` (parsing, type-checking, round-tripping with its spelling preserved). Codegen doesn't implement short-circuit (conditional-branch) evaluation for any boolean operator today, so rather than silently emit eager (behaviorally wrong, unsafe) bytecode, compiling an `AND_THEN` expression now fails explicitly with `P9999` instead of miscompiling.

## Test plan
- [x] `AND_THEN` parses and type-resolves like `AND` with the flag enabled
- [x] Demotes to an ordinary identifier when the flag is disabled (regression, matching every other vendor-extension keyword)
- [x] Real-world motivating shape (guarding a dereference behind a null-pointer check) parses correctly
- [x] plc2plc round-trip preserves the `AND_THEN` spelling (not normalized to `AND`)
- [x] Codegen explicitly refuses to compile it (`P9999`) rather than emitting incorrect eager bytecode

### DIFF
--- a/compiler/dsl/src/textual.rs
+++ b/compiler/dsl/src/textual.rs
@@ -610,9 +610,7 @@ pub enum CompareOp {
     /// CODESYS/TwinCAT short-circuit `AND` (Beckhoff/CODESYS extension).
     /// Kept as a distinct variant from `And` (not normalized) since the
     /// short-circuit vs. eager evaluation distinction is real and
-    /// externally-visible in TwinCAT/CODESYS itself, unlike e.g.
-    /// `REFERENCE TO`/`POINTER TO` which really are behaviorally
-    /// identical to `REF_TO`.
+    /// externally-visible in TwinCAT/CODESYS itself.
     AndThen,
     Eq,
     Ne,
--- a/compiler/dsl/src/textual.rs
+++ b/compiler/dsl/src/textual.rs
@@ -607,6 +607,13 @@ pub enum CompareOp {
     Or,
     Xor,
     And,
+    /// CODESYS/TwinCAT short-circuit `AND` (Beckhoff/CODESYS extension).
+    /// Kept as a distinct variant from `And` (not normalized) since the
+    /// short-circuit vs. eager evaluation distinction is real and
+    /// externally-visible in TwinCAT/CODESYS itself, unlike e.g.
+    /// `REFERENCE TO`/`POINTER TO` which really are behaviorally
+    /// identical to `REF_TO`.
+    AndThen,
     Eq,
     Ne,
     Lt,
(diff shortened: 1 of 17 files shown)


garretfick:
Please don't stack this PR. I would like to review this independently.

thusser:
Un-stacked this from the rest of the chain — it's now based directly on `main`, not on the explicit-enum-values/sized-string/mixed-located/var-initializer-expressions/PI branches. Ready for independent review whenever convenient.

compiler/dsl/src/textual.rs
```diff
@@ -607,6 +607,13 @@ pub enum CompareOp {
     Or,
     Xor,
     And,
+    /// CODESYS/TwinCAT short-circuit `AND` (Beckhoff/CODESYS extension).
+    /// Kept as a distinct variant from `And` (not normalized) since the
+    /// short-circuit vs. eager evaluation distinction is real and
+    /// externally-visible in TwinCAT/CODESYS itself, unlike e.g.
+    /// `REFERENCE TO`/`POINTER TO` which really are behaviorally
```
garretfick: Let's remove discussion of REF TO. That's it really relevant for this code

thusser: Done — removed the REFERENCE TO/POINTER TO comparison from the doc comment (and the same aside in the plan doc). Pushed in d449f189.

PR #1229, opened 2026-07-24
docs(adr): ADR-0040 diagnose dialect violations in a policy phase

garretfick:
Records the decision that dialect and feature-flag violations are
diagnosed outside the PEG grammar, never by making grammar acceptance
conditional on CompilerOptions. The grammar stays maximal and
option-free; disabled features are rejected in a phase that can name the
feature, with a dedicated per-feature problem code instead of a generic
P0002 SyntaxError.

Motivated by the TwinCAT dialect series (#1199), where structural
extensions with no distinguishing token (#1220 initializer expressions,
#1221 mixed located vars) had to parse permissively into AST provenance
markers and reject far downstream in the analyzer.

Claude-Session: https://claude.ai/code/session_01VCnJyKxgSfE4vL9dfiYsrZ

### DIFF
--- a/specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
+++ b/specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
@@ -1,0 +1,242 @@
+# ADR-0040: Dialect Violations Are Diagnosed in a Policy Phase, Not by Gating the Grammar
+
+status: proposed
+date: 2026-07-24
+
+## Context and Problem Statement
+
+IronPLC accepts Structured Text under several configurations — the IEC 61131-3
+editions and the vendor-compatible `rusty`/`codesys` dialects — expressed as
+[`Dialect`] presets plus per-feature `--allow-*` flags on [`CompilerOptions`]
+(ADR-0036, ADR-0038). A recurring question is *where* the compiler should reject
+a construct that is syntactically recognizable but disabled by the current
+dialect, and *what* it should report when it does.
+
+The parser is deliberately option-free. `parse_library`/`parse_statements`
+(`compiler/parser/src/parser.rs`) take only tokens; [`CompilerOptions`] never
+reaches the PEG grammar. Instead, options are "compiled away" into the token
+stream *before* parsing, inside `tokenize_program`
+(`compiler/parser/src/lib.rs`). Today there are three sanctioned mechanisms for
+making a feature conditional, documented in
+[`syntax-support-guide.md`](../steering/syntax-support-guide.md):
+
+1. **Token demotion** (`xform_demote_*.rs`) — a keyword token is rewritten to
+   `Identifier` when its flag is off, so the vendor grammar path structurally
+   cannot match and the word remains usable as an ordinary identifier.
+2. **Token rejection rule** (`rule_token_no_*.rs`) — the lexer always tokenizes
+   the construct; a token-level rule with access to options emits a diagnostic
+   when the flag is off.
+3. **Token transform** (`xform_tokens.rs`) — fix up the stream (e.g. missing
+   semicolons).
+
+All three require a *distinguishing token*. This breaks down for **structural
+extensions**, where the vendor grammar overlaps the standard grammar with
+nothing to demote — a constant *expression* where the standard allows only a
+literal initializer ([PR #1220]), or an `AT`-located variable inside an ordinary
+`VAR` block ([PR #1221]). For these, the TwinCAT series ([#1199]) reached for a
+fourth, undocumented pattern: parse permissively into a placeholder or marker
+AST node (`InitialValueAssignmentKind::SimpleExpr`, `in_mixed_var_block`), then
+reject far downstream in the analyzer. This is the "uncomfortable decision"
+that prompted the ADR — it smears the "is this legal in this dialect?" question
+across the parser, an AST provenance marker, and an analyzer rule.
+
+A related symptom compounds it. The parser itself emits exactly one problem
+code:
+
+```
+P0002  SyntaxError  "Syntax error"
+```
+
+Whenever the grammar simply fails to match — which is precisely what token
+demotion produces when a flag is off (the demoted word derails the parse
+*somewhere else*) — the user gets a generic `P0002`, often pointing at the
+wrong span. Meanwhile the token-rule and analyzer-rule layers already emit
+*specific*, feature-named codes following a consistent convention:
+
+```
+P0004  CStyleComment                "C-style comment not allowed"
+P0011  EmptyVarBlock                "Empty variable block requires --allow-empty-var-blocks flag"
+P4028  TopLevelVarGlobalNotAllowed  "Top-level VAR_GLOBAL requires --allow-top-level-var-global flag"
+P4033  PartialAccessSyntaxDisabled  "Partial-access syntax (.%Xn) requires --allow-partial-access-syntax flag"
+```
+
+Two questions this ADR settles:
+
+1. **Should dialect gating move *into* the grammar** (give the PEG parser
+   [`CompilerOptions`] and guard rules on flags), or stay in a phase outside the
+   grammar?
+2. **How does a disabled feature get a specific problem code** instead of a
+   generic `P0002 SyntaxError`?
+
+## Decision Drivers
+
+* **One maximal grammar to reason about** — the grammar should describe the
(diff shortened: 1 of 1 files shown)


specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
```diff
@@ -0,0 +1,242 @@
+# ADR-0040: Dialect Violations Are Diagnosed in a Policy Phase, Not by Gating the Grammar
+
+status: proposed
+date: 2026-07-24
+
+## Context and Problem Statement
+
+IronPLC accepts Structured Text under several configurations — the IEC 61131-3
+editions and the vendor-compatible `rusty`/`codesys` dialects — expressed as
```
garretfick: We also support twincatm we desire to support more

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
```diff
@@ -0,0 +1,242 @@
...
+(ADR-0036, ADR-0038). A recurring question is *where* the compiler should reject
+a construct that is syntactically recognizable but disabled by the current
+dialect, and *what* it should report when it does.
+
+The parser is deliberately option-free. `parse_library`/`parse_statements`
+(`compiler/parser/src/parser.rs`) take only tokens; [`CompilerOptions`] never
+reaches the PEG grammar. Instead, options are "compiled away" into the token
+stream *before* parsing, inside `tokenize_program`
+(`compiler/parser/src/lib.rs`). Today there are three sanctioned mechanisms for
+making a feature conditional, documented in
```
garretfick: I think we have some checks that reject after parsing

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
```diff
@@ -0,0 +1,242 @@
...
+
+1. **Should dialect gating move *into* the grammar** (give the PEG parser
+   [`CompilerOptions`] and guard rules on flags), or stay in a phase outside the
+   grammar?
+2. **How does a disabled feature get a specific problem code** instead of a
+   generic `P0002 SyntaxError`?
+
+## Decision Drivers
+
+* **One maximal grammar to reason about** — the grammar should describe the
```
garretfick, line 88: One other driver is keeping code analysis and generation simple

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
```diff
@@ -0,0 +1,242 @@
...
+  grammar, at the earliest point that has both options and enough context to
+  name the feature, emitting a dedicated `P####` per feature.
+* **Status quo** — continue case-by-case: demotion where a token distinguishes
+  the feature, and parse-permissive-plus-AST-marker-plus-analyzer-rule for
+  structural cases, with no stated rule for which to use or where the reject
+  belongs.
+
+## Decision Outcome
+
+Chosen option: **Diagnose in a policy phase with per-feature codes.**
```
garretfick, line 134: I think this is going too far. I'm ok with token demotion - we can up with that to address ambiguous syntax that we couldn't otherwise easily parse

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
```diff
@@ -0,0 +1,242 @@
...
+
+| The disabled feature is distinguishable by… | Mechanism | AST marker |
+| --- | --- | --- |
+| a single token (`//`, `.%Xn`, a keyword being present) | token rejection rule (`rule_token_no_*`), pre-parse, has options | none |
+| structure only (located var in a plain block, expression where a literal is required) | a dialect-policy pass over the freshly parsed AST | avoid — see rule 4 |
+
+Prefer a **token rejection rule over demotion when the word is not a plausible
+identifier in standard code.** Demotion silently turns the keyword into an
+identifier and lets the parse fail elsewhere with a generic `P0002`; a rejection
+rule keeps the token and emits a specific code (e.g. "`REF_TO` requires
```
garretfick: But REF_RO could be used as an identifier in real code. We cannot assume it is never used as an identifier.

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
```diff
@@ -0,0 +1,242 @@
...
+`--allow-ref-to`"). Reserve *demotion* for words that genuinely double as
+identifiers in conforming programs (`AND_THEN`, `TIME`), where the trade — a
+specific error versus letting existing code use the word as a name — is real
+and resolves in favor of demotion.
+
+### 4. Prefer re-deriving structure over adding an AST marker
+
+The provenance markers in the structural cases (`in_mixed_var_block`,
```
garretfick, line 202: I think this is a key constraint. We collapsed var declarations to make code generation easier. I think we should have offered functions on the ast rather than collapse

garretfick (commented):
Please address comments

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
garretfick: Updated the context to include TwinCAT (noted as CODESYS-derived) and to state that the dialect set is expected to grow, so the gating mechanism has to scale to many dialects.

---

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
garretfick: Right — the analyzer already rejects option-gated constructs after parsing (`rule_ref_to.rs` gates on the `--allow-ref-*` flags, for instance). I reworked this section to list mechanisms at two points in the pipeline (pre-parse token-level and post-parse over the AST) and to state explicitly that "reject after parsing" is an established mechanism the ADR builds on, not something new. The genuinely-new/uncomfortable part is narrowed to smuggling parse-path provenance through an AST marker field.

---

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
garretfick: Added as a decision driver: "Keep code analysis and generation simple — downstream stages should see a clean, uniform AST and not branch on dialect provenance." I also tied it to rule 4, since keeping codegen simple is exactly what motivated collapsing the declaration shapes in the first place.

---

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
garretfick: Agreed — I walked this back. Rule 3 now states that token demotion stays a first-class mechanism, kept precisely because it handles ambiguous/hard-to-parse syntax, and is not something to migrate away from. The "prefer a rejection rule over demotion" language is gone. Token rejection rules are now scoped only to syntax that can't be confused with an identifier (`//`, `.%Xn`), where nothing legitimate is lost by rejecting.

---

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
garretfick: Correct, and I removed the wrong claim. The ADR now says explicitly: "we cannot assume any keyword is never used as an identifier — `REF_TO`, `AND_THEN`, `TIME`, and the like can all legitimately appear as variable or type names." That's exactly why `REF_TO` belongs in the demotion column. I also added a carve-out to rule 2: because a demoted keyword is deliberately made identifier-shaped when the flag is off, a program that meant the keyword is genuinely ambiguous input, so a `P0002`/undeclared-identifier result there is expected rather than a defect — the accepted cost of demotion.

---

specs/adrs/0040-dialect-violations-diagnosed-in-policy-phase.md
garretfick: This shaped rule 4 directly. It's rewritten to "Expose structure through AST accessors, rather than collapsing it or marking it" — the point being that collapsing the declarations and then re-adding a marker are two halves of the same mistake. The recommended shape is to keep the AST faithful and offer accessor functions (e.g. "is this declaration located?", "does this block mix located and plain declarations?") that analysis, codegen, and the policy check all call — so codegen stays simple without the AST being lossy, and no provenance flag is threaded through the DSL. A marker field is the last resort.

---

PR #1236, opened 2026-07-25
reference-to-twincat.md with REQ-RTO markers (PR1)

garretfick:
Author the single design document for TwinCAT REFERENCE TO support with
all PR-1 REQ-RTO-<slug>-NNN markers (parser/analyzer/codegen/plc2plc) and
the requirements->test traceability table. Supersedes
beckhoff-twincat-dialect.md 2.1/3.6.

Claude-Session: https://claude.ai/code/session_01Xm88yRD7YfLG9RtAvEVxfv

### DIFF
--- a/compiler/parser/src/options.rs
+++ b/compiler/parser/src/options.rs
@@ -263,7 +263,7 @@ define_compiler_options! {
 
     "Allow Beckhoff TwinCAT/CODESYS REFERENCE TO reference types and the REF= binding operator",
     "--allow-reference-to",
-    [Codesys],
+    [Codesys, TwinCat],
     allow_reference_to,
 
     "Allow arithmetic (+, -) and ordering comparisons (<, >, <=, >=) on REF_TO types",
@@ -471,13 +471,13 @@ mod tests {
 
     /// The TwinCAT dialect is close to CODESYS (TwinCAT 3 runs on the CODESYS
     /// V3 runtime) but does *not* enable the `REF_TO` reference extensions.
-    /// TwinCAT spells references and pointers `REFERENCE TO` / `POINTER TO`
-    /// (not the CODESYS `REF_TO` / `REF()` / `NULL`), and IronPLC does not
-    /// parse those yet, so none of `allow_ref_to`, `allow_ref_arithmetic`,
-    /// `allow_ref_stack_variables`, or `allow_ref_type_punning` are enabled --
-    /// enabling them would accept `REF_TO` code that TwinCAT itself rejects.
-    /// Listed explicitly so an accidental divergence from the intended set is
-    /// caught.
+    /// TwinCAT spells references `REFERENCE TO` (not the CODESYS `REF_TO` /
+    /// `REF()` / `NULL`), so it enables `allow_reference_to` instead, and none
+    /// of `allow_ref_to`, `allow_ref_arithmetic`, `allow_ref_stack_variables`,
+    /// or `allow_ref_type_punning` are enabled -- enabling those would accept
+    /// `REF_TO` code that TwinCAT itself rejects. (Pointer types `POINTER TO`
+    /// with `ADR()` are not parsed yet.) Listed explicitly so an accidental
+    /// divergence from the intended set is caught.
     #[test]
     fn twincat_dialect_enables_exactly_these_vendor_flags() {
         assert!(!CompilerOptions::from_dialect(Dialect::TwinCat).allow_iec_61131_3_2013);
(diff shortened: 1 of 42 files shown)


compiler/parser/src/options.rs
```diff
@@ -245,6 +245,11 @@ define_compiler_options! {
     [Rusty, Codesys],
     allow_ref_to,
 
+    "Allow Beckhoff TwinCAT/CODESYS REFERENCE TO reference types and the REF= binding operator",
+    "--allow-reference-to",
+    [Codesys],
```
garretfick: This should also have TwinCat

compiler/parser/src/parser.rs
```diff
@@ -858,17 +859,32 @@ parser! {
...
+      tok(TokenType::RefTo) { RefSyntax::RefTo }
+      / tok(TokenType::Reference) _ tok(TokenType::To) { RefSyntax::ReferenceTo }
+    // Matches the TwinCAT/CODESYS `REF=` binding operator, returning the `=`
+    // token (used for the assignment span). `REF` may be the `Ref` keyword
+    // token (when `--allow-ref-to`/Edition 3 is also active) or a demoted
+    // identifier `REF` (when only `--allow-reference-to` is set); both are
+    // accepted, case-insensitively.
+    rule ref_bind_op() -> &'input Token =
+      tok(TokenType::Ref) _ eq:tok(TokenType::Equal) { eq }
+      / [t if t.token_type == TokenType::Identifier && t.text.eq_ignore_ascii_case("REF")] _ eq:tok(TokenType::Equal) { eq }
```
garretfick: 1. This is allowing a space between `REF` and `=`. I don't think that's allowed.
2. This ignores the case but from docs I don't see a reason to ignore case sensitivity.

Should this be a new token that matches `REF=`. Having that would also make for nicer token output, however, that could also be ambiguous and prevent matching other tokens.

docs/reference/compiler/ironplcc.rst
```diff
@@ -133,6 +133,11 @@ Options
    Edition 3. This is a vendor extension useful when you need references
    but want to keep Edition 2 keyword handling for the rest of your code.
 
+``--allow-reference-to``
+   Allow the Beckhoff TwinCAT / CODESYS ``REFERENCE TO`` reference type and the
+   ``REF=`` binding operator — the TwinCAT/CODESYS-facing alternative to
```
garretfick, line 138: We need to correct which dialects enable this.

compiler/parser/src/parser.rs
garretfick: Fixed the space: `REF` and `=` must now be adjacent (removed the `_` between them), so `x REF = y` is rejected — added a regression test for that.

On case: I kept it case-insensitive on purpose. The `REF` half is the `Ref` keyword token, which the lexer already produces case-insensitively (`#[token("REF", ignore(case))]`), as it does for every keyword — IEC 61131-3 is a case-insensitive language. Making only `REF=` case-sensitive would be inconsistent (`ref=` would fail while `ref_to`, `if`, etc. all work). Happy to flip it if you'd prefer strict casing.

On a dedicated `REF=` token: I looked at it but left it as a parser rule because a single `REF=` lexer token has the ambiguity you flagged — it would also match `REF=` when a program has a variable literally named `REF` used in an adjacent equality (e.g. `x := (REF=5)`), breaking that in standard mode unless we split the token back apart in a transform. The parser-rule approach keeps `REF` an ordinary identifier everywhere and only forms the binding in statement position. If you'd still like the dedicated token for nicer token output, I'll add it with the demotion/split to keep standard mode safe — just let me know.

---

compiler/parser/src/options.rs
garretfick: Added a `twincat` dialect. Since TwinCAT is CODESYS-derived, it enables the same vendor extensions as `codesys` and additionally `allow_reference_to`, so `REFERENCE TO` is now bundled by both `twincat` and `codesys`. Wired it through the CLI `--dialect` value enum (the drift-guard test covers it) and MCP/playground pick it up via `Dialect::ALL`; there's a `twincat_dialect_enables_exactly_these_vendor_flags` test asserting the set. If you'd rather TwinCAT enable a narrower/different set than CODESYS, point me at the intended flags and I'll adjust.

---

docs/reference/compiler/ironplcc.rst
garretfick: Corrected — the docs (this page, `enabling-dialects-and-features.rst`, `reference-types.rst`, and the steering flag table) now say `REFERENCE TO` is enabled by the `twincat` and `codesys` dialects, and the new `twincat` dialect is documented alongside them.

---

PR #1242, opened 2026-07-26
docs(adr): ADR-0041 staged method/property dispatch and interface values

thusser:
## Summary

Proposes an architecture direction for the next piece of the CODESYS/TwinCAT OOP vendor extension work: real `METHOD`/`PROPERTY` dispatch and interface values. This PR is the ADR only — no implementation code, following the same pattern as ADR-0040.

Everything up through static shape-checking (`EXTENDS` field/method inheritance, `IMPLEMENTS` conformance against an interface's required members) is already merged/in review elsewhere. None of it has runtime semantics yet — every OOP-using `FUNCTION_BLOCK` is still uniformly `P9004`-gated. This ADR is about the mechanism for eventually lifting that gate, not the gate itself.

## Key proposal

Split "OOP dispatch" into two independently-deliverable phases, since they're genuinely different in size and risk:

1. **Static dispatch** — calls where the concrete type is known at compile time (direct instance calls, including a method overridden via `EXTENDS`), plus `THIS^`/`SUPER^` and `PROPERTY` accessor rewriting. No function block memory-layout change, no new runtime indirection — reuses the exact traversal already built for `IMPLEMENTS` conformance checking.
2. **Dynamic dispatch** — calls through a `REFERENCE TO`/`POINTER TO` a base type used polymorphically, or through an interface-typed value. Needs a per-instance runtime type tag and a compile-time-built dispatch table, added *only* to function block types actually used polymorphically (not universally on every instance).

Grounded directly in the current codegen before proposing anything: `compile_fb_call` (`compiler/codegen/src/compile_stmt.rs`) confirmed to resolve call targets statically via a compile-time-only `type_id` today, with zero per-instance runtime type information anywhere in the existing memory layout.

Happy to adjust the phasing or the Phase 2 mechanism sketch based on feedback — the intent here is to settle the overall direction before either phase gets an implementation plan.

### DIFF
--- a/specs/adrs/0041-staged-method-and-interface-dispatch.md
+++ b/specs/adrs/0041-staged-method-and-interface-dispatch.md
@@ -1,0 +1,229 @@
+# Staged Method/Property Dispatch and Interface Values
+
+status: proposed
+date: 2026-07-27
+
+## Context and Problem Statement
+
+The CODESYS/TwinCAT OOP vendor extensions (`EXTENDS`, `IMPLEMENTS`,
+`METHOD`, `PROPERTY`, `THIS^`/`SUPER^`) parse fully and are checked for
+static shape (field inheritance, `IMPLEMENTS` conformance against an
+interface's required members). None of this has runtime semantics yet
+-- every `FUNCTION_BLOCK` using any of these constructs is still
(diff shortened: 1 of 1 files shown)


garretfick:
I modified some content in this so had to create a new PR.

PR #1243, opened 2026-07-27
feat(twincat): accept hex/binary/octal bit-string literals as CASE labels

thusser:
## Summary

- Fixes `16#D012:`/`2#1010:`-style `CASE` labels failing to parse with `P0002` — the lexer already tokenizes hex/binary/octal bit-string literals correctly, but `case_list_element()` never accepted them as a label. Adds `CaseSelectionKind::BitStringLiteral` as a last-resort alternative, deliberately narrower than the general `bit_string_literal()` rule to avoid a PEG hazard with plain decimal labels.
- Gets full codegen support, not a stub — a plain literal value has no runtime-computation ambiguity, reusing the exact `u32`/`u64` narrowing already used for `ConstantKind::BitStringLiteral`. End-to-end tests confirm correct *execution* (matching the right `CASE` arm), not just successful compilation.
- Found and documented a pre-existing, unrelated round-trip quirk while testing: bit-string literals already render decimalized everywhere in this codebase, so re-parsing a rendered label resolves to `SignedInteger`, not `BitStringLiteral` again. Not fixed here (out of scope — would mean changing bit-string rendering everywhere, not just `CASE` labels); the plc2plc test asserts render *idempotency* instead of AST equality, matching the existing pattern for this exact situation.

## Test plan
- [x] Parser: hex/binary/octal `CASE` labels parse as `BitStringLiteral`; plain decimal labels are unaffected (regression)
- [x] End-to-end codegen: correct `CASE` arm executes for a bit-string-literal label
- [x] plc2plc: render idempotency for the pre-existing decimalization quirk

### DIFF
--- a/compiler/codegen/src/compile_stmt.rs
+++ b/compiler/codegen/src/compile_stmt.rs
@@ -696,6 +696,7 @@ fn compile_case(
 /// - `SignedInteger`: `selector == value`
 /// - `Subrange`: `(selector >= start) AND (selector <= end)`
 /// - `EnumeratedValue`: not yet supported (returns todo diagnostic)
+/// - `BitStringLiteral`: `selector == value` (same shape as `SignedInteger`)
 fn compile_case_selector(
     emitter: &mut Emitter,
     ctx: &mut CompileContext,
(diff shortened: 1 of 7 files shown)


garretfick (commented):
Thanks for this — the codegen approach (reusing the `SignedInteger` shape) is clean. But I don't think this is standards compliant as written, on two counts.

**1. The label form isn't IEC 61131-3.** The standard grammar for a case label (Annex B, unchanged Ed 2 → Ed 3) is:

```
case_list_element ::= subrange | signed_integer | enumerated_value
```

where `signed_integer ::= ['+'|'-'] integer` and `integer` is a **decimal** digit sequence. Hex/binary/octal literals are separate productions (`hex_integer` / `binary_integer` / `octal_integer`) that the standard groups under `bit_string_literal`, and those are deliberately *not* in `case_list_element`. So `16#D012:` / `2#1010:` as a CASE label is a vendor extension (as the `feat(twincat)` title and #1199 acknowledge), not standard syntax. The pre-existing `case_list_element()` matched the standard exactly.

**2. It's added ungated to the default path — this is the blocking issue.** The new alternative is unconditional; there's no `options.allow_*` check anywhere in the diff:

```rust
rule case_list_element() = ... / bsl:case_bit_string_literal() { CaseSelectionKind::BitStringLiteral(bsl) }
```

Per `specs/steering/syntax-support-guide.md`:
> - Line 20: **Non-standard gating** — If not standard IEC 61131-3, gate behind `--allow-x` flag
> - Line 146: **Rule**: Anything not in the IEC 61131-3 standard **must** be gated behind an `--allow-x` flag.

So "no new flag" in the summary is actually the defect, not a simplification. As-is, the **default** dialect (`iec61131-3-ed2`) and strict Ed 3 will now silently accept these labels, when a conformant parser must reject them. This should sit behind an `--allow-*` flag (a new one, or grouped with an existing vendor flag), enabled only for the TwinCAT/CODESYS/RuSTy dialect presets, with the LSP + docs wiring from the checklist at lines 20–22.

**Minor, related:** the tests select on `x : DWORD` (`CASE x OF`). The standard requires the CASE selector to be `ANY_INT` or an enumerated type; `DWORD` is a bit-string (`ANY_BIT`) type, so that's outside the standard too — worth using an integer selector in the tests, or covering the bit-string-selector case explicitly under the same flag.

(Also flagging that the branch currently shows a merge conflict against `main` — unrelated to the above.)

Happy to be corrected if I'm working from an out-of-date copy of the grammar, but I believe the label-form restriction holds in the current edition.

---

garretfick:
Implemented

PR #1244, opened 2026-07-27
feat(twincat): allow general expressions as struct/FB-instance init values

thusser:
## Summary

- Fixes `tonDelta : TON := (PT := pDevice^.Delta);` failing to parse — a pointer dereference plus member access as a call-style/struct-init value, which is a genuine runtime expression, not a compile-time constant. Adds `StructInitialValueAssignmentKind::Expression(Expr)`, tried as a last-resort alternative in `structure_element_initialization()` after `constant()`/`enumerated_value()`/`array_initialization()`/`structure_initialization()`.
- Not the same shape as `--allow-constant-initializer-expressions` — this value is **not** required to fold to a compile-time constant, matching how FB call-style construction params already accept arbitrary runtime expressions.
- Hit a PEG ordered-choice hazard along the way: `constant()`/`enumerated_value()` are a strict grammatical prefix of `expression()`, so simply appending `expression()` last wasn't enough — `enumerated_value()` matched just the leading identifier and left the rest unconsumed. Fixed with a trailing lookahead requiring those two alternatives to be immediately followed by the list terminator (`,`/`)`) to win the choice.
- Codegen returns `not_implemented` (`P9999`) for the new variant, same precedent already used elsewhere in this series: `ironplcc check` fully supports it, `ironplcc compile` fails clearly instead of silently skipping the field.

## Test plan
- [x] Parser: the motivating shape parses as `StructInitialValueAssignmentKind::Expression`
- [x] Regression: the standard `:= (member := value)` named-struct-init form still parses unchanged
- [x] plc2plc: round-trips through render/re-parse
- [x] End-to-end codegen: `P9999 not_implemented` for the new variant

### DIFF
--- a/compiler/codegen/src/compile_struct.rs
+++ b/compiler/codegen/src/compile_struct.rs
@@ -405,6 +405,16 @@ fn compile_struct_field_init(
             // Nested structures and arrays in struct init are not yet supported.
             Ok(())
         }
+        StructInitialValueAssignmentKind::Expression(expr) => {
+            // A general (possibly non-constant) expression, e.g.
+            // `pDevice^.Delta` -- `ironplcc check` fully supports this;
+            // codegen does not yet implement evaluating it at instance
+            // construction time.
+            Err(Diagnostic::not_implemented(Label::span(
+                expr.span(),
+                "Expression-valued struct/FB-instance field initializer",
+            )))
+        }
     }
 }
 
--- a/compiler/codegen/tests/it/end_to_end_struct.rs
+++ b/compiler/codegen/tests/it/end_to_end_struct.rs
@@ -1,7 +1,7 @@
 //! End-to-end integration tests for structure field read support.
 //! Compiles ST programs with struct field access and runs them through the VM.
 
-use crate::common::parse_and_run;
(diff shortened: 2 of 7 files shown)


garretfick:
Superseded by #1276. This branch lives on a fork and had merge conflicts against `main` (the parser/plc2plc test files were split into per-feature modules by #1261), so the work was rebased onto current `main` on a branch in this repo.

Two changes from this PR in #1276:
- The two new tests are relocated into dedicated `tests/struct_init_expressions.rs` modules to match the #1261 split.
- The feature is now gated behind a new `--allow-struct-initializer-expressions` dialect flag (with problem code P4043 when off), rather than accepted unconditionally — expression-valued struct/FB initializers aren't standard IEC 61131-3, so the strict default dialect should keep rejecting them like every other vendor extension.

This one can be closed in favor of #1276.

---

thusser:
Superseded by #1276, which rebases this onto current main (after the #1261 test-file split) and adds the dialect-flag gating (--allow-struct-initializer-expressions, P4043) that this PR was missing. Closing in favor of that.

PR #1245, opened 2026-07-27
feat(twincat): allow implicit REAL to LREAL widening

thusser:
## Summary

- `ElementaryTypeName::can_widen_to()` had no case at all for `(Real, Real)` — a typed `REAL` expression (a variable, field, or return value, not a bare literal) could never widen to `LREAL`, even though it's always lossless (32-bit float → 64-bit float) and IEC 61131-3 treats it the same as any other same-family widening. The untyped-literal case (`3.14` inferred as `ANY_REAL`, per ADR-0028) already worked via a separate code path — this is specifically about typed expressions.
- Fixed with a single match arm: `(Real, Real) => tgt_bits > src_bits`, same shape as the existing integer-chain arms. No new dialect flag — this is standard, lossless widening, not a vendor extension, so it doesn't belong behind `allow_cross_family_widening` (that flag is specifically for genuinely *cross-family* widening).
- `LREAL` → `REAL` narrowing remains disallowed, confirmed by a dedicated regression test.
- This unblocks real function calls to Beckhoff library functions declaring an `LREAL` parameter when the argument is a typed `REAL` variable rather than a literal.

## Test plan
- [x] Typed `REAL` expression widens cleanly to an `LREAL` parameter/assignment
- [x] Regression: `LREAL` → `REAL` narrowing is still rejected


compiler/analyzer/src/rule_function_call_type_check.rs
```diff
@@ -735,6 +735,59 @@ END_PROGRAM";
         assert!(result.is_ok());
     }
 
+    #[test]
```
garretfick, line 738: this is good but we need to update the docs here:

https://www.ironplc.com/explanation/type-conversions.html

garretfick (changes requested):

thusser:
Added a "Real widening" section to `type-conversions.rst` documenting REAL → LREAL (and noting the reverse direction still requires `LREAL_TO_REAL`).

garretfick (approved):

PR #1246, opened 2026-07-27
feat(twincat): register BOOL_TO_STRING/LREAL_TO_FMTSTR/ADR stdlib functions

thusser:
## Summary

Standalone PR, no dependency on any other open PR. Adds two new flags (`--allow-address-operator`, `--allow-extended-string-functions`).

- `BOOL_TO_STRING` joins the existing, unconditional `TO_STRING` conversion family — this is a standard IEC 61131-3 conversion, not a separate vendor library function, so no new flag. Found and fixed a real codegen hazard along the way: `BOOL` and `DINT` resolve to the identical `VarTypeInfo` shape, so without a dedicated `StringConversion::BoolToString` variant, `BOOL_TO_STRING` would have silently compiled via the generic integer path, producing `"1"`/`"0"` instead of the documented `"TRUE"`/`"FALSE"` — wrong output that compiles cleanly, strictly worse than a clean rejection. Now returns `not_implemented` instead.
- `LREAL_TO_FMTSTR` (Beckhoff `Tc2_Utilities`) and `ADR` (the Address Operator) are new stdlib registrations behind two new flags, `allow_extended_string_functions` and `allow_address_operator`, with signatures verified against Beckhoff's own docs (`LREAL_TO_FMTSTR(in : LREAL, iPrecision : INT, bRound : BOOL) : STRING(510)`; `ADR(IN) : DWORD` — `PVOID` is Beckhoff's recommended portable return type, but IronPLC has no `PVOID` type, so `DWORD` is used as a known simplification).
- All three get `not_implemented` (`P9999`) codegen stubs: `ironplcc check` fully supports all three (the actual motivating "undeclared function" gap from a corpus-check pass); `ironplcc compile` fails clearly instead of computing a wrong address or silently-wrong formatted output.

## Test plan
- [x] `BOOL_TO_STRING` is unconditional and produces `not_implemented` at codegen (not silently wrong output)
- [x] `ADR`/`LREAL_TO_FMTSTR` are gated correctly by their new flags
- [x] End-to-end: both new functions resolve and produce `P9999` stubs, not silently-wrong codegen

### DIFF
--- a/compiler/analyzer/src/intermediates/stdlib_function.rs
+++ b/compiler/analyzer/src/intermediates/stdlib_function.rs
@@ -331,8 +331,8 @@ const STRING_TO_INT_TYPES: &[&str] = &["SINT", "INT", "DINT", "USINT", "UINT", "
 ///
 /// Covers all W32 numeric types: signed integers (SINT, INT, DINT),
 /// unsigned integers and bit-strings (USINT, UINT, UDINT, BYTE, WORD, DWORD),
-/// and REAL. Each type gets a *_TO_STRING function, and a subset gets
-/// STRING_TO_* functions.
+/// REAL, and BOOL. Each type gets a *_TO_STRING function, and a subset
+/// gets STRING_TO_* functions.
 fn get_string_conversion_functions() -> Vec<FunctionSignature> {
     let mut functions = Vec::new();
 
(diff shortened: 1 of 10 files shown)


garretfick:
See my comment on PI. Can we separate this into what's part of the standard and what's vendor extension. the vendor extension would be included as a "library"

thusser:
BOOL_TO_STRING, LREAL_TO_FMTSTR, and ADR are all now covered by the compatibility-library mechanism (Tc2_BuiltIns, Tc2_Utilities) and the new ADR()/POINTER TO language feature. Closing in favor of that, see comment on #1199.

PR #1250, opened 2026-07-28
feat(parser): add TwinCAT dialect preset

garretfick:
Add a `twincat` dialect that enables the vendor extensions needed to
parse Beckhoff TwinCAT code. TwinCAT 3 is built on the CODESYS V3
runtime, so the preset enables the same 16 vendor-extension flags as the
`codesys` dialect (Edition 2 base with REF_TO and the shared
CODESYS/TwinCAT extensions), and likewise does not pre-bind the implicit
`__SYSTEM_UP_TIME` globals, which are an IronPLC runtime convention.

Wire the new dialect through the CLI, LSP, docs, steering guide, and the
VS Code extension setting. The MCP list_options tool and LSP option
extraction are already data-driven off `Dialect::ALL`/`FromStr`, so they
pick up the new dialect with only added tests.

Claude-Session: https://claude.ai/code/session_01VWtfXQdCfFqjZChuXkGNqg


compiler/codegen/tests/it/end_to_end_dialect.rs
```diff
@@ -139,6 +139,61 @@ END_PROGRAM
...
+    let (_c, bufs) = parse_and_run(source, &CompilerOptions::from_dialect(Dialect::TwinCat));
+    // TwinCAT does not pre-bind __SYSTEM_UP_TIME/__SYSTEM_UP_LTIME, so the
+    // user variables start at index 0: LDT=0, result=1.
+    assert_eq!(bufs.vars[1].as_i32(), 42);
+}
+
+#[test]
+fn end_to_end_when_twincat_dialect_then_ref_to_works() {
+    // The TwinCAT dialect enables REF_TO.
+    let source = "
```
garretfick, line 166: I think this is wrong. Twincat uses REFERENCE TO

compiler/ironplc-cli/src/lsp.rs
```diff
@@ -913,6 +913,34 @@ mod test {
         assert!(!options.allow_system_uptime_global);
     }
 
+    #[test]
+    fn extract_compiler_options_when_twincat_dialect_then_enables_ref_to_without_uptime_global() {
```
garretfick: This test name doesn't scale as we add more options and dialects. Let's drop the ending part that mentions each option even if that doesn't follow our BDD naming convention

compiler/parser/src/options.rs
```diff
@@ -212,62 +224,62 @@ macro_rules! define_compiler_options {
...
     "Allow REF_TO, REF(), and NULL without full Edition 3",
     "--allow-ref-to",
-    [Rusty, Codesys],
+    [Rusty, Codesys, TwinCat],
     allow_ref_to,
 
     "Allow arithmetic (+, -) and ordering comparisons (<, >, <=, >=) on REF_TO types",
     "--allow-ref-arithmetic",
-    [Rusty, Codesys],
+    [Rusty, Codesys, TwinCat],
```
garretfick: I don't think this is right

compiler/parser/src/options.rs
```diff
@@ -212,62 +224,62 @@ macro_rules! define_compiler_options {
...
     "Allow REF() on stack-allocated variables (VAR_TEMP, FUNCTION VAR_INPUT/VAR_OUTPUT)",
     "--allow-ref-stack-variables",
-    [Rusty, Codesys],
+    [Rusty, Codesys, TwinCat],
     allow_ref_stack_variables,
 
     "Allow assigning between REF_TO types of different base types (type punning)",
     "--allow-ref-type-punning",
-    [Rusty, Codesys],
+    [Rusty, Codesys, TwinCat],
```
garretfick: I also think this isn't right

docs/explanation/enabling-dialects-and-features.rst
```diff
@@ -41,6 +41,19 @@ Supported Dialects
    globals are not pre-bound under this dialect, since they are an IronPLC
    runtime convention rather than a CODESYS feature.
 
+**twincat**
+   Beckhoff TwinCAT-compatible dialect. TwinCAT 3 is built on the CODESYS V3
+   runtime, so this enables the same feature set as the ``codesys`` dialect:
+   an Edition 2 base (so identifiers like
+   :doc:`LDT </reference/language/data-types/elementary/ldate-and-time>` are
+   preserved),
+   :doc:`REF_TO </reference/language/data-types/derived/reference-types>`
```
garretfick: Are we sure about this form?

integrations/vscode/package.json
```diff
@@ -99,26 +99,29 @@
           "editPresentation": "singlelineText"
         },
         "ironplc.dialect": {
-          "markdownDescription": "Language dialect preset. Controls IEC edition and vendor extensions.\n\n- `iec61131-3-ed2`: Strict IEC 61131-3:2003 (Edition 2)\n- `iec61131-3-ed3`: IEC 61131-3:2013 (Edition 3) with `LTIME`, `REF_TO`, etc.\n- `rusty`: RuSTy-compatible — Edition 2 base with `REF_TO` and all vendor extensions\n- `codesys`: CODESYS-compatible — Edition 2 base with `REF_TO` and CODESYS vendor extensions",
+          "markdownDescription": "Language dialect preset. Controls IEC edition and vendor extensions.\n\n- `iec61131-3-ed2`: Strict IEC 61131-3:2003 (Edition 2)\n- `iec61131-3-ed3`: IEC 61131-3:2013 (Edition 3) with `LTIME`, `REF_TO`, etc.\n- `rusty`: RuSTy-compatible — Edition 2 base with `REF_TO` and all vendor extensions\n- `codesys`: CODESYS-compatible — Edition 2 base with `REF_TO` and CODESYS vendor extensions\n- `twincat`: TwinCAT-compatible — Edition 2 base with `REF_TO` and TwinCAT vendor extensions",
```
garretfick: Same comment here

compiler/codegen/tests/it/end_to_end_dialect.rs
garretfick: Agreed — TwinCAT uses `REFERENCE TO`/`POINTER TO`, not the CODESYS/IEC `REF_TO` spelling (and `REFERENCE TO` isn't parsed yet anyway). Replaced this test with `end_to_end_when_twincat_dialect_then_pragmas_are_skipped`, which exercises a genuinely idiomatic, supported TwinCAT construct end-to-end. Fixed in aed146e.

---

compiler/ironplc-cli/src/lsp.rs
garretfick: Renamed to `extract_compiler_options_when_twincat_dialect` (dropped the per-option suffix). Fixed in aed146e.

---

compiler/parser/src/options.rs
garretfick: Right — TwinCAT's `REFERENCE TO` is a managed reference with no pointer arithmetic. Removed `TwinCat` from `allow_ref_arithmetic` (stays CODESYS-only), updated the exact-flags test, and documented the divergence. Fixed in aed146e.

---

compiler/parser/src/options.rs
garretfick: Same here — removed `TwinCat` from `allow_ref_type_punning` (managed references don't type-pun). Now CODESYS-only, and the exact-flags test asserts it's off for TwinCAT. Fixed in aed146e.

---

docs/explanation/enabling-dialects-and-features.rst
garretfick: Reworded to not headline `REF_TO` (TwinCAT uses `REFERENCE TO`/`POINTER TO`). The description now points to the extensions TwinCAT actually shares with CODESYS (pragmas, C-style comments, `AND_THEN`) and notes it's stricter on references. Fixed in aed146e.

---

integrations/vscode/package.json
garretfick: Updated the VS Code setting description the same way — dropped the `REF_TO` claim, now reads "the vendor extensions TwinCAT shares with CODESYS (stricter on references)". Fixed in aed146e.

---

garretfick:
Follow-up on the "don't enable syntax the env doesn't support" point: dropped the **entire `REF_TO` family** from the `twincat` preset, not just the two ref rules — `allow_ref_to`, `allow_ref_arithmetic`, `allow_ref_stack_variables`, and `allow_ref_type_punning` are now all CODESYS-only. TwinCAT spells references/pointers `REFERENCE TO` / `POINTER TO` (with `ADR()`), which IronPLC doesn't parse yet, so enabling `REF_TO` would have accepted code TwinCAT itself rejects.

Net `twincat` flag set is now the CODESYS set minus the `REF_TO` family and minus `allow_system_uptime_global`. Updated the exact-flags test, LSP assertions, docs, steering table, and VS Code setting to match. In `378740c` (rebased onto the latest branch head; full CI green).

---

PR #1251, opened 2026-07-28
fix(parser): enforce --allow-top-level-var-global (P4028)

### LINKED ISSUE #1233: `--allow-top-level-var-global` is not enforced: top-level VAR_GLOBAL compiles regardless of the flag
## Summary

The `allow_top_level_var_global` feature flag (`--allow-top-level-var-global`, MCP `allowTopLevelVarGlobal`) is declared, surfaced by `list_options`, exposed on the CLI/LSP, and documented — but it is **never enforced**. A `VAR_GLOBAL ... END_VAR` block at the top level of a file compiles cleanly whether the flag is on or off. Enabling the flag has no effect, and (more importantly) the construct that the flag is supposed to gate is accepted unconditionally in strict IEC 61131-3 mode.

## Evidence

- The reserved problem code **P4028 `TopLevelVarGlobalNotAllowed`** ("Top-level VAR_GLOBAL requires --allow-top-level-var-global flag") is declared in `compiler/problems/resources/problem-codes.csv:75` and has a doc page (`docs/reference/compiler/problems/P4028.rst`), but grepping the source shows it is **emitted nowhere** — the only non-generated occurrence is the CSV declaration.
- The parser accepts `VAR_GLOBAL` as a top-level element unconditionally (no `allow_top_level_var_global` check on the parse path).
- The `allow_top_level_var_global` field is only read in `#[cfg(test)]` parse helpers, never in enforcement.

## Reproduction

```st
VAR_GLOBAL CONSTANT
    X : INT := 250;
END_VAR
PROGRAM p
END_PROGRAM
```

- **Expected** (flag off, strict Ed2): rejected with P4028.
- **Actual**: compiles cleanly with `--dialect=iec61131-3-ed2` and no flag.

## Suggested fix

Emit P4028 during parse/analysis when a top-level `VAR_GLOBAL` is encountered and `allow_top_level_var_global` is false. See `parser/src/rule_token_no_c_style_comment.rs` / `parser/src/rule_no_empty_var_blocks.rs` for the validation-rule pattern used by comparable flags.

## Related

A behavioral fixture already records this gap in `compiler/mcp/src/feature_flag_conformance.rs` with `Expectation::NotYetEnforced`. When enforcement lands, that fixture will need to move to `Expectation::GatesBehavior` (its test asserts the current not-enforced state, so it will fail once the flag actually gates behavior — a built-in reminder).

garretfick:
The allow_top_level_var_global feature flag was declared, surfaced by
list_options, exposed on the CLI/LSP, and documented, but never enforced:
a top-level VAR_GLOBAL block compiled cleanly whether the flag was on or
off, so strict IEC 61131-3 mode accepted the vendor extension
unconditionally. Problem code P4028 (TopLevelVarGlobalNotAllowed) was
reserved and documented but emitted nowhere.

Add a token-level validation rule (following the rule_no_empty_var_blocks
/ rule_token_no_c_style_comment pattern) that tracks CONFIGURATION and
RESOURCE nesting depth and emits P4028 when a VAR_GLOBAL keyword appears
at the top level (depth zero) with the flag off. Register it in
check_tokens.

Move allow_top_level_var_global from UNENFORCED to FLAG_FIXTURES in the
MCP feature-flag conformance suite, per the file's own instructions, with
a source snippet that is rejected off and accepted on.

Update three codegen end-to-end tests that legitimately exercise
top-level global structs (a vendor extension) to enable the flag, matching
how the existing top-level VAR_GLOBAL tests already opt in.

Fixes #1233

Claude-Session: https://claude.ai/code/session_01ReoYt12CE6RDg2AttNRYdS


compiler/parser/src/rule_no_top_level_var_global.rs
```diff
@@ -0,0 +1,209 @@
...
+/// In strict IEC 61131-3, `VAR_GLOBAL` is only permitted inside
+/// `CONFIGURATION` and `RESOURCE`. Accepting it at the top level is a vendor
+/// extension gated by `--allow-top-level-var-global`. When the flag is on the
+/// rule is a no-op.
+///
+/// "Top level" is determined by tracking block-nesting depth over the token
+/// stream: `CONFIGURATION`/`RESOURCE` open a block and
+/// `END_CONFIGURATION`/`END_RESOURCE` close it. A `VAR_GLOBAL` keyword seen at
+/// depth zero is top-level.
+pub fn apply(tokens: &[Token], options: &CompilerOptions) -> Result<(), Vec<Diagnostic>> {
```
garretfick: Why is this operating on tokens? I think this would have much less code if this checks the AST.

garretfick: Good call — switched to an AST check. The parser already emits a top-level `VAR_GLOBAL` as its own `LibraryElementKind::GlobalVarDeclarations` element (config/resource globals get folded into `ConfigurationDeclaration`), so the rule is now just "reject that element kind when the flag is off" — no CONFIGURATION/RESOURCE depth tracking. It moved from a token rule in `check_tokens` to `compiler/analyzer/src/rule_no_top_level_var_global.rs`, registered in the `semantic` stage alongside the other flag-gated rules. Diagnostic (P4028), help text, and behavior are unchanged. Done in 5051228 (rebased to 9d9e5b9).

---

PR #1256, opened 2026-07-29
fix(analyzer): diagnose constant division-by-zero and overflow

### LINKED ISSUE #1249: Constant folding: overflow/int-division-by-zero yield no (or a misleading) diagnostic, and real division by zero folds silently to inf
## Summary

Two pre-existing correctness gaps in the constant-folding arithmetic in `compiler/analyzer/src/xform_fold_constant_expressions.rs`. Both exist on `main` today; they were surfaced while reviewing #1220 (which reuses this pass's helpers for folding `VAR` initializer expressions), but neither is caused by that PR.

## 1. Real division by zero folds silently to `inf`/`NaN`

`fold_real_binary` evaluates division with a raw `left / right` and no zero check:

```rust
fn fold_real_binary(op: &Operator, left: f64, right: f64) -> f64 {
    match op {
        ...
        Operator::Div => left / right,   // 1.0 / 0.0 => inf, no diagnostic
        ...
    }
}
```

So a constant real division by zero (e.g. `x := 1.0 / 0.0;`) folds to an `inf` (or `NaN` for `0.0/0.0`) literal with **no diagnostic at all**. This is asymmetric with the integer path, which declines to fold division by zero.

## 2. Integer overflow and integer division by zero produce no diagnostic (or, in initializer position, a misleading one)

`fold_integer_binary` correctly returns `None` on `checked_*` overflow and on integer division/mod by zero:

```rust
Operator::Add => left.checked_add(right),   // None on overflow
Operator::Div => if right == 0 { None } else { left.checked_div(right) },
```

Returning `None` means "don't fold." The expression is then left as-is:

- In `xform_fold_constant_expressions` (statement bodies), the unfolded binary op simply passes through with **no diagnostic** — an overflowing or divide-by-zero constant expression is silently accepted and left for a later stage (or codegen) to deal with.
- Reused from `xform_fold_initializer_expressions` (see #1220), the unfolded node reaches `normalize()`, which reports it as **"not a constant expression" (P4037)** — but `1000000000 * 1000000000` (DINT) or `5 / 0` *is* a constant expression; it just doesn't evaluate. The diagnostic tells the user the wrong thing.

## Expected

Constant folding should treat "this is a constant expression that cannot be evaluated" as a distinct, diagnosable outcome rather than either silently producing `inf`/`NaN` (real division) or silently passing through / mislabeling as "not constant" (integer overflow & division by zero). Suggested:

- Reject constant division/modulo by zero (integer **and** real) with a dedicated diagnostic.
- Reject constant integer overflow with a dedicated "constant expression overflows its type" diagnostic.
- Keep the "genuinely not a constant expression" diagnostic (P4037) distinct from "was constant, could not be evaluated."

## Repro sketch

```iecst
PROGRAM main
VAR
    a : LREAL := 1.0 / 0.0;   (* folds to inf, no diagnostic *)
    b : DINT  := 5 / 0;       (* not folded; no diagnostic in body context *)
END_VAR
END_PROGRAM
```

## Pointers

- `compiler/analyzer/src/xform_fold_constant_expressions.rs` — `fold_real_binary` (real div) and `fold_integer_binary` (integer overflow / div / mod by zero), consumed by `try_fold_binary`.
- Related review context: #1220.

thusser:
## Summary

Fixes #1249. `fold_integer_binary`/`fold_real_binary` in `constant_folding.rs` collapsed two different situations into a single `None`:

- operands aren't both constant literals (nothing to fold, not an error)
- operands *are* both constant, but the operation has no defined result (division by zero, overflow)

That meant:
- Real division by zero (`1.0 / 0.0`) silently folded to `inf`/`NaN` with no diagnostic.
- Integer overflow / integer division-or-modulo by zero silently passed the unfolded node through with no diagnostic in statement bodies (`xform_fold_constant_expressions`).
- The same case reused from `xform_fold_initializer_expressions` (#1220) got reported as **P4038 "not a constant expression"** — misleading, since the expression *is* constant, it just can't be evaluated.

## Changes

- New `FoldError` enum (`DivisionByZero`, `Overflow`) in `constant_folding.rs`. `fold_integer_binary`/`fold_real_binary`/`try_fold_binary` now return `Result` instead of `Option`, since the caller has already established both operands are constant by the time these run.
- Integer `Pow` with a negative exponent is unchanged behavior (still silently left unfolded, not misreported as overflow) — handled explicitly by the caller before the fallible path runs.
- Two new problem codes: `P4039 ConstantExpressionDivisionByZero`, `P4040 ConstantExpressionOverflow`, each with docs.
- Both `xform_fold_constant_expressions` and `xform_fold_initializer_expressions` convert a `FoldError` into the appropriate diagnostic and propagate it through their existing `Result`-threaded `Fold` passes (no new plumbing needed for the statement-body pass; `substitute_and_fold` in the initializer pass now returns `Result<Expr, Diagnostic>`).
- New tests in both files covering integer div/mod by zero, real div by zero, integer overflow, and confirming the negative-exponent case still silently doesn't fold. Also a regression test proving the initializer-expression path reports the new codes instead of the misleading P4038.

## Test plan

- [x] `cargo test --workspace` — all passing
- [x] `cd compiler && just` (compile, coverage ≥85%, clippy, fmt, dupes) — all green

### DIFF
--- a/compiler/analyzer/src/constant_folding.rs
+++ b/compiler/analyzer/src/constant_folding.rs
@@ -11,7 +11,33 @@
 
 use ironplc_dsl::common::*;
 use ironplc_dsl::core::SourceSpan;
+use ironplc_dsl::diagnostic::{Diagnostic, Label};
 use ironplc_dsl::textual::*;
+use ironplc_problems::Problem;
+
+/// A constant expression whose operands are both known at compile time,
+/// but whose operation has no defined result (as opposed to an expression
+/// that simply isn't constant at all, which is `try_fold_binary`/
+/// `try_fold_unary` returning `Ok(None)` and is not an error here).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum FoldError {
+    DivisionByZero,
+    Overflow,
+}
+
+/// Converts a `FoldError` into a user-facing diagnostic at the given span.
+pub(crate) fn fold_error_to_diagnostic(err: FoldError, span: SourceSpan) -> Diagnostic {
+    match err {
+        FoldError::DivisionByZero => Diagnostic::problem(
+            Problem::ConstantExpressionDivisionByZero,
+            Label::span(span, "Division or modulo by zero"),
+        ),
+        FoldError::Overflow => Diagnostic::problem(
+            Problem::ConstantExpressionOverflow,
+            Label::span(span, "Arithmetic overflow"),
+        ),
+    }
+}
 
 /// Extracts the value of an integer literal as an i128.
 pub(crate) fn integer_value(lit: &IntegerLiteral) -> i128 {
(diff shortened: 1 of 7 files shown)


garretfick:
Please address the build failure.

PR #1257, opened 2026-07-29
Design: library support

garretfick:
## Summary

Design document for a real **library** mechanism in IronPLC, to replace the per-function `--allow-*` flag approach currently used to add vendor functions (e.g. `LTRUNC`/`LMOD` in #1217). Part of the #1199 lineage.

The design follows the conventional compiler approach:

- Ship IronPLC's standard library as **IEC source** under `compiler/library/`, compiled by the normal pipeline (not Rust baked into the binary).
- Make library POUs/types available to user programs by **reference**, with a **standard library implicitly in scope** (prelude-style, `--no-stdlib` to opt out) and additional libraries resolved from a **search path**.
- Verify in **CI that every bundled library file compiles**, so the library can't silently break.
- Ship the library files with every installer.
- Add the **VM primitives** (float trunc/mod under the unified `BUILTIN` opcode, per ADR-0008) that let functions like `LTRUNC`/`LMOD` be authored as library content — the compiler grows *primitives*, not *functions*.

## What's in this PR

Just the design doc: `specs/design/library-support.md`. No code.

- Annotated with `LIB`-area spec requirements partitioned by owning crate (`sources` / `parser` / `analyzer` / `codegen` / `container` / `vm`), following the `reference-to-twincat.md` hundreds-block convention.
- **Design-only / CI-neutral:** the doc is intentionally *not* wired into any `build.rs`, so the spec-conformance guard does not enforce it yet (verified: `spec_conformance_guard` passes 10/10). Each implementation slice wires the doc into its crate's `build.rs` and adds its `#[spec_test]` tests.

## Notable design points to review

- **Open decision D1 — delivery mechanism:** embed the library into the binary via `include_str!` (recommended; no installer changes, no runtime path lookup) vs. install the files alongside the binary and resolve at runtime (needs new path-resolution code + edits to all four installers). Both keep the source under `compiler/library/`.
- Reframes #1217: `LTRUNC`/`LMOD` become two VM primitives + a library alias, not a per-function flag.
- Two new problem codes proposed: `P4039` (referenced library not found), `P4040` (function requires an unreferenced library).
- A **10-PR implementation breakdown** for landing this in digestible pieces, including an up-front "enforcement bootstrap" PR (mirroring how `reference-to-twincat.md` turned on conformance).

## Test plan

- [x] `spec_conformance_guard` passes (design doc is not enforced; no unclaimed slugs).
- [ ] Design review / feedback on the branch.

---
_Generated by [Claude Code](https://claude.ai/code/session_018TmDFQKZDRLasYiYREHXRx)_

### DIFF
--- a/specs/design/library-support.md
+++ b/specs/design/library-support.md
@@ -2,103 +2,119 @@
 
 ## Overview
 
-IronPLC has no library mechanism. Its "standard library" (TON, CTU, string
-functions, …) is hand-written Rust baked into the compiler and registered
-programmatically (`analyzer/src/stages.rs`, `analyzer/src/stdlib.rs`). Every
-function IronPLC lacks — e.g. Beckhoff's `LTRUNC`/`LMOD`
-([#1217](https://github.com/ironplc/ironplc/pull/1217), part of #1199) — today
-requires new Rust and, often, a per-function `--allow-*` flag. That does not
-scale and pushes ordinary library functions into the compiler.
+IronPLC's standard library (TON, CTU, string functions, …) is hand-written Rust
+baked into the compiler (`analyzer/src/stages.rs`, `analyzer/src/stdlib.rs`).
+**That Rust-baked standard library stays.** This design adds a complementary
+mechanism: **compatibility libraries** — IronPLC-authored IEC 61131-3 **source**
+that the compiler reads and compiles through the same pipeline as user code,
+shipped and installed alongside the compiler.
 
-This design introduces a **library**: IronPLC's own IEC 61131-3 **source** that
-the compiler reads and compiles through the same pipeline as user code, shipped
-with the compiler and installed alongside it. It follows the conventional model
-— a library that ships with the toolchain and is available to every program,
-like a language's standard library.
+The purpose is to provide implementations **equivalent to the base libraries of
+other runtimes** (e.g. Beckhoff TwinCAT and CODESYS), so a program written for
+one of those runtimes compiles against IronPLC-authored source instead of
+requiring new Rust and a per-function `--allow-*` flag for every missing function
+(e.g. `LTRUNC`/`LMOD`, [#1217](https://github.com/ironplc/ironplc/pull/1217),
+part of #1199).
 
-IronPLC provides its **own** libraries. It does **not** redistribute third-party
-vendor libraries (Beckhoff `Tc2_*`, etc.) — that is a licensing non-starter.
-Instead, to be compatible with vendor and edition dialects, IronPLC provides
-**compatibility libraries** it authors itself.
+**This is deliberately not a complete, general library system, and not a
+replacement for the Rust-baked standard library.** It is a targeted way to ship
+runtime-equivalent library content as source, opted into per library. IronPLC
+provides only its **own** source; it does **not** redistribute third-party vendor
+libraries (Beckhoff `Tc2_*`, etc.) — a licensing non-starter.
 
 ### Scope
 
-**In scope:** shipping a single IronPLC-authored library as `.st` source,
-selected by the active dialect/edition, loaded from disk, compiled by the normal
-pipeline, verified by CI, and installed with the compiler.
+**In scope:** shipping IronPLC-authored `.st` libraries, each selected
+**explicitly** (by a compiler flag and/or a library reference in project
+configuration), loaded from disk, compiled by the normal pipeline, verified by
+CI, and installed with the compiler. The first library defines the constant `PI`.
 
 **Out of scope (future designs):**
 
-- *How to offer multiple or vendor-specific compatibility libraries* and
-  reference them individually. For now IronPLC offers a single dialect-matched
-  library.
+- A complete/general library system (versioning, dependency resolution, a full
+  reference model). Libraries here are a flat set of named files, opted into
+  individually.
 - Functions that require an operation the language cannot express (a float→float
-  truncation, etc.) and the compiler/VM support they need. `LTRUNC`/`LMOD` fall
-  here and are deferred to a follow-on design; this design builds the library
-  foundation they will use.
+  truncation, etc.) and the compiler/VM primitives they need — `LTRUNC`/`LMOD`
+  are deferred to a follow-on design that builds on this foundation.
 
 ### Goals
 
-- Ship IronPLC's library as IEC 61131-3 `.st` source the compiler reads, not
-  Rust baked into the binary.
-- Make the dialect-appropriate library available to every compilation.
+- Ship IronPLC-authored compatibility libraries as `.st` source the compiler
+  reads.
+- Let a user **opt into** a library by name (compiler flag and/or project-config
+  reference); dialects do not select libraries.
 - Install the library files alongside the compiler so users can see and read
   them.
 - Verify in CI that every shipped library file compiles.
+- Ship a first library that defines `PI` (the right resolution for #1219).
 
-### Approach: what other compilers do
+### Approach
 
-- A library ships with the toolchain and is available by default (Rust's
-  `std`/prelude, Go's `std`, the C runtime).
-- Library code is **ordinary source** compiled by the same front end. The
-  compiler already supports this: `analyze()` merges all units into one library
-  and `xform_toposort_declarations` prunes everything unreachable from a PROGRAM
-  root, so the library can be supplied and only used POUs are emitted.
+- Libraries ship with the toolchain but are **opted into explicitly**, the way a
+  TwinCAT/CODESYS project adds a library reference or a C build links a library —
+  not implicitly global. Those runtimes offer many libraries and you opt into the
+  specific ones you use.
+- Library code is **ordinary source** compiled by the same front end. `analyze()`
+  merges selected libraries with user code into one unit, and the reachability
+  pass prunes everything unused — **including global constants** — so a library
+  can define `PI` and it is emitted only when a program uses it.
 
 ## Architecture
 
 ### Library source
 
 **REQ-LIB-sources-100** IronPLC ships library source under `compiler/library/`.
-Each immediate subdirectory of `compiler/library/` is one library, named by the
-directory.
-
-**REQ-LIB-sources-101** A library is named to match a dialect/edition (the same
-names the compiler uses for its dialects), so the standard library can differ by
-IEC 61131-3 edition. The library matching the active dialect is available to
-every compilation by default. IronPLC ships a single such library initially.
+Each immediate subdirectory of `compiler/library/` is one library, identified by
+its directory name.
 
 **REQ-LIB-sources-102** Library source files are `.st` files, parsed by the same
-parser as user source with no library-only grammar.
+parser as user source, with no library-only grammar.
+
+**REQ-LIB-sources-180** The first shipped library defines the constant `PI` (an
+`LREAL` global constant), providing the resolution for #1219 in library source
+rather than a compiler flag.
 
 Keeping the files as real `.st` source (rather than Rust string literals) makes
 them reviewable, diffable, editable, **visible to users after install**, and —
 critically — compilable by the same CI that builds user programs (see
 [Build-time compile check](#build-time-compile-check)).
 
-### Delivery: installed alongside the binary
+### Selecting a library
 
-Library files are shipped as files and installed next to the compiler, then
-loaded from disk at runtime. This keeps the library **visible and readable to
-users** and makes the set **extensible** — adding or adjusting a library is a
-file change, not a compiler rebuild.
+**REQ-LIB-sources-130** A library is included in a compilation only when
+explicitly selected — by a compiler flag naming the library, and/or by a library
+reference embedded in project configuration (e.g. a TwinCAT `.plcproj`
+`<LibraryReference>`). Dialects do not select libraries, and no library is
+included by default.
 
-**REQ-LIB-sources-110** The compiler locates its library directory relative to
-the installed executable, independent of the current working directory and of
-the user's project directory.
+Selection is resolved in the CLI/`sources` layer. The **parser crate does not
+learn about libraries**; if library selection ever needs option plumbing,
+`CompilerOptions` should be restructured so the parser holds only its
+syntax-affecting subset and the CLI owns the rest.
 
-**REQ-LIB-sources-140** If the library for the active dialect cannot be located
-or read, the compiler emits a clear diagnostic (proposed `P4039`) identifying
-the missing library, rather than silently omitting standard functions.
+### Compiling a library — per-file compiler flags via pragma
 
-Because the library is read from disk as ordinary files, its elements carry
-normal file-based `FileId`s — no special builtin tagging is needed, and
-diagnostics and debug info point at the real library file on disk.
+A library may need to be compiled with compiler flags that differ from the
+user's. The immediate case: defining `PI` needs a **global constant declaration
+block** (a top-level `VAR_GLOBAL CONSTANT`, gated by `--allow-top-level-var-global`
+today). Rather than compile all libraries under a fixed flag set, a library
+source file declares the flags it needs via a **pragma**.
+
+**REQ-LIB-parser-200** A source file may enable specific compiler flags for its
+own compilation via a pragma. Initially this covers allowing a global constant
+declaration block (needed to define `PI`). The mechanism is a general,
+file-scoped flag directive — it is not library-specific.
+
+This separates two kinds of rule, and the pragma feeds both: **parse-time rules**
+a pragma switches on while parsing, and **analysis rules** that operate on what
+was written. Start with only the flag `PI` needs; grow the recognized set as
+further libraries require it, toward a general mechanism.
 
 ### Semantic integration
 
-**REQ-LIB-analyzer-300** POUs, types, and global variables exported by the
-active library resolve in user code as if declared in the compilation, using the
+**REQ-LIB-analyzer-300** POUs, types, and global variables exported by a selected
+library resolve in user code as if declared in the compilation, using the
 existing merge performed by `analyze()`.
 
 **REQ-LIB-analyzer-330** A user declaration whose name collides with a library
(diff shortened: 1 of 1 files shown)


specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
+# Design: Library Support
+
+## Status
+
+Design document for the library-support subsystem. This PR contains the design
```
garretfick: Omit this "Status" section. The design doesn't reference state.

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+### Goals
+
+- Ship IronPLC's own standard library as IEC 61131-3 **source** the compiler
+  reads, not Rust baked into the binary.
+- Let user programs use library POUs/types by **referencing** a library, with a
+  standard library available by default.
+- Verify in **CI that every bundled library file compiles**, so the library
+  cannot silently break.
+- Ship the library files with every installer.
+- Provide the **primitive opcodes** that let library functions like
```
garretfick: Let's not say 'primitive opcodes'. Let's just say 'primitives'

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+  `LTRUNC`/`LMOD` be authored outside the compiler.
+
+### Non-goals
+
+- **Redistributing third-party vendor libraries** (Beckhoff `Tc2_*`, etc.).
```
garretfick: We don't ship third-party vendor libraries for licensing reasons. We want compatibility and IronPLC will provide compatibility libraries.

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+  IronPLC ships only its own library. Consuming a user's *installed* vendor
+  libraries is a downstream extension that reuses this design's reference and
+  resolution mechanism; it is out of scope here.
+- A full namespace / version / transitive-dependency resolver. First cut is
+  name-based resolution of a flat set of bundles.
+
```
garretfick, line 51: Out of scope is specifically how to offer vendor specific compatibility libraries. For now, we will just offer one.

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+files under `compiler/library/`. Each immediate subdirectory of
+`compiler/library/` is one named library bundle (the directory name is the
+bundle name).
+
+**REQ-LIB-sources-101** The bundle named `standard` is available to every
+compilation by default, without an explicit reference, unless `--no-stdlib` is
+set.
+
+**REQ-LIB-sources-102** Library source files use the same Structured Text / IEC
+syntax and file extensions (`.st`, `.iec`) as user source and are parsed by the
```
garretfick: Let's make it only `.st`

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+same parser with no library-only grammar.
+
+Keeping the files as real source under `compiler/library/` (rather than Rust
+string literals) makes them reviewable, diffable, editable by contributors, and
+— critically — **compilable by the same CI that builds user programs** (see
+[Build-time compile check](#build-time-compile-check)).
+
+### Delivery to the compiler — Open decision D1
+
+The library files must reach the *installed* compiler. Two mechanisms, both
```
garretfick: I prefer Option B long term. I want this to be extensible and Option B makes this extensible. I want users to be able to see the file content

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+
+**REQ-LIB-sources-110** Library bundle sources are available to the compiler
+independent of the current working directory and of the user's project
+directory.
+
+**REQ-LIB-sources-111** *(Option A)* Library sources are embedded into the
+`ironplcc` binary at build time; the installed compiler loads them with no
+filesystem access.
+
+**REQ-LIB-sources-120** Elements originating from a library bundle are tagged
```
garretfick: Since we are loading from disk, I think we don't need anything special here

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+
+**REQ-LIB-sources-131** When a project is discovered from a `.plcproj`, each
+`<LibraryReference>` / placeholder entry in the project file is extracted as a
+library reference (discovery currently reads only `<Compile Include>`).
+
+**REQ-LIB-sources-140** Referencing a bundle that cannot be resolved on the
+search path produces a diagnostic (proposed `P4039`, "referenced library not
+found") naming the missing bundle, rather than a generic error.
+
+**REQ-LIB-parser-200** The `--no-stdlib` compiler option disables the implicit
```
garretfick: We need to allowdifferent versions of standard library depending on the version of 61131-3. The name must be more specific than 'standard'. We can use the same name as the dialect?

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+availability of the `standard` bundle.
+
+**REQ-LIB-parser-210** The `--library <name>` compiler option references an
+additional bundle by name; it is repeatable.
+
+**REQ-LIB-parser-220** The `--library-path <dir>` compiler option adds a
```
garretfick: I don't think we need this to start

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+directory to the library search path; it is repeatable.
+
+These three become fields on `CompilerOptions` (owned by the `parser` crate),
```
garretfick: I think CompilerOptions needs to move out if parser crate into its own crate or the parser just has the minimum of what we need and the cli defines the total set with a mapping to the parser. I don't want the parser to know about libraries

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+**REQ-LIB-analyzer-300** POUs, types, and global variables exported by an
+in-scope library bundle resolve in user code as if declared in the compilation,
+using the existing merge performed by `analyze()`.
+
+**REQ-LIB-analyzer-310** A bundle's exports are in scope only when the bundle is
+referenced. The `standard` bundle is implicitly referenced unless `--no-stdlib`;
+all other bundles are in scope only when referenced via `--library`,
+`--library-path`, or a project `<LibraryReference>`.
+
+**REQ-LIB-analyzer-320** A call to a function that exists only in a bundle that
```
garretfick: This isn't right. We don't know if a missing function should be in a library

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+is **not** referenced produces a diagnostic (proposed `P4040`, "function
+requires an unreferenced library") that names the bundle providing it, distinct
+from the generic undeclared-function diagnostic (`P4017`).
+
+**REQ-LIB-analyzer-330** A user declaration whose name matches a library export
+shadows the library export within the user compilation (local-over-library
```
garretfick: I think this should be an error. Just use the standard error we already have for duplicate function defs

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+precedence); it is not a redefinition error.
+
+**REQ-LIB-analyzer-340** A function declared with an external-binding attribute
```
garretfick: I'd actually like this to be out of scope for this design. We can add this if needed in a future design just about this topic

specs/design/library-support.md
```diff
@@ -0,0 +1,353 @@
...
+compiled as a `CALL` to the compiled library POU, using the same path as a call
+to a user-defined function.
+
+**REQ-LIB-codegen-420** A call to an external-binding library function emits the
+bound `BUILTIN` opcode inline at the call site (no `CALL` frame), the way
+`compile_trunc` already inlines its conversion.
+
+### Primitive-backed library functions
+
+Most of a library is ordinary ST. A few functions need an operation the language
```
garretfick: I want to omit this from the current design

garretfick:
Updated in aee0d78. How each comment was addressed:

**Scope narrowed to a single dialect-matched library**
- Compatibility framing: IronPLC authors its own **compatibility libraries**; no third-party redistribution. Reworded the intro + non-goals.
- Out-of-scope is now specifically *how to offer multiple / vendor-specific* compatibility libraries — for now, one.
- Deferred to a follow-on design (and removed from here): the primitive-backed functions section, the external-binding attribute, and the `LTRUNC`/`LMOD` mechanics. This design is now just the library foundation they'll build on.

**Delivery**
- Went with **Option B** (install alongside, loaded from disk) — files stay user-visible and the set is extensible. Removed the embed option and its recommendation.
- Since it's loaded from disk, dropped the special builtin `FileId` — library elements get normal file-based IDs.

**Library identity**
- Named to match the **dialect/edition** rather than `standard`, so the standard library can vary per IEC 61131-3 edition (your suggestion — flagging for confirmation).
- `.st` only.

**Options / parser boundary**
- Parser no longer knows about libraries — dropped `--library` / `--library-path`; selection follows `--dialect` in the CLI/`sources` layer. Added a note that `CompilerOptions` should be restructured (own crate, or CLI-owns-total-with-parser-subset) if library options are ever added.

**Semantics**
- Name collision with a library export is now the **existing duplicate-definition error**, not shadowing.
- Removed the "used-but-not-referenced" diagnostic (`P4040`) — agreed we can't know a missing function belongs to a library. `P4039` now covers only "library for the active dialect couldn't be loaded".

**Cleanup**
- Removed the Status section; "primitive opcodes" → "primitives" (the remaining primitive discussion moved to the future design).

Two things I'd like you to confirm: (1) naming libraries after the dialect, and (2) that deferring the primitives/`LTRUNC`-`LMOD` mechanics to a separate follow-on design is what you intended by the out-of-scope comments.

---

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+like a language's standard library.
+
+IronPLC provides its **own** libraries. It does **not** redistribute third-party
+vendor libraries (Beckhoff `Tc2_*`, etc.) — that is a licensing non-starter.
+Instead, to be compatible with vendor and edition dialects, IronPLC provides
+**compatibility libraries** it authors itself.
+
+### Scope
+
+**In scope:** shipping a single IronPLC-authored library as `.st` source,
```
garretfick: I want to say "selected by a compiler flag and/or selected by library references embedded in project files (e.g. TwinCAT)"

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+selected by the active dialect/edition, loaded from disk, compiled by the normal
+pipeline, verified by CI, and installed with the compiler.
+
+**Out of scope (future designs):**
+
+- *How to offer multiple or vendor-specific compatibility libraries* and
```
garretfick: We will offer different files by name and use that to select them. The purpose here is to provide implementations equivalent to base libraries in various runtimes (e.g. TwinCAT and Codesys). This is not a replacement for a complete library system. We should call that out.

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+- Functions that require an operation the language cannot express (a float→float
+  truncation, etc.) and the compiler/VM support they need. `LTRUNC`/`LMOD` fall
+  here and are deferred to a follow-on design; this design builds the library
+  foundation they will use.
+
+### Goals
+
+- Ship IronPLC's library as IEC 61131-3 `.st` source the compiler reads, not
+  Rust baked into the binary.
+- Make the dialect-appropriate library available to every compilation.
```
garretfick: This isn't quite right. Dialects should not directly select libraries. twinCAT has many libraries and you have to opt in to specific libraries

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+- Verify in CI that every shipped library file compiles.
+
+### Approach: what other compilers do
+
+- A library ships with the toolchain and is available by default (Rust's
+  `std`/prelude, Go's `std`, the C runtime).
+- Library code is **ordinary source** compiled by the same front end. The
+  compiler already supports this: `analyze()` merges all units into one library
+  and `xform_toposort_declarations` prunes everything unreachable from a PROGRAM
+  root, so the library can be supplied and only used POUs are emitted.
```
garretfick: We need to prune global constants (included when we say everything).

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Architecture
+
+### Library source
```
garretfick, line 65: We'll also need to allow libraries to run with different compiler flags ( at least some). The key one here is we'll need to support constant declaration blocks for this library (to define PI). I think the right way to do this is via a pragma. I want to start simple but then make this a general mechanism.

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+**REQ-LIB-sources-100** IronPLC ships library source under `compiler/library/`.
+Each immediate subdirectory of `compiler/library/` is one library, named by the
+directory.
+
+**REQ-LIB-sources-101** A library is named to match a dialect/edition (the same
+names the compiler uses for its dialects), so the standard library can differ by
+IEC 61131-3 edition. The library matching the active dialect is available to
+every compilation by default. IronPLC ships a single such library initially.
+
+**REQ-LIB-sources-102** Library source files are `.st` files, parsed by the same
```
garretfick, line 71: This implies separation of rules - we have parsing rules affected by pragmas and then analysis rules that are on what is written

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+Implemented as an integration test that runs the real `check()` + `compile()`
+entry points over the `compiler/library/` tree (directory input is already
+supported; see the existing `check(&[resource_path("set")], …)` test in
+`ironplc-cli/src/cli.rs`) and asserts success. It runs as part of `just`
+(`compile`/`coverage`) so a change that breaks a library file fails CI the same
+as any other test. It is the guardrail the current Rust-baked stdlib lacks.
+
+## Compiler-options boundary
+
+Selecting the library follows the active dialect/edition, which the compiler
```
garretfick: This isn't right. Either the user selects via a command line argument or we read from project configuration

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+- Runtime path-resolution locates the library directory relative to the
+  executable (REQ-LIB-sources-110).
+
+The existing release workflows (`partial_compiler.yaml`,
+`partial_upload_release_artifacts.yaml`) manipulate whole artifacts and need no
+change.
+
+## Problem codes
+
+- `P4039` — the library for the active dialect could not be located or read
```
garretfick: I think this isn't needed for this MLP

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+enforcement is turned on once, up front.
+
+- **PR 1 — this document.** Design only; not wired into any `build.rs`.
+- **PR 2 — enforcement bootstrap.** Wire `library-support.md` into the `build.rs`
+  of every owning crate (`sources`, `analyzer`, `codegen`) and add `#[ignore]`d
+  `#[spec_test]` stubs for every requirement, so the requirements become
+  enforced-but-pending (mirrors how `reference-to-twincat.md` established
+  enforcement). Each later PR removes `#[ignore]` as it implements.
+- **PR 3 — library skeleton + compile check** (`sources`): create
+  `compiler/library/<dialect>/`, add one or two trivially-correct `.st` POUs, and
```
garretfick: I want to add one first library that defines PI and provides the "right way" to resolve https://github.com/ironplc/ironplc/pull/1219

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+  the REQ-LIB-sources-190 compile-check test. Small; locks in the guardrail first
+  and unblocks authoring.
+- **PR 4 — delivery + packaging** (`sources`): install the library alongside the
+  binary and resolve it at runtime relative to the executable; update all
+  installers; REQ-LIB-sources-110/140.
+- **PR 5 — pipeline integration** (`sources` + `analyzer` + `codegen`): load the
+  dialect's library into every compilation; exports resolve; duplicate-definition
+  on collision; reachability-pruned linking; REQ-LIB-analyzer-300/330,
+  REQ-LIB-codegen-400/410.
+- **PR 6+ — library content**: port stdlib functions from Rust to library source
```
garretfick: We are not porting anything today

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+  enforcement). Each later PR removes `#[ignore]` as it implements.
+- **PR 3 — library skeleton + compile check** (`sources`): create
+  `compiler/library/<dialect>/`, add one or two trivially-correct `.st` POUs, and
+  the REQ-LIB-sources-190 compile-check test. Small; locks in the guardrail first
+  and unblocks authoring.
+- **PR 4 — delivery + packaging** (`sources`): install the library alongside the
+  binary and resolve it at runtime relative to the executable; update all
+  installers; REQ-LIB-sources-110/140.
+- **PR 5 — pipeline integration** (`sources` + `analyzer` + `codegen`): load the
+  dialect's library into every compilation; exports resolve; duplicate-definition
```
garretfick: Not right. It is based on flags

specs/design/library-support.md
```diff
@@ -0,0 +1,233 @@
...
+- Complements **[ADR-0012](../adrs/0012-accept-vendor-dialect-files-as-is.md)**
+  (accept vendor dialect files as-is) — the library provides the compatibility
+  counterpart in IronPLC-authored source.
+- Conformance mechanics:
+  **[cross-crate-spec-conformance.md](cross-crate-spec-conformance.md)** and
+  **[ADR-0037](../adrs/0037-mandatory-crate-slug-in-requirement-ids.md)**.
+
+### Decisions to ratify as ADRs
+
+1. IronPLC ships a **source** library compiled by the normal pipeline, replacing
```
garretfick: We will still have the rust baked for the time being

garretfick:
Updated in cc0ff1e. This round was a bigger reframe:

**Purpose / positioning**
- Reframed as **compatibility libraries**: IronPLC-authored implementations equivalent to the base libraries of other runtimes (TwinCAT, CODESYS). Explicitly called out that this is **not a complete/general library system**.
- Made explicit that it's **additive** — the Rust-baked standard library **stays**; this doesn't replace it. Removed the "replacing the Rust-baked stdlib" ADR line and the "port stdlib from Rust" work.

**Selection**
- Dialects no longer select libraries. Selection is **explicit opt-in**: a compiler flag naming the library and/or a library reference in project configuration (e.g. TwinCAT `<LibraryReference>`). Nothing is included by default. Updated scope, goals, the selection requirement, and the pipeline PR accordingly.

**First library = PI**
- The first shipped library defines `PI`, resolving #1219 in source rather than a flag (REQ-LIB-sources-180).

**Per-file compiler flags via pragma**
- Added the mechanism: a library file declares the flags it needs via a **pragma** (initially: allow a top-level global constant block, for `PI`). Noted the **parse-time-rules vs analysis-rules** separation, and "start simple, then generalize." New REQ-LIB-parser-200; it's a general file-scoped directive, so the parser still doesn't know about *libraries*.

**Pruning**
- Extended the reachability requirement to prune **global constants** too (REQ-LIB-codegen-400), so an unused library `PI` isn't emitted.

**MLP trims**
- Dropped the "library not found" diagnostic and the whole Problem codes section (not needed for the MLP).

**PR breakdown** now: bootstrap → pragma → PI library + compile check → delivery/packaging → selection + pipeline → more libraries. The pragma lands before the PI library since PI depends on it.

One thing worth confirming: what the **compiler flag that names a library** looks like (e.g. `--library <name>`) vs relying solely on project-config references for now — I left it described but not pinned to an exact flag spelling.

---

PR #1258, opened 2026-07-29
FMI co-simulation support design doc

garretfick:
## Summary

Adds `specs/design/fmi-co-simulation-support.md` — an architecture and feasibility design for exporting an IronPLC program as an **FMU** implementing the **Functional Mock-up Interface (FMI)** for **co-simulation** (for the Open Industry Project and other FMI masters).

This is a **design/feasibility document for review**, not an implementation. It ends with numbered open questions to be answered as PR feedback; the design will be finalized from those answers before any plan lands in `specs/plans/`.

## Key findings

- **Feasible, and the fit is unusually strong.** The two things that normally make FMI export hard for a PLC runtime are already externalized in IronPLC's VM:
  - **The master owns the clock** — `VmRunning::run_round(current_time_us)` takes injected time; the VM never sleeps (the sleep loop is in the CLI). The runtime model already specifies a `simulated` clock for deterministic replay. This is `fmi*DoStep`'s contract directly.
  - **The I/O boundary is pluggable** — the runtime model deliberately leaves the "I/O driver" out of scope, touching I/O only at `INPUT_FREEZE`/`OUTPUT_FLUSH`. **An FMU is exactly that driver.**
- **I/O is relevant but not a blocker.** Located `%I`/`%Q` variables currently compile to ordinary variable-table slots; the process image is specified but unimplemented (the two `run_round` stubs, unused `input_image_offset`/`output_image_offset`, no `LOAD_INPUT`/`STORE_OUTPUT` opcodes). A **minimal path** maps FMI variables straight to `VarIndex` and needs no process-image work; the full process image is a later, independent upgrade with an identical external FMU contract.
- **The one real friction:** the FMI C ABI is inherently `unsafe`, which collides with the workspace `unsafe_code = "deny"` lint. The design quarantines all `unsafe` in a separate `ironplc-fmi` cdylib shim that wraps the safe VM, leaving `ironplc-vm` unchanged.

## What the doc covers

- Why the architecture fits (with `file:line` references into the VM, scheduler, container, and analyzer).
- Current state of I/O and the two implementation paths (minimal vs process image).
- Four new components: value-reference mapping, `modelDescription.xml` generation, the unsafe FFI shim crate, and the `DoStep`↔scan time mapping.
- FMI 2.0-vs-3.0 recommendation (3.0 for faithful 64-bit type mapping).
- A phased implementation plan and out-of-scope list.
- **10 open questions** for reviewer feedback (unsafe-boundary ADR, interface-metadata source, causality of globals, step/scan semantics, reset, I/O ordering, FMI version, multi-task, packaging entry point, target platforms).

## Notes

- Docs-only change (single new markdown file under `specs/design/`); no code, tests, or build inputs are touched.
- Related project issue https://github.com/Open-Industry-Project/Open-Industry-Project/issues/273

---
_Generated by [Claude Code](https://claude.ai/code/session_01M1avGdVvaqQ3JkjLuoHJZp)_


specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+  **[Debugger Support](debugger-support.md)** — the variable-name/scope/type
+  metadata FMI needs to generate `modelDescription.xml`, and the existing
+  external variable read/write path.
+- **[No-std VM](no-std-vm.md)** / **ADR-0010** — the commitment that the VM core
+  stays `unsafe`-free and embeddable. This constrains where the FMI C ABI lives
+  (see [The FMI ABI shim](#component-3-the-fmi-abi-shim-crate)).
+
+## Design Goals
+
+1. **A standards-conformant FMU** — a compiled IronPLC program packages into a
```
garretfick: What exactly is a `.fmu`? Is that an executable program?

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+   initialize, step, and read, with no IronPLC-specific tooling on the master
+   side.
+2. **The VM core stays `unsafe`-free** — the C ABI (`extern "C"`, raw pointers)
+   is inherently `unsafe`. It must be quarantined in a thin, auditable boundary
+   crate that lives outside the workspace `unsafe_code = "deny"` lint. The
+   `ironplc-vm` crate is not modified to add `unsafe`.
+3. **Deterministic co-simulation** — given the same input trajectory and the
+   same sequence of communication points, the FMU produces identical outputs.
+   This follows directly from the VM's `simulated` clock model.
+4. **Reuse, don't fork, the execution core** — FMI is a new *driver* around the
```
garretfick: Open question here is whether FMU needs the debugger-style capabilities.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+- `resources/` — arbitrary files the binary may load at instantiation.
+
+The co-simulation C API is a small lifecycle (names shown for FMI 2.0; FMI 3.0
+is analogous):
+
+| FMI call | Meaning |
+|----------|---------|
+| `fmi2Instantiate` / `fmi2FreeInstance` | Create / destroy an instance |
+| `fmi2SetupExperiment`, `fmi2EnterInitializationMode`, `fmi2ExitInitializationMode` | Configure start/stop time and run initialization |
+| `fmi2SetReal/Integer/Boolean/String` | Write inputs (and tunable parameters) by value reference |
```
garretfick: PLC has more types than just Integer. We need to handle this.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+`VmRunning::run_round(current_time_us)` (`compiler/vm/src/vm.rs:339`) takes time
+as an **argument**. The VM never sleeps — the sleep loop lives in the CLI
+(`compiler/vm-cli/src/cli.rs:86-92`), not the VM. The scheduler is explicitly
+documented as "time-agnostic: callers pass the current time as a `u64`
+microsecond value … fully testable without mocking clocks"
+(`compiler/vm/src/scheduler.rs:30-33`). The runtime model already specifies a
+`simulated` clock source for "deterministic replay"
+(`runtime-execution-model.md`, VM Configuration Parameters / Runtime Clock).
+
+This is precisely `fmi2DoStep`'s contract. No scan thread has to be suppressed;
```
garretfick: What do we do if the simulation skips a cycle? We need some way for the PLC to inform the minimum step size.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+the time boundary is already a function parameter.
+
+### 2. Set / get variables by index — mostly present
+
+`VmRunning` exposes `read_variable`, `read_variable_raw`, `read_variable_i64`,
```
garretfick: True, but we don't really want to expose these values. We should be mapping IO values (I think). Showing internal state is good, but that's not the right interaction pattern.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+external contract is independent of debug info.
+
+### Component 2: `modelDescription.xml` generation
+
+A build-time generator reads the container's interface metadata and emits
+`modelDescription.xml`: one scalar variable per exported `VarIndex`, with
+causality, variability, declared type, and start value. Type mapping from
+`iec_type_tag`:
+
+| IEC type | FMI 2.0 | FMI 3.0 |
```
garretfick: Let's not make FMI 2.0 an option. We only support FMI 3.0

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+| `GetReal/…` (outputs) | `read_variable_raw` / `read_variable_i64` + reinterpret by type |
+| `DoStep` | Component 4 |
+| trap during a step | `VmRunning::fault(ctx)` → `VmFaulted`; return `fmi*Error`/`Fatal` |
+| `Terminate` | `stop()` → `VmStopped` |
+| `Reset` | re-initialize to `VmReady` and re-run init — needs a reset path exposed to the shim ([Q6](#open-questions)) |
+
+### Component 6: FMU packaging
+
+A build step (proposed: an `ironplcc` subcommand, e.g. `ironplcc fmu <project>`,
+or a `just` recipe) that:
```
garretfick: not a just recipe. I think this should be `ironplcc compile` with an output format flag (I think we have `compile` already).

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+
+1. Compiles the project to a `.iplc` container.
+2. Generates `modelDescription.xml` (Component 2).
+3. Builds `ironplc-fmi` as a `cdylib` for the target platform(s).
```
garretfick: This would not build `ironplc-fmi`. This would already exist and so the build would create the ZIP including the binary.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+- **Path A — minimal (recommended first).** Map each FMI variable directly to
+  its `VarIndex` in the existing flat variable table. Inputs are written with
+  `write_variable_*` before the step; outputs are read after. **No VM execution
+  changes, no new opcodes, no process image.** Sufficient for real
+  co-simulation. Limitation: no hardware-accurate "inputs frozen for the whole
+  scan / outputs staged atomically" guarantee — but for a program modeled as one
+  cyclic task stepped once per communication point, the shim's write-before /
+  read-after ordering already provides that at the step boundary.
+
+- **Path B — process image (later, independent).** Implement the designed `%I` /
```
garretfick: I want to Path B. This will make this real.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+   addition to the safe VM.
+2. **Interface metadata + `modelDescription.xml` generator.** Decide the
+   metadata source ([Q2](#open-questions)); emit the XML from a container.
+3. **`ironplc-fmi` shim (Path A) + FMU packaging.** The `cdylib`, lifecycle,
+   time accumulator, and `.fmu` builder. First end-to-end FMU.
+4. **Conformance + a golden vertical slice.** A trivial `%I`-in / `%Q`-out
+   program exported as an FMU, stepped by a reference master (e.g. the FMI
+   cross-check / OMSimulator), with outputs asserted against a known trajectory.
+5. **(Optional, independent) Path B process image.** Implement `%I`/`%Q`/`%M`,
+   the input/output opcodes, and wire the `INPUT_FREEZE`/`OUTPUT_FLUSH` stubs.
```
garretfick: We need a #6 which is integration test

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+- The full process-image implementation (Path B) — scoped here only as it
+  relates to the FMI boundary; its own design belongs with the runtime model.
+- Real-time / hardware-in-the-loop execution. Co-simulation here is
+  logical-time, master-driven.
+
+## Open Questions
+
+Please answer these inline as PR feedback; they drive the plan.
+
+- **Q1 — Unsafe-boundary crate.** Confirm the approach of quarantining all FMI
```
garretfick: Yes. Quarantine into the FMI crate. Is there an existing FMI crate we can use or is this something we would build? I wonder if there is something reusable here.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+  `unsafe` FFI in a separate `ironplc-fmi` crate outside the
+  `unsafe_code = "deny"` lint, leaving `ironplc-vm` unchanged and safe. Should
+  this be recorded as an ADR? Should the crate be a workspace member with a
+  local `[lints]` override, or live outside the compiler workspace entirely?
+
+- **Q2 — Interface metadata source.** For the exported-variable contract, do we
```
garretfick: I think we should require located variable information (we need to build something there). We should be able to generate the modelDescription without the debug section. I think we need to create to output formats for debug sectoin. One is this mode description.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+  (a) require the (optional, strippable) debug section to be present for FMU
+  export, or (b) add a dedicated non-strippable "interface" section to the
+  container that pins `{value reference, name, causality, type, start}`
+  independently of debug info? (b) is more robust but adds container surface.
+
+- **Q3 — Causality of globals.** Should non-located `VAR_GLOBAL` map to FMI
```
garretfick: I'm not clear why you are asking about this.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+  `parameter` (tunable at init), to `local` (observable only), or be
+  configurable per variable (e.g. via a pragma/attribute)? Is there appetite for
+  a source-level annotation to mark FMI causality explicitly, rather than
+  inferring it from section/address?
+
+- **Q4 — TIME / date-time types.** How should `TIME`/`LTIME`/date-time IEC types
```
garretfick: These should be integers.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+  be exposed? As FMI `Real` seconds, as scaled integers, or excluded from the
+  interface (kept `local`)?
+
+- **Q5 — Step ↔ scan semantics.** Is "communication step size = task interval,
```
garretfick: Is there some way to influence `h`? To say the largest `h` or communicate the required `h` size?

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+  one `DoStep` = one scan" the right default? How should the FMU behave when the
+  master's `h` is not a multiple of the task interval (sub-step and run partial,
+  reject, or snap)? Should the task interval be fixed by the program or
+  overridable by the master at instantiation?
+
+- **Q6 — Reset semantics.** `fmi*Reset` returns to the instantiated state. Is
+  re-running the init functions (as on a FAULTED restart) the intended behavior,
```
garretfick: I think the shim should rebuild. That would ensure state is cleanly wiped.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+  and should the VM expose an explicit `VmRunning`/`VmFaulted → VmReady` reset,
+  or should the shim rebuild the instance from the container each time?
+
+- **Q7 — I/O path ordering.** Agree to ship Path A (direct variable mapping,
```
garretfick: I think Path B is the right way

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+  no process image) first, with Path B (process image) as a later, independent
+  improvement? Or is hardware-accurate freeze/flush a hard requirement for the
+  first FMU?
+
+- **Q8 — FMI version.** Target **FMI 3.0 CS** (faithful 64-bit type mapping),
```
garretfick: 3.0

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,431 @@
...
+  master(s) must the first release interoperate with (this pins the required
+  version and the conformance target in Phase 4)?
+
+- **Q9 — Multi-task programs.** For a program with multiple tasks at different
+  intervals, is the FMU a single unit stepping all tasks per its own internal
+  schedule within a `DoStep`, or is multi-task FMI export out of scope for v1
+  (single-task/single-program only, matching the runtime model's current
+  single-instance focus)?
+
+- **Q10 — Packaging entry point.** Preferred build surface for producing a
```
garretfick: We would require running the simulation on the same platform you compile. If that's not OK, we need a more complicated installer approach.

garretfick:
Revised in `80f5a31`. Consolidated map of feedback → change (rather than replying on each of the 19 threads):

**Settled and applied**
- **FMI 2.0 vs 3.0** → FMI **3.0 only**. Removed the 2.0/3.0 tradeoff section; type table and all API names are 3.0.
- **"Map I/O values, not internal state"** → The interface is now the program's **located `%I`/`%Q`** variables (`%I`→`input`, `%Q`→`output`). `%M`/internal state is no longer part of the default interface; read-only observability is called out as an optional add-on (Q1).
- **"PLC has more types than Integer"** → Full FMI 3.0 type table: `SINT/INT/DINT/LINT`→`Int8/16/32/64`, unsigned→`UInt*`, `REAL/LREAL`→`Float32/64`, etc.
- **TIME/date types** → mapped to **integers** (`Int64` counts with unit/epoch recorded in the model description).
- **Path B (process image)** → adopted as the substrate. New "Process Image (Path B)" section: regions, located-access opcodes, and wiring the `INPUT_FREEZE`/`OUTPUT_FLUSH` stubs. The minimal flat-slot shortcut is dropped.
- **modelDescription without the debug section** → new "Interface Metadata in the Container" section: located-variable info becomes a **non-strippable interface artifact**, and the debug info gains a **second emit format whose product is the model description**.
- **Packaging** → `ironplcc compile --format fmu` (not a `just` recipe, not a new subcommand). The compile step **does not build `ironplc-fmi`**; that prebuilt library is zipped in.
- **`fmi3Reset`** → the shim **rebuilds** the instance from the container.
- **Phase #6 integration test** → added: compile→FMU→load with an FMI 3.0 importer→multi-step scenario→assert, in CI.
- **Platform (Q10)** → v1 requires running on the **same host** it was compiled for; cross-platform noted as a deferred installer question.

**Researched and answered in the doc**
- **"Some way to influence `h` / skipped cycle"** → new "Time and Stepping" section: the FMU advertises `fixedInternalStepSize` = scan period; `fmi3DoStep` reports `lastSuccessfulTime`/early return for misaligned steps. That's how the PLC informs the master of its required step size.
- **"Existing FMI crate to reuse?"** → Yes, a strong candidate: rust-fmi's **`fmi-export`** (derive/`export_fmu!` macros) + **`cargo-fmi`**, which generate the FMI 3.0 C API, type-safe variable handling, and `modelDescription.xml` from a Rust struct and hide the FFI. Flagged for a short evaluation spike (Q7) since its export path is newish and we'd feed it a container-driven variable set.

**Two feedback points I want to confirm I read correctly**
- On the old Q3 ("not clear why you're asking") — that question was about the causality of non-located globals. Since the interface is now I/O-only, it's moot and removed. What remains is the narrower **Q5**: should any non-I/O values be exportable as tunable FMI `parameter`s (and if so, how marked in source), or is that out of v1 scope?
- **New section for you specifically:** added **"For the Open Industry Project Maintainer"** near the top — what IronPLC ships, the interface/time/type contract, the same-host constraint, and four things we'd want from the OIP side — so the OIP maintainer can evaluate fit without reading internals.

Remaining open questions trimmed to **seven** (Q1–Q7 in the doc): internal-state observability, conformance-master target, step-size policy, multi-task scope, parameters, cross-platform packaging, and the `fmi-export` spike.

---

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,475 @@
...
+well-defined points. The recent task/scheduler and time-control work supplies
+the co-simulation time contract almost directly.
+
+The FMU's interface is the program's **I/O**: located input variables (`%I`)
+become FMU inputs and located output variables (`%Q`) become FMU outputs. This
+is the correct interaction pattern for co-simulation — the simulation drives the
+PLC's sensors and reads its actuators, exactly as physical hardware would.
+Delivering this cleanly depends on one piece of IronPLC that is specified but
+not yet built: the **process image** (the `%I`/`%Q`/`%M` regions). This design
+commits to building it (referred to below as **Path B**).
```
garretfick: Let's remove reference to Path A or Path B. We only have Path B to just say that's what we are doing without citing "Path B"

specs/design/fmi-co-simulation-support.md
garretfick: Done in `2ea43a3` — removed all "Path A"/"Path B" labels throughout (Overview, Decisions, the section heading, and the phases list) and state the process-image approach directly.

---

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,475 @@
...
+**What the interface looks like.** The FMU's variables **are the PLC's I/O
+points**:
+
+| PLC declaration | FMU variable | Causality |
+|-----------------|--------------|-----------|
+| `sensor AT %IW0 : INT;` | `sensor` | `input` (you write it before a step) |
+| `valve AT %QX0.0 : BOOL;` | `valve` | `output` (you read it after a step) |
+| tunable setpoints (see [Open Questions](#open-questions)) | as declared | `parameter` |
+
+Internal program state (`%M`, plain locals) is **not** part of the interface by
```
garretfick: For FMI can we make %M visible but not writable?

garretfick: Yes — that's exactly what FMI `local` causality gives us, and I've folded it into the design in `7d6c0c9`. A `local` variable is readable by the master via `fmi3Get*` but can never be written: FMI only permits the master to `Set` `input`s and `parameter`s. So exporting `%M` as `local` makes it **visible but not writable** with no extra enforcement on our side.

Updated: `%M` → read-only `local` is now a decision (Overview, Decisions, and the interface table), and Q1 is narrowed to the one remaining choice — whether *plain, non-located* internal locals should also be exposed read-only, or `%M` + I/O is the whole v1 surface.

---

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,475 @@
...
+simulated clock, never wall-clock time. What happens when a master requests a
+larger, smaller, or misaligned step is covered in
+[Time and Stepping](#time-and-stepping); the short version is that the FMU tells
+you its required period and will report the actual time it reached.
+
+**Type mapping.** IEC 61131-3 types map onto FMI 3.0's typed variables with no
+loss (this is the main reason for targeting 3.0 rather than 2.0). `BOOL→Boolean`,
+`SINT/INT/DINT/LINT→Int8/16/32/64`, the unsigned variants to `UInt*`,
+`REAL→Float32`, `LREAL→Float64`, `STRING→String`, and `TIME`/date-time values as
+**integer** counts (see [Type Mapping](#type-mapping)).
```
garretfick, line 83: What about other time values like date time? We need to specify all of them. I want to see a table somewhere in this doc.

specs/design/fmi-co-simulation-support.md
garretfick: Added a dedicated **Date/duration types** table in `e16074f` covering all eight IEC date/time type tags. Every one exports as an **integer** (never Float), with an explicit unit + base:

- `TIME`→`Int32` ms and `LTIME`→`Int64` ms are **settled** (ADR-0021 fixes width and millisecond unit; the VM already stores them this way).
- The six calendar types (`DATE`/`LDATE`/`TOD`/`LTOD`/`DT`/`LDT`) get concrete proposed encodings (e.g. `DATE`→`Int32` days since 1970-01-01, `DT`→`Int64` ms since 1970-01-01), but their in-memory representation isn't finalized in IronPLC yet — debug formatting only implements `TIME`/`LTIME` today. So they're marked "needs storage ADR" and I raised **Q8** for it; until that ADR lands, `compile --format fmu` rejects an interface that uses a calendar type rather than emitting an unspecified encoding.

Please sanity-check the proposed encodings in the table.

---

specs/design/fmi-co-simulation-support.md
garretfick: Done in `e16074f` — added a **Date/duration types** table under Type Mapping covering all eight IEC date/time tags, each exported as an integer with an explicit unit + base. `TIME`/`LTIME` are settled (ADR-0021: `Int32`/`Int64` ms); the six calendar types (`DATE`/`LDATE`/`TOD`/`LTOD`/`DT`/`LDT`) have proposed encodings but are flagged as needing a storage ADR (new **Q8**) since their in-memory form isn't finalized yet — export rejects them until that lands. Please sanity-check the encodings in the table.

Also captured the platform point (IronPLC ships the runtime for the target hardware, so the co-sim must run on that platform) as new **Q9** plus a reworked "Platform constraint" paragraph in the OIP section.

---

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,531 @@
...
+
+**Platform constraint.** IronPLC ships the runtime (`ironplc-fmi`) **only for the
+target hardware** — the platform the PLC program is built for. The FMU's shared
+library therefore matches that one platform, so **the co-simulation must run on
+that same platform**. If your OIP scenarios run the master on a different OS/CPU
+than the PLC's target hardware, the FMU binary won't match and we'd need a
+cross-compilation/installer story. We're treating this as an **open question and
+want your input** ([Q9](#open-questions), related [Q6](#open-questions)): tell us
+what platform(s) your co-simulation host runs on.
+
```
garretfick, line 93: We also need to explain the UX. How does someone actually use this. Explain both how to create the fmi and integration that might be possible with OIP

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,531 @@
...
+**What we would want from you.** (1) Which FMI 3.0 master(s) must the first
+release interoperate with, so we target the right conformance suite
+([Q2](#open-questions))? (2) `%M` memory is exposed read-only for observability;
```
garretfick: Also note that we are not exposing other internal variables (although we could in a follow up version).

specs/design/fmi-co-simulation-support.md
garretfick: Added a **User Experience** section in `5a15a09` with a four-step walkthrough:

1. **Author** — write ordinary ST and declare I/O as located variables (`%I`→input, `%Q`→output, `%M`→read-only local); nothing FMI-specific in the source.
2. **Create the FMU** — `ironplcc compile --format fmu ./my-project` → `my-project.fmu` (compiles the container, generates `modelDescription.xml`, zips with the prebuilt runtime).
3. **Run in any FMI 3.0 master** — a worked FMPy snippet plus how it appears as an input/output block in OMSimulator / OSP.
4. **Integrate with OIP** — the closed loop: PLC FMU's `%I` ← plant model outputs, `%Q` → plant model inputs, with an ASCII diagram. I flagged that the *exact* OIP scenario/config surface (e.g. an OSP-style `SystemStructure` file, component registration) is the one piece I can't specify from the IronPLC side — raised as **Q10** for your input.

---

specs/design/fmi-co-simulation-support.md
garretfick: Noted in `5a15a09`. The Overview and Decisions now state explicitly that **plain, non-located internal locals are not exposed in v1** — `%M` is the only internal state on the interface — with a note that a follow-up version could add opt-in read-only exposure of selected locals. I also narrowed Q1 to just the forward-looking question (whether to leave room in the interface-section format for richer debugger-style capability later).

---

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,531 @@
...
+   clock).
+4. **I/O is the interface** — FMU variables are the program's located `%I`/`%Q`
+   points, driven through the process image, not arbitrary internal variables.
+5. **Reuse, don't fork, the execution core** — FMI is a new *driver* around the
+   existing `VmRunning` state machine, like the DAP debugger. No changes to
+   bytecode or codegen semantics.
+
+## Decisions (settled in review)
+
+These were open questions in revision 1; review resolved them.
```
garretfick, line 269: Omit any resolved questions.

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,531 @@
...
+   `write_variable` (`compiler/vm/src/vm.rs:585-606`), the same external-access
+   path the DAP debugger uses. (Typed setters need widening — see
+   [phase 1](#implementation-phases).)
+3. **Per-instance buffers — clean multi-instantiation.**
+   `VmBuffers::from_container` allocates per-instance state over a read-only
+   shared `Container`; multiple FMU instances are independent.
+4. **Interface metadata partly exists.** The debug section already records
+   per-variable `var_section`, `iec_type_tag`, name, and type
+   (`compiler/container/src/debug_section.rs:11-30`) — but it is strippable, so
+   the exported interface must be promoted to a non-strippable artifact
```
garretfick: We will make this strippable. The part that won't be strippable is the IOM points

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,531 @@
...
+1. **Located-variable information becomes a first-class container artifact.**
+   The compiler records, for each located variable, its `{name, region (I/Q/M),
+   size, address, iec_type, valueReference, start value}`. This is a small,
+   non-strippable **interface section** (or an always-emitted subset of the debug
+   section). It is the single source of truth for both the process-image layout
+   and the FMU interface. *(Note: `input_image_offset`/`output_image_offset`
+   already exist in the task table but are hardcoded to `0`,
+   `compiler/container/src/builder.rs:342-343`; the process image populates
+   them.)*
+2. **The debug info gains a second emit format: the model description.** The same
```
garretfick: My thinking was that we would generate the model info at compile time. That way we don't need to change the format now.

specs/design/fmi-co-simulation-support.md
garretfick: Done in `9916aff`. Removed the two questions whose answers are now decided — old **Q1** (internal-variable observability: settled as I/O + read-only `%M`, no plain locals) and old **Q6** (same-host packaging: decided). Renumbered the remaining eight (now Q1–Q8) and updated every in-text `[Q#]` reference to match.

---

specs/design/fmi-co-simulation-support.md
garretfick: Updated in `9916aff`/`7ef3271`. The debug section stays **strippable** and is off the FMI path; the non-strippable part is the **IOM points** (the located `%I`/`%Q`/`%M` addresses), which the runtime needs for the process image anyway. This is also consistent with your compile-time point below — see that reply for how the two fit together.

---

specs/design/fmi-co-simulation-support.md
garretfick: Adopted — this simplifies things and I've rewritten the section accordingly (`9916aff`). The section is now **"Interface Metadata (generated at compile time)"**:

- `ironplcc compile --format fmu` emits `modelDescription.xml` **at compile time** directly from the located-variable (IOM) info the compiler already has — **no container-format change now**, and nothing is read back out of the compiled container.
- The strippable debug section is never on the FMI path.
- The runtime still needs the `%I`/`%Q`/`%M` layout, but that comes from the already-planned process-image work (the task table's `input_image_offset`/`output_image_offset`, currently `0`) — those IOM points are the durable, non-strippable truth. So this and your "IOM points aren't strippable" note fit together: compile-time XML generation + IOM layout carried by the process image, with no new FMI-specific format.

Updated the Decisions bullet, packaging steps, value-reference paragraph, and Building On to match.

---

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,531 @@
...
+  `parameter`s (set at initialization) — e.g. setpoints/limits — and if so, how
+  are they marked in source (a pragma/attribute), given that `%M`/globals are not
+  the interface? Or are parameters out of scope for v1 (I/O only)?
+
+- **Q6 — Cross-platform packaging.** Is same-host execution acceptable for v1
+  (FMU runs where it was compiled)? If OIP needs an FMU compiled on one machine
+  to run on another OS/CPU, we need a cross-compilation/installer approach —
+  worth scoping now?
+
+- **Q7 — `fmi-export` reuse.** Approve a short spike to evaluate rust-fmi's
```
garretfick: Omit this question

specs/design/fmi-co-simulation-support.md
```diff
@@ -0,0 +1,531 @@
...
+platform does your co-simulation host run on — is it the same as the PLC's target
+hardware ([Q9](#open-questions))?
+
+**Known v1 limitations.** Single program / single scan task (multi-task programs
+deferred, [Q4](#open-questions)); FMI **Co-Simulation** only (not Model
+Exchange); no importing of external FMUs into a PLC program; the FMU binary is
+built for the target hardware, so the co-simulation runs on that platform
+([Q9](#open-questions)); calendar date/time types await a storage ADR before they
+can appear in an FMU interface ([Q8](#open-questions)).
+
```
garretfick, line 116: What about automatic integration testing?

specs/design/fmi-co-simulation-support.md
garretfick: Done in `53a7e8e` — removed the **`fmi-export` reuse** question. It's an implementation detail (whether to reuse rust-fmi's `fmi-export`/`cargo-fmi` vs. hand-roll the shim) rather than a design decision needing sign-off, so it doesn't belong in Open Questions. The reuse recommendation itself stays in "The FMI Shim and Crate Reuse" section. Remaining questions renumbered to Q1–Q7 with all references updated.

---

specs/design/fmi-co-simulation-support.md
garretfick: It's covered — **phase 6** in Implementation Phases is an automated CI integration test: compile a PLC program → FMU → load it in a reference FMI 3.0 importer (FMPy or the rust-fmi importer) → run a multi-step scenario → assert inputs→outputs, run in the normal CI pipeline so export can't silently regress.

Since you asked from the OIP-maintainer section, I also surfaced it there (`53a7e8e`) as an **"Automated testing"** note so it's visible without digging into the phases. Is a specific importer/harness (e.g. the official FMI cross-check, or a particular OIP tool) the one you'd want that CI test to run against? That's Q1 (conformance target).

---

garretfick:
Related issue: https://github.com/Open-Industry-Project/Open-Industry-Project/issues/273

PR #1259, opened 2026-07-30
Enforce --allow-constant-type-params (P4029)

### LINKED ISSUE #1234: `--allow-constant-type-params` is not enforced: constant in type parameter compiles regardless of the flag
## Summary

The `allow_constant_type_params` feature flag (`--allow-constant-type-params`, MCP `allowConstantTypeParams`) is declared, surfaced by `list_options`, exposed on the CLI/LSP, and documented — but it is **never enforced**. A constant reference in a type parameter position (e.g. `STRING[MY_CONST]`, `ARRAY[1..MY_CONST] OF INT`) compiles cleanly whether the flag is on or off. Enabling the flag has no effect, and the construct it is supposed to gate is accepted unconditionally in strict IEC 61131-3 mode.

## Evidence

- The reserved problem code **P4029 `ConstantTypeParamNotAllowed`** ("Constant reference in type parameter requires --allow-constant-type-params flag") is declared in `compiler/problems/resources/problem-codes.csv:76` and has a doc page (`docs/reference/compiler/problems/P4029.rst`), but grepping the source shows it is **emitted nowhere** — the only non-generated occurrence is the CSV declaration.
- The parser accepts a constant in a type parameter unconditionally (no `allow_constant_type_params` check on the parse path); the constant is later resolved by `analyzer/src/xform_resolve_constant_expressions.rs`, whose entry point takes no options.
- The `allow_constant_type_params` field is only read in `#[cfg(test)]` parse helpers, never in enforcement.

## Reproduction

```st
VAR_GLOBAL CONSTANT
    STRING_LENGTH : INT := 250;
END_VAR
FUNCTION_BLOCK fb1
VAR_INPUT
    STR : STRING[STRING_LENGTH];
END_VAR
END_FUNCTION_BLOCK
PROGRAM p
END_PROGRAM
```

- **Expected** (flag off, strict Ed2): rejected with P4029 (only integer literals permitted in type parameters).
- **Actual**: compiles cleanly with `--dialect=iec61131-3-ed2` and no flag.

## Suggested fix

Emit P4029 when a non-literal (constant reference) appears in a type parameter and `allow_constant_type_params` is false. This likely belongs in the analyzer stage that resolves type parameters / constant expressions, since the check needs the flag threaded through.

## Related

A behavioral fixture already records this gap in `compiler/mcp/src/feature_flag_conformance.rs` with `Expectation::NotYetEnforced`. When enforcement lands, that fixture will need to move to `Expectation::GatesBehavior` (its test asserts the current not-enforced state, so it will fail once the flag actually gates behavior — a built-in reminder).

garretfick:
The allow_constant_type_params flag was declared, surfaced by
list_options, and documented, but never enforced: a constant reference
in a type parameter (e.g. STRING[MY_CONST], ARRAY[1..MY_CONST] OF INT)
compiled cleanly regardless of the flag, and the reserved problem code
P4029 (ConstantTypeParamNotAllowed) was emitted nowhere.

Thread the flag into xform_resolve_constant_expressions and reject any
constant reference in a type parameter with P4029 when the flag is off,
before resolution is attempted. With the flag on, references resolve as
before. Move the flag's feature_flag_conformance entry from UNENFORCED
to a behavioral fixture (rejected off / accepted on).

Fixes #1234

Claude-Session: https://claude.ai/code/session_012MgRzbx338Y7tfKhofnHk1

### DIFF
--- a/compiler/mcp/src/feature_flag_conformance.rs
+++ b/compiler/mcp/src/feature_flag_conformance.rs
@@ -19,20 +19,17 @@
 //! that flag and nothing else — the strongest form of "this example is
 //! supported, this other code is not".
 //!
-//! A few flags are declared and surfaced by `list_options` but are not yet
-//! enforced by the compiler — the construct they gate compiles whether the flag
-//! is on or off, so no off-rejects/on-accepts fixture can exist. Those are
-//! listed in [`UNENFORCED`], each tracked by an issue. The gap lives in the
-//! issue tracker, not in a special test expectation.
+//! Every feature flag must have a fixture — there is no escape hatch for
+//! "declared but not yet enforced". A flag that is surfaced by `list_options`
+//! but gates no behavior is a dead flag, and the suite fails until it either
+//! gains a fixture (enforcement lands) or is removed.
 //!
-//! Two meta-tests keep the tables honest and decoupled:
-//! - [`every_feature_flag_has_a_fixture_or_is_tracked_unenforced`] fails when a
-//!   new flag is added without either a fixture or an `UNENFORCED` entry (the
-//!   reminder lives in the suite, not a reviewer's head), mirroring
-//!   `spec_conformance::all_spec_requirements_have_tests`.
-//! - [`fixture_and_unenforced_tables_are_consistent`] fails when an entry names
-//!   a flag that does not exist, or when a flag is listed as both fixtured and
-//!   unenforced (e.g. enforcement was added but `UNENFORCED` was not updated).
+//! Two meta-tests keep the table honest and decoupled:
+//! - [`every_feature_flag_has_a_fixture`] fails when a new flag is added without
+//!   a fixture (the reminder lives in the suite, not a reviewer's head),
+//!   mirroring `spec_conformance::all_spec_requirements_have_tests`.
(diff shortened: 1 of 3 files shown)


compiler/mcp/src/feature_flag_conformance.rs
```diff
@@ -61,8 +61,8 @@ struct FlagFixture {
 /// tracked by an issue. When enforcement lands, add a fixture to
 /// [`FLAG_FIXTURES`] and remove the entry here.
 ///
-/// - `allow_constant_type_params`  → ironplc/ironplc#1234
-const UNENFORCED: &[&str] = &["allow_constant_type_params"];
+/// (none currently)
```
garretfick: Let's remove this escape clause entirely. Everything must be enforced

garretfick: Done in e6e2b99. Removed the `UNENFORCED` list and the "declared but not yet enforced" allowance entirely. `every_feature_flag_has_a_fixture` now fails for any flag without a behavioral fixture (a flag that gates nothing is a dead flag and must be removed), and `fixture_keys_name_real_flags` just validates fixture keys/prereqs against `FEATURE_DESCRIPTORS`. All three conformance tests pass.

---

PR #1268, opened 2026-07-31
feat(parser): accept STRING(n)/WSTRING(n) parenthesis length form

garretfick:
## Summary

CODESYS/TwinCAT accept `STRING(n)`/`WSTRING(n)` with parentheses as an alternate delimiter to the standard `STRING[n]`/`WSTRING[n]` bracket form. This adds parenthesis support in `VAR` declarations and function return types, **gated behind a vendor-extension flag**.

```
FUNCTION_BLOCK FB_Example
VAR
    hostName : STRING(255);   // parse error in strict mode; accepted under --allow-paren-string-length
END_VAR
END_FUNCTION_BLOCK
```

## Not standard IEC 61131-3 → flag-gated

IEC 61131-3 (both Ed. 2 and Ed. 3, Annex B) declares a string length **only** with square brackets (`'STRING' ['[' unsigned_int ']']`). The parenthesis form is purely a CODESYS/TwinCAT extension, so it is gated exactly like every other vendor extension in this compiler:

- New flag `allow_paren_string_length` (`--allow-paren-string-length`), enabled by the `rusty`, `codesys`, and `twincat` dialects; **off** in the strict `iec61131-3-ed2`/`iec61131-3-ed3` defaults.
- Gating follows the `rule_token_no_partial_access_syntax` precedent: the grammar parses the paren form permissively (new shared `string_length_spec()` rule) and a new token-stream rule (`rule_token_no_paren_string_length`) rejects it with **P4042** when the flag is off. A `STRING`/`WSTRING` keyword followed by `(` is an unambiguous length delimiter, so nothing standard is mis-flagged.
- This also closes a pre-existing gap: the `TYPE`-alias paren rule (`string_type_declaration__parenthesis`) was previously accepted unconditionally, so parentheses are now rejected uniformly in strict mode wherever a string length appears.

No DSL or renderer change: the DSL stores only `length: Option<IntegerRef>` with no delimiter marker, and the renderer already normalizes to brackets — so `STRING(255)` round-trips to `STRING [ 255 ]`.

## Split from #1222

This is **split out of #1222**, which the maintainer asked to break up because it bundled two unrelated syntax features. This PR is the **standalone sized-string half**, and it's **part of #1199 (TwinCAT dialect)**. It has **no functional dependency** on the other half (the inline FB-instance call-style initializer).

## Changes

- `compiler/parser/src/parser.rs` — shared `string_length_spec()` rule routed through `single_byte_string_spec()`, `double_byte_string_spec()`, the `var_spec()` and `function_return_type()` String/WString arms.
- `compiler/parser/src/options.rs` — new `allow_paren_string_length` flag (rusty/codesys/twincat) + dialect assertion tests.
- `compiler/parser/src/rule_token_no_paren_string_length.rs` + `lib.rs` — token-stream gate emitting P4042.
- `compiler/problems/resources/problem-codes.csv` + `docs/reference/compiler/problems/P4042.rst` (+ index) — problem code and docs.
- `compiler/ironplc-cli/bin/main.rs` — CLI flag; `compiler/mcp/src/feature_flag_conformance.rs` — off→on fixture.
- Pre-existing `STRING(n)` test inputs (a few analyzer tests + one resource) switched to the standard `STRING[n]` form so they still parse under the strict default.

## Tests

- Parser: `STRING(255)`/`WSTRING(100)` parse under the flag in a `VAR` declaration and a `FUNCTION` return type; strict default dialect rejects the paren form (P4042); bracket form regression.
- Token rule: flag on/off, both keywords, trivia between keyword and `(`, bracket form stays allowed.
- plc2plc: `STRING(255)` (flag on) renders as `STRING [ 255 ]` and the normalized output round-trips under the strict default.

## Verification

Full CI pipeline passes locally (`cd compiler && just`): build, coverage ≥ 85%, clippy, and fmt all green.

https://claude.ai/code/session_013HfEswQaNUvzVgk2S7iBJS


compiler/ironplc-cli/bin/main.rs
```diff
@@ -179,6 +179,12 @@ struct FileArgs {
     /// standard.
     #[arg(long)]
     allow_bit_string_case_labels: bool,
+
+    /// Allow the STRING(n)/WSTRING(n) parenthesis length delimiter instead of
```
garretfick: This says "instead of" but this is really "in addition to"

compiler/ironplc-cli/bin/main.rs
```diff
@@ -179,6 +179,12 @@ struct FileArgs {
...
+    /// the standard STRING[n]/WSTRING[n] brackets. This is a vendor extension
+    /// (CODESYS/TwinCAT) not part of the IEC 61131-3 standard.
```
garretfick: We should not be mentioning which vendor support this here. Just have that determined based on the dialect mapping.

compiler/parser/src/rule_token_no_paren_string_length.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use crate::{
+    options::CompilerOptions,
+    token::{Token, TokenType},
+};
+
+fn is_trivia(t: &TokenType) -> bool {
+    matches!(
+        t,
+        TokenType::Whitespace | TokenType::Newline | TokenType::Comment
```
garretfick: What about pragmas - those are also trivia

docs/reference/compiler/problems/P4042.rst
```diff
@@ -0,0 +1,45 @@
...
+
+.. code-block::
+
+   FUNCTION_BLOCK FB_Example
+   VAR
+       hostName : STRING[255];
+   END_VAR
+   END_FUNCTION_BLOCK
+
+Or enable the vendor extension with the ``--allow-paren-string-length``
```
garretfick, line 43: Do not cite dialects here. Only cite an `allow` and then point to the docs on vendor extension. Please also update the language support guide so that we remember this decision.

compiler/ironplc-cli/bin/main.rs
garretfick: Fixed — reworded to "in addition to the standard STRING[n]/WSTRING[n] brackets" (both the CLI help and the `FEATURE_DESCRIPTORS` description).

---

compiler/ironplc-cli/bin/main.rs
garretfick: Done — removed the `(CODESYS/TwinCAT)` mention from the flag help. Which vendors/dialects enable it is left to the dialect mapping (`enabling-dialects-and-features.rst` and `describe_dialects()`).

---

compiler/parser/src/rule_token_no_paren_string_length.rs
garretfick: Good catch — the grammar's `_` rule skips collapsed `Pragma` tokens too, so the gate was under-enforcing: with `allow_pragmas` on and `allow_paren_string_length` off, `STRING {attribute 'x'} (255)` parsed yet slipped past. Added `Pragma` to the gate's trivia set so it matches `_` exactly, plus unit tests for newline/comment/pragma trivia (only whitespace was covered before) and two end-to-end tests through the full pragma-collapse + token-check + grammar pipeline (commit 9d83008).

---

docs/reference/compiler/problems/P4042.rst
garretfick: Done — P4042 now cites only `--allow-paren-string-length` and links `/explanation/enabling-dialects-and-features` for how to enable vendor extensions and which dialects include them, rather than listing the dialects inline. Also recorded the decision in the flag table in `specs/steering/syntax-support-guide.md`, and added the flag to the two docs the guide's checklist requires that I'd missed: `docs/explanation/enabling-dialects-and-features.rst` (dialect "Enables:" lists + a per-feature entry) and `docs/reference/compiler/ironplcc.rst`.

---

PR #1270, opened 2026-07-31
feat(parser): inline FB-instance call-style initializer

garretfick:
## Summary

Adds support for the CODESYS/TwinCAT **call-style function-block instance initializer** — passing an initialization parameter list directly after the type name, using the same positional-or-named parameter shape as an ordinary FB call:

```iecst
FUNCTION_BLOCK FB_Example
VAR
    comm : FB_Comm(retries := 3, THIS);   // previously a parse error -- only `FB_Comm := (...)` parsed
END_VAR
END_FUNCTION_BLOCK
```

This is **split out of #1222** (which bundled two unrelated syntax gaps) and is **part of #1199** (TwinCAT dialect support). The `STRING(n)`/`WSTRING(n)` parenthesis-length half of #1222 is a separate PR.

## Design notes

- **No new dialect flag.** `(` is already a token, and codegen already silently ignores FB instance initializer values for the existing `:= (member := value, ...)` form (only `type_name` drives memory layout). Gating *only* the new form as "recognized but unsupported" would be inconsistent — so both forms parse the same permissive way. The call-style argument list is parsed and stored but not yet wired into codegen, matching `init`'s own pre-existing status.
- **New optional DSL field** `FunctionBlockInitialValueAssignment.call_params: Option<Vec<ParamAssignmentKind>>` (`#[recurse(ignore)]`) — `Some` only for the call-style form, `None` for the `:=` struct-init form or a bare declaration.
- **Dedicated grammar rule.** Uses a new `fb_call_style_var_decl()` built on the working `var1_list()` combinator and requiring the parens unconditionally (so it can never match a bare `name : Type;`, which continues through the late-bound path). It deliberately does **not** use the pre-existing dead `fb_name_decl()` rule, whose `commasep_oneplus()`-based name list requires a spurious trailing comma and never matches real input — fixing that combinator is out of scope.
- **plc2plc** renders the call params back in parentheses; round-trips through parse → render → parse.

## Dependency

Depends on the toposort edge-direction fix **#1269** (also split from #1222, now merged into `main`). Constructing an eager `InitialValueAssignmentKind::FunctionBlock` for a user-declared FB type — which the call-style grammar does at parse time — surfaces a spurious `P2011` "Parent type is not declared" without that fix. This PR **does not** re-include the production edge flip; it **does** include the call-style-syntax tests that exercise the fixed ordering (an end-to-end test in `stages.rs` and a toposort-level test).

## Changes

| File | Change |
|------|--------|
| `compiler/dsl/src/common.rs` | New `call_params` field + `VarDecl::function_block` construction site |
| `compiler/parser/src/parser.rs` | New `fb_call_style_var_decl()` / `fb_call_style_init_params()` rules; added to `var_init_decl()`; `call_params: None` at existing sites |
| `compiler/analyzer/src/xform_resolve_late_bound_type_initializer.rs` | `call_params: None` at 3 construction sites |
| `compiler/plc2plc/src/renderer.rs` | Render call params when `Some` |
| `specs/plans/2026-07-31-twincat-inline-fb-call-style-initializer.md` | Implementation plan |

## Tests

- **Parser:** mixed named+positional call params; empty parens; bare declaration still `LateResolvedType` (regression); `:= (...)` struct-init still parses (regression).
- **plc2plc:** call-style initializer round-trips.
- **Analyzer:** end-to-end no-diagnostics for a call-style initializer referencing an earlier-declared FB (P2011 regression, green with #1269 merged); toposort ordering test.

Full CI pipeline (`cd compiler && just` — compile, coverage ≥85%, clippy, fmt, dupes) passes locally.

https://claude.ai/code/session_01MVvRfY3XHuutbjewWKX5v3

---
_Generated by [Claude Code](https://claude.ai/code/session_01MVvRfY3XHuutbjewWKX5v3)_

### DIFF
--- a/compiler/analyzer/src/stages.rs
+++ b/compiler/analyzer/src/stages.rs
@@ -453,4 +453,41 @@ END_FUNCTION_BLOCK";
 
         assert!(context.has_diagnostics());
     }
+
+    // ---------------------------------------------------------------------
+    // FB-instance call-style initializer.
+    // See specs/plans/2026-07-31-twincat-inline-fb-call-style-initializer.md.
+    // ---------------------------------------------------------------------
+
+    #[test]
+    fn analyze_when_fb_call_style_init_references_earlier_declared_fb_then_no_diagnostics() {
+        // End-to-end regression for the toposort edge-direction bug found
+        // while adding this feature: `comm : FB_Comm(retries := 3);`
+        // constructs InitialValueAssignmentKind::FunctionBlock directly at
+        // parse time (unlike a bare declaration, which defers to
+        // LateResolvedType). Before the xform_toposort_declarations.rs fix,
+        // this produced a spurious P2011 "Parent type is not declared"
+        // because the referenced type was ordered after its referencing
+        // POU instead of before it.
+        let program = "
+FUNCTION_BLOCK FB_Comm
+VAR_INPUT
+    retries : INT;
+END_VAR
+END_FUNCTION_BLOCK
+
+FUNCTION_BLOCK FB_Example
+VAR
+    comm : FB_Comm(retries := 3);
+END_VAR
+END_FUNCTION_BLOCK";
+        let lib = parse_program(program, &FileId::default(), &CompilerOptions::default()).unwrap();
+        let (_library, context) = analyze(&[&lib], &CompilerOptions::default()).unwrap();
+
+        assert!(
+            !context.has_diagnostics(),
+            "unexpected diagnostics: {:?}",
+            context.diagnostics()
+        );
+    }
 }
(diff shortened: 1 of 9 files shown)


garretfick:
Superseded by #1273 and closing in its favor.

This PR modeled the call-style FB initializer as an `Option<Vec<ParamAssignmentKind>>` field on the existing `FunctionBlockInitialValueAssignment` node. Since the call-style form (`FB_Type(args)`, passed to the FB's `FB_init` constructor) and the standard `:= (member := value)` member-init form are semantically distinct constructs, that field is effectively a discriminant flag. #1273 instead models the call form as its own AST node (`InitialValueAssignmentKind::FunctionBlockCall`), which makes the two forms impossible to conflate and lets exhaustiveness checking drive every consumer. It also adds a deferring semantic rule (P9004) so the construct is recognized but reported as not-yet-code-generated rather than silently ignored.

No unique work is lost — #1273 re-implements the feature from `main`.

---

PR #1273, opened 2026-08-01
feat(parser): model FB call-style initializer as a distinct AST node

garretfick:
## Summary

Adds support for the CODESYS/TwinCAT **call-style function-block instance initializer** — an argument list passed directly after the type name, with no `:=`:

```iecst
FUNCTION_BLOCK FB_Example
VAR
    comm : FB_Comm(retries := 3, THIS);   -- previously a parse error
END_VAR
END_FUNCTION_BLOCK
```

Part of #1199 (TwinCAT dialect support). **Supersedes #1270**, which modeled the feature as an `Option<Vec<ParamAssignmentKind>>` field on the existing node.

## Why a distinct node (the change from #1270)

The call-style form and the standard `:= (member := value, ...)` form are **semantically different constructs**, not two spellings of one thing (per the CODESYS/Beckhoff `FB_init` docs):

- `inst : FB_Type := (member := value)` — sets the instance's own member/`VAR_INPUT` values.
- `inst : FB_Type(arg, ...)` — passes arguments to the FB's **constructor** (`FB_init` method); positional args like `THIS` cannot be expressed in the member-init grammar at all.

Modeling the second as an optional field on the first is a discriminant flag — it encodes "which of two meanings" in a field value rather than in the type, forcing every consumer to honor a by-convention "mutually exclusive" invariant and leaving the ambiguity live at codegen. This PR instead adds a dedicated variant:

```rust
InitialValueAssignmentKind::FunctionBlockCall(FunctionBlockCallInitializer { type_name, params })
```

Two meanings → two nodes: illegal states are unrepresentable, exhaustiveness forces every `match` to decide what the call form means, and codegen dispatches on the variant with no field-inspection ambiguity.

## Deferring semantic rule (recognized, not yet code-generated)

ironplc models no `FB_init` method today. Rather than silently ignore the arguments, the construct is parsed and preserved, and a dedicated semantic rule (`rule_function_block_call_unsupported`) reports a new documented problem code **P9004** ("recognized but not yet supported in code generation"). This mirrors `rule_unsupported_stdlib_type` / P9001. When `FB_init` codegen is implemented, the rule is removed and a codegen arm fills in — a one-place change.

**Scope:** params-only. The combined CODESYS form (constructor args *and* a `:=` member list on one declaration) is out of scope.

## Changes

| Area | Change |
|------|--------|
| `dsl/common.rs`, `dsl/visitor.rs`, `dsl/fold.rs` | New `FunctionBlockCallInitializer` node + `FunctionBlockCall` enum variant + `dispatch!` wiring |
| `parser/parser.rs` | `fb_call_style_var_decl()` builds the distinct node; added to `var_init_decl()` |
| `analyzer/` | Handle the new node in toposort (dependency edge), type/expr resolution, and const rules — so a call-style program yields only P9004, not spurious P2011/unknown-type |
| `analyzer/rule_function_block_call_unsupported.rs` | New deferring rule → P9004 |
| `problems/…/problem-codes.csv`, `docs/…/P9004.rst` | New problem code + docs |
| `plc2plc/renderer.rs` | `visit_function_block_call_initializer` renders `Type ( params )` |
| `mcp/pou_lineage.rs` | Records the FB-type lineage edge for the call form |

## Tests

- Parser: call form → `FunctionBlockCall` (named + positional); empty parens; bare decl still `LateResolvedType`; `:=` member form is *not* a call node (regressions).
- plc2plc: call form round-trips.
- Analyzer: deferring rule yields exactly P9004; member/bare forms are not flagged; end-to-end a forward-referencing call form yields P9004 and **not** P2011.
- Toposort: referenced type ordered before the referencing POU.

Full CI (`cd compiler && just` — compile, coverage ≥85%, clippy, fmt, dupes) passes locally.

https://claude.ai/code/session_01MVvRfY3XHuutbjewWKX5v3

---
_Generated by [Claude Code](https://claude.ai/code/session_01MVvRfY3XHuutbjewWKX5v3)_


compiler/dsl/src/common.rs
```diff
@@ -2594,6 +2604,29 @@ pub struct FunctionBlockInitialValueAssignment {
...
+/// positional-or-named shape as an ordinary function-block call
+/// (`param_assignment`), because in CODESYS these arguments are passed to
+/// the function block's constructor (`FB_init` method).
+///
+/// This is deliberately a separate node from
+/// [`FunctionBlockInitialValueAssignment`] (whose `init` sets the
+/// instance's own member values via the `:= (member := value, ...)` form):
+/// the two are different constructs that target different things, so
+/// keeping them as distinct AST nodes prevents them from being conflated
+/// at code generation. Not yet supported in codegen -- flagged by
```
garretfick: Can we remove the comment here about what's supported. We'll likely forget to update this later

compiler/parser/src/parser.rs
```diff
@@ -996,6 +996,35 @@ parser! {
...
+    // CODESYS/TwinCAT FB instance declaration with a call-style
+    // initialization parameter list (`FB_Type(args)`, no `:=`), using the
+    // same positional-or-named shape as an ordinary FB call -- e.g.
+    // `comm : FB_Comm(retries := 3, THIS);`. In CODESYS these arguments are
+    // passed to the function block's constructor (FB_init method), which is
+    // a distinct construct from the `:= (member := value)` member-init form
+    // -- hence a distinct AST node (InitialValueAssignmentKind::
+    // FunctionBlockCall), not a flag on FunctionBlockInitialValueAssignment.
+    //
+    // Deliberately uses var1_list() (not fb_name_list(), which has a
```
garretfick: Is this a bug that affects the general grammar? If so, can we fix that bug so we can use the right rule?

compiler/problems/resources/problem-codes.csv
```diff
@@ -100,5 +100,6 @@ P8001,McpInputValidation,MCP tool input validation error
 P9001,UnsupportedStdLibType,Referenced type is valid but not implemented
 P9002,NoContent,Set of valid source files has no content
 P9003,XmlBodyTypeNotSupported,POU body language is not supported
+P9004,FunctionBlockCallInitUnsupported,Function block call-style initializer is recognized but not yet supported in code generation
```
garretfick: I want this to be P9999. Not a new code.

garretfick (commented):
Please address feedback

compiler/dsl/src/common.rs
garretfick: Done — removed the "Not yet supported in codegen…" sentence from the `FunctionBlockCallInitializer` doc comment so it can't drift. The support status now lives only in the deferring rule (`rule_function_block_call_unsupported`), which is the thing that would actually change when codegen lands.

---

compiler/parser/src/parser.rs
garretfick: Confirmed genuine and fixed generally, separately. `commasep_oneplus` required a spurious trailing comma, which also broke `prog_conf_elements` (program-configuration element lists — no test covered it) and rendered `fb_name_decl` dead. The naive one-line fix regressed 6 tests because it made the unsound `fb_name_decl` reachable (it eagerly tags bare `x : SomeType` as `FunctionBlock` instead of deferring via `LateResolvedType`).

That was fixed holistically in #1275 (remove the trailing comma + remove the unsound `fb_name_decl`/`fb_name_list`, keeping `fb_name()`). This PR is now rebased onto that: the call-style rule simply uses `var1_list()` as the correct rule, and the workaround comment is gone.

---

compiler/problems/resources/problem-codes.csv
garretfick: Done — dropped the new P9004 code (CSV row + `P9004.rst`). The deferring rule now emits **P9999 (`NotImplemented`)** via `Diagnostic::not_implemented`, which keeps the IEC 61131-3 source span while recording the compiler call site for the P9xxx dashboards. (Also: P9004 is already reserved for the planned `UnsupportedExtension` framework in the beckhoff/siemens design docs, so reusing P9999 avoids that collision.) Tracked the deferred codegen in #1277.

---

Issue #962, opened 2026-04-22
VM: honour container `max_call_depth` and cover it with a spec test


garretfick:
## Background

The stack-overflow fix in #TBD (branch `claude/fix-bytecode-stack-overflow-LNAds`) added a call-depth guard in `compiler/vm/src/vm.rs` that returns `Trap::CallStackOverflow` at both recursive call sites (CALL and user FB_CALL). The guard uses a hardcoded constant:

```rust
const MAX_CALL_DEPTH: u32 = 32;
```

That's enough to keep the `execute_when_arbitrary_bytecode_then_never_panics` proptest honest, but it doesn't match the spec.

## Spec gap

`specs/design/bytecode-container-format.md:87` defines a `max_call_depth: u16` field at offset 194 of the container header, and the RAM budget formula on line 106 explicitly includes `(max_call_depth × frame_size)` — i.e. the limit is meant to be **per-program**, not a VM-wide constant.

`specs/design/runtime-execution-model.md:407` also already specifies the trap contract:

> `CALL_DEPTH_EXCEEDED | CALL, FB_CALL | Call stack depth exceeds max_call_depth`

So the behaviour is specified; the VM just isn't reading the field.

## Work

1. Read `container.max_call_depth` at VM load time and thread it (or the resolved bound) through `execute` / `execute_with_hook` in place of the `MAX_CALL_DEPTH` constant. Remove the constant.
2. Decide and document a sensible minimum / default if the header field is 0 (load-time reject vs. fallback).
3. Add a `REQ-RT-###` (or appropriate prefix) ID next to the `CALL_DEPTH_EXCEEDED` row in `runtime-execution-model.md` so the spec-test macro can bind to it. See `specs/steering/development-standards.md:73` and `specs/design/spec-conformance-testing.md` for the convention.
4. Convert `execute_when_call_recursion_exceeds_max_depth_then_traps_call_stack_overflow` in `compiler/vm/tests/execute_stack_overflow.rs` into a `#[spec_test(REQ_RT_###)]`, parameterising the container's `max_call_depth` so the test verifies the per-program contract rather than the current hardcoded value.
5. Audit existing container fixtures / `ContainerBuilder` defaults to make sure `max_call_depth` is populated; add to the builder API if missing.

## Acceptance

- No hardcoded call-depth constant in `compiler/vm/src/vm.rs`.
[...]

PR #1274, opened 2026-08-01
test(vm): bind CALL_DEPTH_EXCEEDED to a spec-conformance requirement

garretfick:
The VM already honours the per-program container `max_call_depth` (the
hardcoded `MAX_CALL_DEPTH` constant is gone and a zero depth is rejected
at load), but the `CALL_DEPTH_EXCEEDED` trap contract had no `REQ-` ID
and no `#[spec_test]` binding, so issue #962's spec-test half was still
open.

- Add `REQ-RT-vm-001` to the Trap Sources table in
  runtime-execution-model.md (new `RT` area, `vm` crate slug), via a
  Requirement column matching the existing house style.
- Wire the `ironplc-vm` crate into the spec-conformance pipeline:
  `build.rs` lists the doc through `spec_requirements_gen`; add the
  `spec_requirements` module and `all_spec_requirements_have_tests`
  meta-test to the `it` integration binary.
- Convert the self-recursion trap test to `#[spec_test(REQ_RT_vm_001)]`,
  parameterised over declared depths [2, 4, 8, 32] so it verifies the
  per-program contract rather than a hardcoded value.

Claude-Session: https://claude.ai/code/session_011QBGNjb5bsPcbxA3v8g9Nv

### DIFF
--- a/compiler/vm/tests/it/main.rs
+++ b/compiler/vm/tests/it/main.rs
@@ -11,6 +11,26 @@
 
 mod common;
 
+/// Spec-conformance requirements generated from `specs/design/runtime-execution-model.md`.
+/// Referenced by `#[spec_test(REQ_RT_vm_NNN)]`. See `vm/build.rs`.
+#[allow(dead_code)]
+mod spec_requirements {
+    include!(concat!(env!("OUT_DIR"), "/spec_requirements.rs"));
+}
+
+/// Meta-test: every `REQ-RT-vm-NNN` requirement in
+/// `specs/design/runtime-execution-model.md` has a `#[spec_test(...)]`
+/// somewhere in this crate's `src/` or `tests/`. The build script populates
+/// `UNTESTED` from the files it scans.
+#[test]
+fn all_spec_requirements_have_tests() {
+    assert!(
+        spec_requirements::UNTESTED.is_empty(),
+        "Requirements in spec with no conformance test: {:?}",
+        spec_requirements::UNTESTED
+    );
+}
+
 mod debug_engine;
 mod execute_add_i32;
 mod execute_arith_f32;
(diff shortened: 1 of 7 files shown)


compiler/vm/tests/it/main.rs
```diff
@@ -11,6 +11,26 @@
 
 mod common;
 
+/// Spec-conformance requirements generated from `specs/design/runtime-execution-model.md`.
+/// Referenced by `#[spec_test(REQ_RT_vm_NNN)]`. See `vm/build.rs`.
+#[allow(dead_code)]
```
garretfick: Can we avoid the allow dead allow by putting this into a test cfg section?

garretfick: Good catch — turned out the blanket `#[allow(dead_code)]` was simply redundant, so I removed it entirely (19445f4) rather than gate the module.

The build script already emits a targeted `#[allow(dead_code, non_upper_case_globals)]` on every item it generates (each `REQ_*` const, `ALL`, and `UNTESTED`), so nothing inside the module warns on its own. The module itself is used (`spec_requirements::UNTESTED` in the meta-test, and the `REQ_*` const via the `#[spec_test]` expansion), so the module-level allow was covering nothing.

A `#[cfg(test)]` gate wouldn't have helped here: files under `tests/` are only ever compiled as the integration-test binary (always under `--test`), so there's no non-test build to exclude, and `cfg(test)` doesn't suppress `dead_code` anyway. `cargo clippy -p ironplc-vm --tests` is clean with the attribute gone.

Note this leaves the vm test binary slightly cleaner than the `vm-cli` equivalent, which still carries the module-level allow — happy to send a follow-up to drop it there too if you'd like consistency.

---

PR #1266, opened 2026-07-31
docs(design): ADR + plan for directly-represented variables via process image

garretfick:
Records the decision (ADR-0037) to implement %I/%Q/%M located variables
against the specified process image rather than flat variable-table slots,
aligning with the runtime execution model and the FMI co-simulation design
(PR #1258). Adds a phased implementation plan that fixes the anonymous
inline direct-address gap (compile_expr.rs#L867) as a natural consequence.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Y8WL4Tn1m9wZFmknF6bX7W

PR #1278, opened 2026-08-02
docs(adr): ADR-0041 — scope to Phase 1, defer dynamic dispatch on safety grounds

garretfick:
## Summary

This PR takes the ADR-0041 proposal from #1242 (authored by @thusser, **original commit preserved as-is**) and layers a review revision on top, addressing alignment gaps with IronPLC's ADRs — chiefly the safety-first principle (ADR-0005) and the bytecode-verification requirement (ADR-0006).

It is opened as a **new** PR because #1242 lives on a fork branch that can't be pushed to from here. Commit 1 is the original ADR; commit 2 is the review revision.

## What the review found

The Phase 1 reasoning (static dispatch) is sound and well-grounded. The concerns are Phase 2 and some factual claims.

### 1. Phase 2 (dynamic dispatch) isn't reconciled with the safety ADRs
Phase 2's mechanism — *load an instance's runtime type tag, index a table, dispatch* — is exactly the "polymorphic dispatch with a runtime tag" shape that **ADR-0005 explicitly rejects** (its STRING/WSTRING example: a correct runtime tag is "an assumption, not a proof"), and it is precisely what **ADR-0006**'s verifier must check statically (call-target validity, stack-type consistency). The original ADR invoked ADR-0005 only on the narrow static-memory axis and never mentioned ADR-0006, despite introducing a new indirect-call opcode.

### 2. Factual claims that don't hold against `main`
- `ABSTRACT` enforcement was attributed to `P4040`; `P4040` is `ConstantExpressionOverflow`.
- `P9004` (the OOP gate), the OOP foundation, and `collect_all_fb_members` live in the unmerged foundation PRs (#1247 and siblings), not `main` — but were stated as current fact.
- ADR-0026 was cited for a "pay only for what you use" framing; it actually chose a *uniform* 8-byte slot layout.

### 3. Convention
- Missing `ADR-0041:` title prefix (the 0036–0040 convention).

## Changes in the review commit

- **Decide only Phase 1.** Phase 2 is demoted from a chosen direction to a deferred open question, requiring its own ADR that supplies the ADR-0005 safety weighing and the ADR-0006 verification story (or finds a more verifiable mechanism for interface-typed values) before any implementation.
- **Add an honest "Relationship to ADR-0004/0005/0006" section** naming the verification tension head-on.
- **Correct the factual claims** above (P4040/ABSTRACT, P9004/foundation/`collect_all_fb_members` marked as not-yet-in-`main`, ADR-0026 attribution).
- **Add the `ADR-0041:` title prefix.**

## Notes for the maintainer

- This revises another contributor's proposed decision, so it should be reconciled with #1242 rather than merged blindly — the two shouldn't both land.
- `0041` is free in `main` today (max is `0040`), but the numbering process is collision-prone (`main` already has duplicate `0014/0021/0022`; concurrent #1266 re-uses `0037`). Worth confirming no other in-flight PR claims `0041`.
- Docs-only change (ADR under `specs/`, outside the Sphinx build), so no Rust/CI surface is affected.

---
_Generated by [Claude Code](https://claude.ai/code/session_019LYzcDi8t3zMBxZyMXmBDz)_


specs/adrs/0041-staged-method-and-interface-dispatch.md
```diff
@@ -0,0 +1,285 @@
+# ADR-0041: Staged Method/Property Dispatch and Interface Values
+
+status: proposed
+date: 2026-07-27
+
+## Context and Problem Statement
+
+The CODESYS/TwinCAT OOP vendor extensions (`EXTENDS`, `IMPLEMENTS`,
```
garretfick: Note - at of the time, these elements don't actually parse. We can omit a statement of what's currently supported.

specs/adrs/0041-staged-method-and-interface-dispatch.md
```diff
@@ -0,0 +1,285 @@
...
+  *type*, assigned once during compilation, never stored in the
+  instance's own memory.
+- `iec_type_tag` (`compiler/codegen/src/compile_setup.rs`) is unrelated
+  debug-info metadata used for naming variables in tooling output --
+  not a real type tag usable for dispatch.
+- Function block instances already live in the shared 8-byte-slot data
+  region (ADR-0026), addressed via field offsets resolved entirely at
+  compile time (ADR-0027).
+
+"OOP dispatch" is actually two structurally different problems:
```
garretfick: We can probably start from about here. Say that OOP is a thing in 61131-3 (some dialects), then say there are variations of the idea.

specs/adrs/0041-staged-method-and-interface-dispatch.md
garretfick: Done in c6cd9c3. Removed the "these parse fully / are P9004-gated / not yet merged" statement of current support, and dropped the other current-status notes further down (the foundation-PR references and the P4040/ABSTRACT note). The ADR now states only the dispatch mechanism, not what's supported today.

---

specs/adrs/0041-staged-method-and-interface-dispatch.md
garretfick: Done in c6cd9c3. The Context now opens by saying OOP is an IEC 61131-3 feature in some dialects (CODESYS/TwinCAT), then goes straight into the two variations of dispatch — essentially starting from where you pointed. The old current-support preamble and the verified-codegen block are gone.

---

PR #1279, opened 2026-08-02
fix(sources): merge multiple .plcproj sub-projects into one compilation unit

thusser:
## Summary

Part of #1199 (TwinCAT vendor dialect support) -- addresses the cross-project type resolution gap (P2008) I flagged there as the next-biggest bucket in the corpus, right after the OOP gate.

`detect_twincat` in `compiler/sources/src/discovery/mod.rs` already searches recursively and finds *every* `.plcproj` under a directory tree, but then keeps only the first one (sorted by path) and silently discards the rest:

```rust
let mut candidates: Vec<PathBuf> = files....collect();
candidates.sort();
let plcproj_path = candidates.into_iter().next()?;
```

A real TwinCAT solution commonly has more than one `.plcproj` -- a main PLC project plus one or more library/shared sub-projects. Only ever loading one means types declared in the others are never parsed at all, so any reference to them from the loaded project fails type resolution with P2008.

This mirrors the fix already applied to LSP workspace folders in #1216 ("a multi-sub-project TwinCAT solution needs all of them loaded together for cross-project type resolution") -- merging every sub-project into one compilation unit rather than using only the first, just applied one level down inside `discover()` itself so it also covers `ironplcc check`/CLI usage, not just the LSP.

## A wrinkle: same-directory duplicates are a different case

An existing test (`discover_when_multiple_plcproj_candidates_then_picks_deterministically`) documents a different real corpus pattern: two `.plcproj` in the *same* directory, identified as a stale duplicate/rename artifact rather than a second sub-project. Merging blindly by file count would double-load that case.

The fix distinguishes them by grouping candidates by directory first: keep only the first (sorted) `.plcproj` per directory (preserving today's behavior for the duplicate case exactly), then merge across the resulting per-directory-deduplicated set (the new behavior, for genuine sub-projects in different directories).

## Other details

- `root_dir`: unchanged (the `.plcproj`'s own directory) when exactly one is selected -- an existing test depends on this for resolving a `.plcproj`'s own further-nested `<Compile>` paths. Falls back to the top-level directory passed to `discover()` when more than one is merged (nothing downstream consumes this beyond tests).
- A file referenced by more than one sub-project (a shared dependency) is deduplicated by canonical path, so it's only loaded/declared once instead of producing a spurious duplicate-declaration error.

## Test plan
- [x] New tests: multiple `.plcproj` in different directories are merged; `root_dir` falls back correctly; a file shared between two sub-projects is deduplicated.
- [x] Existing same-directory-duplicate test and all other existing discovery tests pass unchanged.
- [x] `cargo test --workspace` -- all passing
- [x] `cd compiler && just` (compile, coverage ≥85%, clippy, fmt, dupes) -- all green


compiler/sources/src/discovery/mod.rs
```diff
@@ -8,6 +8,7 @@
 //! The first match wins.
 
 use std::{
+    collections::HashSet,
```
garretfick, line 11: I'm OK with this to move forward, but is there a reason this doesn't instead start with the .sln file? The solution file should list all of the projects in the solution. Searching for .plcproj files is a hack around how I think Visual Studio actually works.

thusser: Good catch, and you're right that it's not just a hack — I checked one of my own solutions (MONETN) and found the exact failure mode: the `MONETNRuntime` folder has both `MONETNRuntime.plcproj` and a stale `MONETRuntime.plcproj` left over from a project rename. The current dedup only picks the right one because "MONETNRuntime" happens to sort before "MONETRuntime" alphabetically — that's luck, not correctness. The `.tsproj` is what actually says which one is live (`PrjFilePath="MONETNRuntime\MONETNRuntime.plcproj"`), so `.sln → .tsproj → PrjFilePath` is the only way to resolve this unambiguously.

I'd rather not fold that into this PR though — it's a bigger change (new `.sln` line-format parser plus `.tsproj` XML parsing, with the current recursive walk kept as a fallback for directories without a `.sln`). This PR fixes a real bug (multi-plcproj merging) for the common case. Let me follow up with a separate PR for `.sln`/`.tsproj`-based discovery so the two aren't coupled. Sound OK?

garretfick: I'll merge this once the checks pass.

PR #1285, opened 2026-08-02
fix: don't revert whole library when one unrelated declaration fails to resolve

thusser:
## Summary
- `xform_resolve_late_bound_type_initializer` accumulated per-declaration diagnostics while folding the whole merged library, then converted a non-empty diagnostics list into a hard `Err` — discarding the already-successfully-resolved library. `resolve_types()` then reverted the *entire* library on that `Err`, undoing type resolution for every unrelated declaration because of one genuinely broken one elsewhere.
- Confirmed with a minimal repro: an unrelated, valid `FUNCTION_BLOCK` instance variable was collaterally misreported as "not a variable in scope" (`FunctionBlockNotInScope`) purely because a different, unrelated `FUNCTION_BLOCK` referenced an undeclared type.
- Fixed by having the pass return `Ok((Library, diagnostics))` instead of discarding the result on soft diagnostics; only a genuine fold failure remains a hard `Err`. `xform_resolve_late_bound_expr_kind` shares the exact same shape (though its diagnostics accumulator turned out to be dead code — zero behavior change) and was fixed alongside it since both live in the same `recoverable_xforms` list.
- Audited the rest of `resolve_types()`'s recoverable passes for the same "accumulate diagnostics, then discard on non-empty" shape. Two are unaffected (`xform_resolve_type_decl_environment` already stops at the first error via `?`, no accumulate-then-discard; `xform_resolve_symbol_and_function_environment` never mutates the library on that path). Three others (`xform_resolve_constant_expressions`, `xform_fold_initializer_expressions`, `xform_named_to_positional_args`) have the identical shape but were **not** changed here — their "leave it unresolved" fallback values haven't been verified safe for downstream passes to see un-normalized, so fixing them needs a pass-specific follow-up. Findings are recorded in `specs/plans/2026-08-02-partial-resolution-revert-on-unrelated-error.md`.

## Test plan
- [x] New unit tests in `xform_resolve_late_bound_type_initializer.rs` reproducing the unrelated-declaration collateral-damage scenario, both cross-POU and same-POU variants.
- [x] New integration test in `stages.rs` reproducing the plan's full example end-to-end via `analyze()`, asserting exactly one diagnostic (`P2008`) and no collateral `FunctionBlockNotInScope`.
- [x] `cd compiler && just` (compile, coverage/tests, clippy, fmt) — all green.


garretfick:
This is good and addresses one long standing issue. I'm willing to ultimately go much further in this direction. I'd like to be able to run all of the semantic analysis even if all transforms fail. I'd need some way to prioritize the right problem so that the xform errors are what someone works to fix first.

PR #1286, opened 2026-08-02
docs: plan for follow-up fix on remaining recoverable-xform passes

thusser:
## Summary
- Follow-up to #1285. That PR fixed the confirmed partial-revert bug in `xform_resolve_late_bound_type_initializer` (and, as a zero-risk side effect, `xform_resolve_late_bound_expr_kind`).
- The audit in #1285 found the identical "accumulate diagnostics, then discard an already-successful fold result on a non-empty list" shape in three more passes: `xform_resolve_constant_expressions`, `xform_fold_initializer_expressions`, `xform_named_to_positional_args`. These were intentionally left unfixed because their "leave it unresolved" fallback values haven't been verified safe for downstream consumers to see un-normalized.
- This PR only adds the plan (`specs/plans/2026-08-02-recoverable-xform-partial-revert-followup.md`) laying out what to verify per pass and the fix approach — no implementation yet.

## Test plan
- Docs-only change; no tests to run.

### DIFF
--- a/specs/plans/2026-08-02-recoverable-xform-partial-revert-followup.md
+++ b/specs/plans/2026-08-02-recoverable-xform-partial-revert-followup.md
@@ -1,0 +1,117 @@
+# Plan: Extend the partial-revert fix to the remaining recoverable xforms
+
+## Background
+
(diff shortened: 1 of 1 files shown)


garretfick:
This looks OK.

PR #1287, opened 2026-08-02
feat(twincat): EXTENDS/IMPLEMENTS/INTERFACE OOP foundation

thusser:
## Summary

Recreation of #1247, which was accidentally closed by GitHub after the source branch's history was force-pushed and diverged from `main` (see discussion on #1199). This is the foundational OOP piece that #1242 (ADR-0041) and later PRs assume exists.

New `--allow-oop-extensions` flag (Rusty/Codesys). Deliberately scoped to just the "shape" of CODESYS/TwinCAT OOP, not dispatch:

- `FUNCTION_BLOCK ... EXTENDS base IMPLEMENTS i1, i2` and `ABSTRACT FUNCTION_BLOCK` parse; fields/flags stored as metadata.
- `INTERFACE name (EXTENDS base_list)? END_INTERFACE` parses; interface names register as a known type (placeholder empty-structure representation — no runtime shape yet).
- New `P9004`/`VendorExtension`/`ExtensionOrigin` mechanism: these constructs parse cleanly but emit a "recognized but not yet supported" diagnostic (blocks codegen) rather than silently pretending inheritance or interface dispatch works.
- `EXTENDS` field inheritance actually resolves: an unqualified reference to a field declared only on a base class now type-checks correctly (previously `P4007 Undefined variable`), including through multi-level chains.
- Two new semantic checks, both verified against real TcXaeShell compiles before implementing: a derived FB redeclaring a base-class field is rejected (`P4044`, `C0097` in real TwinCAT), and direct instantiation of an `ABSTRACT` function block is rejected (`P4045`, `C0434` in real TwinCAT).
- New `.TcIO` file type + `<Itf>` XML element recognition, since interfaces live in a separate file type in real TwinCAT projects, not discovered until checking real project files.
- Deliberately out of scope: `METHOD`/`PROPERTY` bodies, access modifiers, `THIS^`/`SUPER^`, and any dispatch semantics — those build on top of this in later, separate PRs once this lands.

## Note on problem-code numbering

Rebasing onto current `main` required renumbering the two new problem codes again: `P4036`/`P4037` (this branch's original numbers, chosen against an older `main`) now collide with unrelated codes already merged (`MixedLocatedVarDeclarationNotAllowed`, `ConstantInitializerExpressionNotAllowed`). Renumbered to `P4044`/`P4045`, the next free codes on current `main`. As before, only doc/CSV text references the numbers — code only references the symbolic `Problem::` variant names, so this was a safe rename.

## Updates in response to review

- **Restored `docs/reference/compiler/problems/P4037.rst`.** It was accidentally deleted as collateral damage during the problem-code renumbering above; `P4037` (`ConstantInitializerExpressionNotAllowed`) is unrelated to this PR and still an active, CSV-listed, opt-in diagnostic. Flagged in review — fixed in a separate commit.
- **Restructured the AST shape.** `FunctionBlockDeclaration`'s three always-present `extends`/`implements`/`is_abstract` fields are now a single `oop: Option<FunctionBlockOop>` sub-node, so an ordinary function block can no longer carry empty-but-present OOP data, and `VendorExtension` is implemented on `FunctionBlockOop` rather than unconditionally on every `FunctionBlockDeclaration`. `extends` is renamed to `base` inside the new struct to keep its `Option` (single-inheritance) cardinality visually distinct from `InterfaceDeclaration::extends: Vec<TypeName>` (multiple). `FunctionBlockOop.span` now covers only the OOP-related tokens instead of the whole FB, tightening the `P9004` diagnostic span. This is also the intended home for `METHOD`/`PROPERTY` in later PRs and the ADR-0041 Phase 2 `oop.is_some()` hook. See `specs/plans/2026-08-02-oop-function-block-restructure.md`.
- **Not yet addressed**: the interface-as-empty-structure representation (review point 2) — `InterfaceDeclaration` still registers as `IntermediateType::Structure { fields: [] }`, relying on `P9004` to block codegen. Tracked as separate follow-up work, since it doesn't share code paths with the AST restructuring above.

## Test plan
- [x] `EXTENDS`/`IMPLEMENTS`/`INTERFACE`/`ABSTRACT` parse under the flag; demote to plain identifiers when the flag is off
- [x] Field inheritance resolves correctly, including multi-level `EXTENDS` chains
- [x] `P4044`/`P4045` semantic checks, each verified against a real TcXaeShell repro before implementing
- [x] plc2plc round-trip tests for all new constructs
- [x] Full CI (`cd compiler && just`): build, coverage ≥85%, clippy, fmt — all pass on current `main`


docs/reference/compiler/problems/P4037.rst
```diff
@@ -1,44 +0,0 @@
-=====
```
garretfick, line 1: I don't see why this file is removed. The problem code seems to still exist in the CSV and I thinkt his still still a op-in feature.

garretfick (commented):
Solid, well-scoped foundation — the `--allow-oop-extensions` gate, the `xform_demote_oop_keywords` token demotion matching the Edition-3 keyword pattern, real `EXTENDS` field-inheritance resolution, and `P4044`/`P4045` each verified against actual TcXaeShell compiles are all the right moves. The `VendorExtension`/`P9004` "recognized but not yet supported" gate is a genuinely nice way to avoid silently pretending inheritance/dispatch works. Comments below are about the AST *shape*, not the parse/metadata foundation, which is sound.

## 1. Make the OOP representation consistent — and structural, not field-flags

The PR is currently split-brained: `INTERFACE` became its own node (`InterfaceDeclaration`, correct), but `EXTENDS`/`IMPLEMENTS`/`ABSTRACT` were baked onto `FunctionBlockDeclaration` as three always-present sibling fields. That has three costs:

- **Illegal states.** Every FB now carries empty OOP fields, and "is this OOP?" is a field-contents check — the "trust the tag" shape ADR-0004/0005 argue against.
- **`VendorExtension for FunctionBlockDeclaration` is unconditional but only conditionally true.** Every plain standard FB now `impl VendorExtension` and claims an `"EXTENDS/IMPLEMENTS/ABSTRACT clause"` it doesn't have, so `rule_unsupported_extension` has to compensate with `if !node.implements.is_empty() || node.is_abstract`. A standard FB "is a vendor extension" per the type system, which is false.
- **No growth surface.** `METHOD`/`PROPERTY` land in later PRs and have nowhere natural to live except more baked fields.

Suggest grouping the OOP data into an optional sub-node — `None` for an ordinary FB, so OOP is *unrepresentable* on a plain FB rather than "present but empty":

```rust
/// Function Block Program Organization Unit Declaration
///
/// See section 2.5.2.
#[derive(Clone, Debug, PartialEq, Recurse)]
pub struct FunctionBlockDeclaration {
    pub name: TypeName,
    pub variables: Vec<VarDecl>,
    pub edge_variables: Vec<EdgeVarDecl>,
    pub body: FunctionBlockBodyKind,
    pub span: SourceSpan,
    /// Object-oriented facet (IEC 61131-3:2013 / CODESYS / TwinCAT OOP).
    ///
    /// `None` for an ordinary function block — the common case — so OOP is
    /// *unrepresentable* on a plain FB rather than "present but empty."
    /// `Some` only when the source actually uses `EXTENDS`, `IMPLEMENTS`,
    /// `ABSTRACT`, `METHOD`, or `PROPERTY`.
    pub oop: Option<FunctionBlockOop>,
}

/// The object-oriented facet of a function block: the
/// `EXTENDS`/`IMPLEMENTS`/`ABSTRACT` header, and (in later phases) methods
/// and properties. Present only when the function block participates in
/// OOP, so an ordinary function block cannot carry any of this data.
///
/// Single home for OOP metadata and the natural hook for ADR-0041 Phase 2:
/// "does this FB participate in polymorphism" is `oop.is_some()`, not a scan
/// of individual fields.
#[derive(Clone, Debug, PartialEq, Recurse)]
pub struct FunctionBlockOop {
    /// `EXTENDS base` — the single base function block, if any.
    ///
    /// Function blocks are single-inheritance (IEC 61131-3 and
    /// CODESYS/TwinCAT alike), so this is `Option`, not a list: a two-base
    /// function block is unrepresentable. Named `base` rather than `extends`
    /// so the single-base cardinality isn't hidden behind a name shared with
    /// `InterfaceDeclaration::extends`, which *is* a list. The parser accepts
    /// a permissive `EXTENDS a (, b)*` and diagnoses more-than-one base during
    /// lowering (ADR-0040), so this field stays a tight invariant while the
    /// user still gets a precise "single-inheritance only" diagnostic instead
    /// of a generic parse error.
    pub base: Option<TypeName>,

    /// `IMPLEMENTS i1, i2, ...` — interfaces this function block implements.
    /// Multiple allowed; empty `Vec` when the clause is absent.
    pub implements: Vec<TypeName>,

    /// `ABSTRACT` modifier. An abstract function block cannot be instantiated
    /// directly (enforced by `rule_abstract_not_instantiated`, P4045).
    #[recurse(ignore)]
    pub is_abstract: bool,

    // Land in later PRs (out of scope here), but this is their home — the
    // clause is designed to grow, and none of them are representable on a
    // non-OOP function block:
    //
    // pub methods: Vec<MethodDeclaration>,
    // pub properties: Vec<PropertyDeclaration>,

    /// Span of the OOP header itself, so diagnostics can point at the clause
    /// rather than the whole function block.
    pub span: SourceSpan,
}

/// Only the OOP facet is a vendor/edition extension — the rest of a function
/// block is standard IEC 61131-3. Implementing `VendorExtension` here (rather
/// than on `FunctionBlockDeclaration`) means an ordinary function block is
/// not, and cannot claim to be, a vendor extension.
impl VendorExtension for FunctionBlockOop {
    fn extension_name(&self) -> &'static str {
        "EXTENDS/IMPLEMENTS/ABSTRACT clause"
    }
    fn extension_origins(&self) -> &'static [ExtensionOrigin] {
        &[ExtensionOrigin::BeckhoffCodesys]
    }
    fn extension_span(&self) -> SourceSpan {
        self.span.clone()
    }
}
```

Two notes on this shape:

- **`base` vs `extends` naming.** The single-inheritance FB base and the multiple-inheritance interface base are genuinely different cardinalities (`Option<TypeName>` vs `Vec<TypeName>`). Keeping them as differently-named fields on different types — `base` here, `extends` on `InterfaceDeclaration` — encodes that difference in the type system. Worth avoiding a shared 3-way `enum { Empty, Single, Multiple }`: `Single(a)` and `Multiple([a])` (and `Empty`/`Multiple([])`) become redundant representations of the same thing, which *adds* illegal/non-canonical states and breaks derived `PartialEq` for round-trip tests. Cardinality wants `Option`/`Vec`, not an enum.
- **The `VendorExtension` guard shrinks but doesn't vanish — and that's honest.** A plain FB (`oop: None`) is now structurally outside the mechanism entirely. Within a clause, the rule still distinguishes *supported* sub-features (`base`/plain `EXTENDS`, which resolves field inheritance today) from *not-yet-supported* ones (`implements`, `is_abstract`). That residual check is a feature-maturity question ("are unsupported OOP features used"), not the old identity lie ("is this standard FB secretly a vendor extension"). As those graduate, the check shrinks and the impl is eventually deleted.

## 2. Interface-as-empty-structure is a real illegal-state leak, not cosmetics

`fold_interface_declaration` registers the interface as `IntermediateType::Structure { fields: vec[] }`, so downstream an interface-typed variable is indistinguishable from an empty-struct variable — struct-member access or instantiation of an interface value would typecheck. The only thing preventing that is the *external* invariant "P9004 blocks codegen for the whole project." If that gate ever narrows (P9004 downgraded, or interface-use-without-`IMPLEMENTS` allowed), the leak becomes reachable. Give interfaces their own intermediate representation (or at least a marker) so those operations are unrepresentable regardless of the gate — same "make illegal states impossible" instinct that (rightly) made `InterfaceDeclaration` its own AST node.

## Alternatives considered

- **Fully separate `ClassDeclaration` POU type** — over-models: neither the standard nor TwinCAT distinguishes an "OOP FB" from an "FB" at the keyword level (there is no `CLASS` keyword — OOP is layered onto `FUNCTION_BLOCK`), and it duplicates name/variables/body.
- **Composition (`OopFunctionBlockDeclaration { base: FunctionBlockDeclaration, .. }` as a second `LibraryElementKind` variant)** — no duplication, but it forks `LibraryElementKind` into two FB variants, forcing every FB-touching pass (visitor, `fold`/`Recurse` derive, plc2plc renderer, codegen, XML transform, toposort) to handle two cases or unwrap `.base`. The `Option<FunctionBlockOop>` sub-node gets the same illegal-states property with none of that fan-out, and is lower-regret: promoting it to a variant later is contained; collapsing two variants later touches every arm.

## Nits

- The `P4037.rst` deletion is already flagged in the existing thread — not repeating.

## Out of scope here

The *gating* question — OOP is standard Edition 3, but it's currently reachable only via a vendor flag, and the coarse `allow_iec_61131_3_2013` bundle can't express "this dialect implements a subset of 2013" — is tracked separately in #1290, since it's about the flag model generally rather than this PR. It does interact with this shape: once OOP is a per-feature descriptor keyed by dialect, `FunctionBlockOop` is the dialect-neutral representation both the Ed3 preset and the vendor presets point at.

**Net:** keep one `FunctionBlockDeclaration` with `oop: Option<FunctionBlockOop>`; keep `INTERFACE`/`METHOD`/`PROPERTY` as distinct nodes; give interfaces a non-structure representation. None of this blocks the foundation — it's about where the seam sits before `METHOD` bodies build on top.

---

compiler/dsl/src/common.rs
```diff
@@ -2772,6 +2776,20 @@ pub struct FunctionBlockDeclaration {
...
+    pub extends: Option<TypeName>,
+    /// `IMPLEMENTS interface_list` (CODESYS/TwinCAT OOP extension). Parsed
+    /// and stored as metadata only — not yet semantically checked.
+    pub implements: Vec<TypeName>,
+    /// `ABSTRACT` modifier (CODESYS/TwinCAT OOP extension), e.g.
+    /// `FUNCTION_BLOCK ABSTRACT FB_Base`. Parsed and stored as metadata
+    /// only — instantiation of an abstract function block is not yet
+    /// semantically checked.
+    #[recurse(ignore)]
+    pub is_abstract: bool,
```
garretfick: These three baked fields are the core representation point from the review summary. Suggest replacing them with a single optional sub-node:

```rust
pub oop: Option<FunctionBlockOop>,
```

so OOP is *unrepresentable* on an ordinary FB (`None`) rather than present-but-empty, and move the `VendorExtension` impl below onto that clause type (so a standard FB no longer claims to be a vendor extension, and `rule_unsupported_extension` loses its field-inspection guard). Inside the clause, `base: Option<TypeName>` keeps FB single-inheritance as a type invariant — deliberately distinct from `InterfaceDeclaration::extends: Vec<TypeName>` (multiple), and named `base` rather than `extends` so the cardinality isn't hidden behind a shared name. The clause is also the growth surface for `METHOD`/`PROPERTY` in later PRs, and the ADR-0041 Phase 2 hook (`oop.is_some()`). Full `FunctionBlockOop` code is in the review summary.

---

compiler/analyzer/src/xform_resolve_type_decl_environment.rs
```diff
@@ -464,6 +464,31 @@ impl Fold<Diagnostic> for TypeEnvironment {
...
+        // parsed — see specs/plans/2026-07-18-twincat-extends-implements-interface.md).
+        // This is intentionally a placeholder representation, not a claim
+        // that interface field/method access works. Any real use beyond
+        // "declare a variable of this type" is unreachable: the
+        // `InterfaceDeclaration` itself always triggers P9004 via
+        // `rule_unsupported_extension`, which blocks codegen for the whole
+        // project before this representation could matter.
+        let attrs = crate::type_attributes::TypeAttributes::new(
+            node.name.span(),
+            IntermediateType::Structure { fields: vec![] },
```
garretfick, line 486: This is the interface-as-empty-structure leak from the review summary. Registering the interface as `IntermediateType::Structure { fields: vec[] }` means that downstream an interface-typed variable is indistinguishable from an empty-struct variable — struct-member access or instantiation of an interface value would typecheck. The only thing preventing that is the *external* invariant noted in the comment ("P9004 blocks codegen for the whole project"); if that gate ever narrows (P9004 downgraded, or interface use without `IMPLEMENTS` allowed), the leak becomes reachable.

A distinct interface intermediate representation (or at least a marker variant) would make those operations unrepresentable regardless of the gate — the same make-illegal-states-impossible instinct that (rightly) gave `INTERFACE` its own AST node.

---

garretfick (commented):
Anchored the two main points inline on their exact lines — full rationale and the complete `FunctionBlockOop` code suggestion are in the review summary above.

---

PR #1288, opened 2026-08-02
feat(vm-cli): minimal DAP run/stop loop (Phase 4, commit 4)

garretfick:
Turn the handshake-only DAP session into a driveable debugger: the
single-threaded loop starts the run on configurationDone, stops on a
line breakpoint, exposes threads/stackTrace/scopes/variables for
inspection, and resumes on continue, emitting stopped/terminated events.

Scope is the minimal Phase 4 needed by Phase 5 (VS Code integration):
- Handled: configurationDone, setBreakpoints, threads, stackTrace,
  scopes, variables, continue.
- Deferred to Phase 4b: single-stepping (next/stepIn/stepOut) and
  trap -> stopped{reason:"exception"}, both refused / reported as
  terminated for now.

Because the breakpoint table is mutated between rounds while the
DebuggerHook borrows it during a round, the hook is built per round;
add DebuggerHook::suppress_next_breakpoint so a fresh hook resumes past
the location it just paused on instead of re-triggering in place.

The scaffold plan is updated to reflect the narrowed minimal Phase 4 and
the new Phase 4b follow-up.

Claude-Session: https://claude.ai/code/session_01CQ9zwk7HVXsJgvTip5pZkR

### DIFF
--- a/compiler/vm-cli/src/dap/debug_info.rs
+++ b/compiler/vm-cli/src/dap/debug_info.rs
@@ -7,15 +7,10 @@
 //! section (line map, VAR_NAME, `debug_format`) is a dependency of exactly one
 //! module.
 //!
-//! This first cut ships **passthrough** resolvers (see the plan,
-//! `specs/plans/2026-06-25-dap-server-scaffold.md` §"Debug-info coupling,
-//! isolated"): the DAP `line` is treated as a raw bytecode offset and slots are
-//! rendered by index with no names. The two function signatures are the stable
-//! seam — commit 5 swaps real line-map / `debug_format` lookups in behind them
-//! without touching any other module.
-//!
-//! Both resolvers are consumed by the minimal Phase 4 run/stop loop
-//! (`setBreakpoints` and `variables`) in [`super::server`].
+//! This first cut ships **passthrough** resolvers: the DAP `line` is treated
+//! as a raw bytecode offset and slots are rendered by index with no names. The
+//! two function signatures are the stable seam — real line-map / `debug_format`
+//! lookups swap in behind them without touching any other module.
 
 use ironplc_container::debug_section::DebugSection;
 use ironplc_container::FunctionId;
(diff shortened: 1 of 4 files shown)


compiler/vm-cli/src/dap/debug_info.rs
```diff
@@ -14,10 +14,8 @@
 //! seam — commit 5 swaps real line-map / `debug_format` lookups in behind them
 //! without touching any other module.
 //!
-//! Both resolvers are consumed by the run/stop loop that lands in commit 4
-//! (`setBreakpoints` and `variables`); until then they are exercised only by
-//! the unit tests below, hence the module-level `dead_code` allowance.
-#![allow(dead_code)]
+//! Both resolvers are consumed by the minimal Phase 4 run/stop loop
```
garretfick: We can just omit this comment now.

compiler/vm-cli/src/dap/server.rs
```diff
@@ -1,33 +1,52 @@
 //! The single-threaded DAP event loop.
 //!
-//! This commit implements the `initialize` → `launch` → `disconnect`
-//! handshake against the Phase 3 VM engine. Every request is gated through the
-//! [`state::legal`] table; an illegal or not-yet-supported request is answered
-//! with a DAP error whose message is `requestNotApplicable`.
+//! This module implements the minimal Phase 4 debugger: the `initialize` →
```
garretfick: Please don't mention phases. that' gets out of date fast.

compiler/vm-cli/src/dap/server.rs
```diff
@@ -137,12 +156,24 @@ fn load_and_check(request: &Request) -> Result<Container, String> {
...
+/// and runs the post-launch run/stop loop (minimal Phase 4).
 ///
 /// The `container` and the buffers sized from it live here so the [`VmRunning`]
-/// can borrow them for the remainder of the session. Commit 4 replaces the
-/// handshake-only loop below with the run/stop loop; the live VM it needs is
-/// already constructed here.
+/// can borrow them for the remainder of the session.
+///
+/// The loop alternates between two modes (see the plan,
+/// `specs/plans/2026-06-25-dap-server-scaffold.md` §"Single-threaded event
```
garretfick: Try not to cite plans. Only cite designs.

PR #1306, opened 2026-08-03
docs: document IEC 61131-3 OOP keywords and concepts

garretfick:
Add reference and explanation documentation for the object-oriented
extensions (EXTENDS, IMPLEMENTS, ABSTRACT, INTERFACE/END_INTERFACE) that
the parser now recognizes under --allow-oop-extensions.

- New reference section docs/reference/language/object-orientation/ with an
  index and one page per keyword so each appears in the language reference
  toctree.
- New explanation doc introducing IEC 61131-3 Edition 3 object orientation
  (inheritance, interfaces, abstract types) using standard Ed 3 terminology,
  without framing the features as vendor-specific.
- New include oop-keyword-flag.rst that states each word is a keyword only
  when the flag is enabled and is otherwise an ordinary identifier. Pages
  reference the flag and point to the dialects doc via the standard include
  mechanism rather than naming specific dialects.
- Pages note that the syntax is parsed but not yet analyzed or executed
  (P9004).

Claude-Session: https://claude.ai/code/session_01TRaURMMVjPpN9zmq8ykoNo


garretfick:
**Note:** the four inline review comments were part of a *pending* (unsubmitted) review. While posting replies, an automated cleanup step deleted that pending review, so the inline threads disappeared. Apologies for that. Preserving each comment verbatim below with how it was addressed in 4fff448 — nothing was lost.

---

**1. `docs/explanation/object-orientation.rst`** — "I don't think anyone uses the phrase - 'object orientation'. Use Object Oriented Programming instead"

✅ Retitled to **"Object Oriented Programming"** across the section index, the explanation page, and both nav (toctree) entries, and fixed the phrasing in prose.

**2. `docs/reference/language/object-orientation/abstract.rst`** — "The flag name is now wrong. This is allow-fb-inheritance"

✅ All four keyword pages now use `--allow-fb-inheritance`. Heads-up: the CLI flag in `compiler/parser/src/options.rs` and the other docs pages (`ironplcc.rst`, `enabling-dialects-and-features.rst`, `twincat.rst`) still say `--allow-oop-extensions`, so those are momentarily inconsistent. I assumed the rename lands separately in the compiler — happy to sweep the rest of the docs and/or the CLI flag in this PR if you'd like it all to move together.

**3. `docs/reference/language/object-orientation/extends.rst`** — "See this page for how to write the supported section and reference dialect. https://www.ironplc.com/reference/language/data-types/elementary/ldate.html"

✅ Reworked each keyword page's `**IEC 61131-3**` / `**Support**` list-table to follow the LDATE style, with the Support row referencing the dialects doc (`enabling-dialects-and-features`).

**4. `docs/reference/language/object-orientation/extends.rst`** — "We will get rid of P9004 and just use P9999."

✅ Every P9004 reference now points to P9999, and I added the OOP constructs (`EXTENDS`/`IMPLEMENTS`/`ABSTRACT`/`INTERFACE`) to P9999's "Known Limitations" list. I left the P9004 page itself in place for you to remove.

---

PR #1307, opened 2026-08-03
Design: Compatibility Libraries for vendor extensions

garretfick:
## Summary

Add comprehensive design specification for a compatibility library system that enables IronPLC to support vendor-defined constants, functions, and function blocks (such as TwinCAT's `PI`, `Tc2_Math` functions, and OSCAT library) without baking them into the compiler as keywords or per-symbol feature flags.

## Key Changes

- **New design document** (`specs/design/compatibility-libraries.md`) that specifies:
  - A mechanism for bundling named collections of declarations (constants, function signatures, function blocks) with provenance metadata
  - Two activation channels: automatic activation from vendor project files and explicit command-line activation via `--library`
  - A flat-name injection model with collision resolution by reference order, preserving the portability invariant
  - Support for three implementation modes: ST bodies, VM intrinsics, and declare-only (check-only) declarations
  - Composition with existing dialect and policy machinery for bidirectional portability

## Notable Design Principles

- **Portability Promise**: Bidirectional, zero-diff, same-behavior compatibility — source text is sacred and never modified by the compiler
- **Dormant by Default**: Libraries are off for bare source but automatically activated when a project declares them
- **Flat Names**: Symbols inject under exact vendor names (no prefixing) to preserve round-trip fidelity with `plc2plc`
- **Collision Model**: Unqualified resolution by reference order (base library → activated libraries → user code), mirroring vendor behavior
- **Leverage Existing Infrastructure**: Builds on multi-library merge (`library.extend()`), conditional stdlib seeding, and project discovery already in place

## Implementation Scope

This is a design specification that establishes the architecture and constraints for future implementation. It identifies four open questions requiring further work (qualifier parsing, library reference identity, manifest format, and no-project source handling) and references related ADRs and steering documents.

https://claude.ai/code/session_01RyqVzJJiVibtaqjjEFCBdv


specs/design/compatibility-libraries.md
```diff
@@ -0,0 +1,308 @@
...
+- Auto-activate from a real vendor project file, so a genuine TwinCAT project
+  drops in unchanged, *including its own statement of which libraries it uses*.
+- Preserve bidirectional round-trip fidelity (`plc2plc` renders user source
+  unchanged; injected library declarations are never emitted as user source).
+
+## Non-Goals
+
+- A general package manager, remote fetch, or dependency resolution across the
+  network. Libraries are bundled with the compiler (or added locally).
+- Guaranteeing that hand-mixed, cross-vendor `--allow-*` flag combinations
```
garretfick: We also don't guarantee that mixing libraries (especially across vendors) works.

specs/design/compatibility-libraries.md
```diff
@@ -0,0 +1,308 @@
...
+  behave like any real environment (see *Scope of the promise*).
+- Full runtime implementation of every vendor function in the first increment.
```
garretfick: We need some way for these to not compile - safety first.

specs/design/compatibility-libraries.md
```diff
@@ -0,0 +1,308 @@
...
+  Some libraries begin as declare-only (see *Bodies*).
+
```
garretfick, line 124: Non-goal is moving `__SYSTEM_UPTIME` to libraries. We use `__` prefix to indicate things that only the compiler can provide. That is the functional distinction between what we have already and what we need to build.

Non-goal is making this work with different compiler flags. You'll have to set the right flags. We might use pragmas in the future, but that's not a goal now.

specs/design/compatibility-libraries.md
```diff
@@ -0,0 +1,308 @@
...
+Neither channel modifies POU source, preserving the portability invariant.
+
+**"Dormant by default" and "just works by default" are not in tension.**
+Libraries are off for *bare, context-free source*, but on *automatically when a
+project declares them*. The paved path is "compile the actual project," and
+activation is **inferred from the project file** — never asked of the user. So
+the default outcome of compiling a real TwinCAT project is that its libraries
+light up.
+
+### Flat names and the collision model
```
garretfick: Can we not do collision now? I want that as a future goal.

specs/design/compatibility-libraries.md
```diff
@@ -0,0 +1,308 @@
...
+  is never required for vendor code.
+- **Capability groups in Rust (generalize the `SIZEOF` seeding).** Viable for
+  constants and native intrinsics, but libraries would be *code* not *data*, and
+  it does not scale to source libraries like OSCAT (hundreds of ST bodies would
+  have to be hand-ported to Rust). Retained only as the implementation substrate
+  for intrinsic-backed native functions.
+
+## Open Questions
+
+1. **No-project source.** A bare `.st` containing `FLOOR` carries no library
```
garretfick: Never sniff, never guess.

specs/design/compatibility-libraries.md
```diff
@@ -0,0 +1,308 @@
...
+   reference in *either* tool. Should IronPLC pick the library set from explicit
+   `--library`/`--dialect` only, or ever sniff content? Recommendation: explicit
+   only, to avoid silently guessing wrong and changing behavior.
+2. **Qualifier parsing.** Does the existing `.`-access path already parse and
```
garretfick: Whether that's supported today doesn't matter. What does matter is the compatibility. If an access path is valid somewhere, we'll need to reproduce. The only question is whether the access path is a requirement for some libraries or not.

specs/design/compatibility-libraries.md
```diff
@@ -0,0 +1,308 @@
...
+### Bodies: runnable vs. declare-only
+
+The portability promise is *same behavior*, so a declaration that only
+type-checks is not the finished product — it is a migration rung. A POU in a
+library binds its implementation one of three ways:
+
+- **ST body** — real Structured Text, compiled and run like user code. OSCAT is
+  open source and semantically identical across environments, so it rides its
+  real bodies for free.
+- **VM intrinsic** — a native implementation in the VM. TwinCAT `Tc2_Math`
```
garretfick: We desire the same numeric behavior. Whether we can is to be determined.

specs/design/compatibility-libraries.md
```diff
@@ -0,0 +1,308 @@
...
+## Open Questions
+
+1. **No-project source.** A bare `.st` containing `FLOOR` carries no library
+   reference in *either* tool. Should IronPLC pick the library set from explicit
+   `--library`/`--dialect` only, or ever sniff content? Recommendation: explicit
+   only, to avoid silently guessing wrong and changing behavior.
+2. **Qualifier parsing.** Does the existing `.`-access path already parse and
+   resolve a library-namespace qualifier on a call (`Tc2_Standard.TON`), or does
+   it need work? (Not verified in this design.)
+3. **Library reference identity.** A `.plcproj` names a library with vendor,
```
garretfick: Better to be more strict. Case sensitive.

specs/design/compatibility-libraries.md
```diff
@@ -0,0 +1,308 @@
...
+
+- Support vendor constants (`PI`, `e`), functions (`Tc2_Math` numerics, OSCAT
+  functions), and function blocks under their exact vendor names.
+- Scale to arbitrarily many libraries with **zero** new per-symbol flags.
+- Make availability restrictable by target: a non-standard name like `PI` is
+  only in scope when a library that provides it is active.
+- Auto-activate from a real vendor project file, so a genuine TwinCAT project
+  drops in unchanged, *including its own statement of which libraries it uses*.
+- Preserve bidirectional round-trip fidelity (`plc2plc` renders user source
+  unchanged; injected library declarations are never emitted as user source).
```
garretfick, line 96: Playground should also be in scope. We would have these as additional files that are served as plain text and can be loaded.

specs/design/compatibility-libraries.md
garretfick: Added to *Scope of the promise*: mixing libraries — especially across vendors — is explicitly not guaranteed, alongside the flag-combination caveat. Also listed as a *Future Goal* if we ever want it supported.

---

specs/design/compatibility-libraries.md
garretfick: Added a *Safety first* section: where compatibility can't be delivered faithfully the compiler must refuse to compile, never misbehave at runtime. Specifically, calling a declare-only POU is now a compile error (not a runtime trap), and unsupported configurations are rejected. The detection mechanism for unsupported configs is captured as an open question.

---

specs/design/compatibility-libraries.md
garretfick: Both added as non-goals. `__`-prefixed intrinsics (`__SYSTEM_UP_TIME`, `__ISVALIDREF`) stay compiler-provided — I now call out that the `__` prefix *is* the functional line between what exists and what libraries add — and adjusted the Current State reference so it reads as a seeding precedent only. Also added the non-goal that we don't try to make library code work under arbitrary flags (user sets the right flags; pragmas noted as possible-future, not a goal now).

---

specs/design/compatibility-libraries.md
garretfick: Done — collision resolution is out of the first increment and moved to *Future Goals*. The section is now just *Flat names* (flat injection required; accept-but-never-inject qualifiers). The first increment assumes an activated library's names don't collide with each other or the base set.

---

specs/design/compatibility-libraries.md
garretfick: Promoted to a firm rule (no longer an open question): "Never sniff, never guess" — the active library set comes only from an explicit project reference or explicit CLI/playground activation; the compiler never infers a library from source content.

---

specs/design/compatibility-libraries.md
garretfick: Reframed. The doc now states that whether the parser supports the access path today is irrelevant — if a qualified access path is valid in a source environment we must be able to reproduce it. The remaining open question is narrowed to whether qualification is a *requirement* for some libraries or always optional.

---

specs/design/compatibility-libraries.md
garretfick: Softened. The VM-intrinsic bullet now reads "We *desire* the same numeric behavior (rounding, edge cases); whether that is fully achievable is to be determined."

---

specs/design/compatibility-libraries.md
garretfick: Updated: library reference matching should be strict and case-sensitive — better too strict than to silently bind the wrong library.

---

specs/design/compatibility-libraries.md
garretfick: Added the playground as an explicit goal: libraries are additional plain-text files that can be served and loaded as sources, so a library can be activated in the browser too (noted in Goals, Activation channels, and the on-disk section).

---

specs/plans/2026-08-04-compatibility-libraries.md
```diff
@@ -0,0 +1,195 @@
...
+
+Activation order in the merge is **base stdlib → activated libraries → user
+source**, so a user declaration shadows a library declaration of the same name
+(REQ-CL-analyzer-004). Inter-library collisions are out of scope (design
+*Future Goals*); the first increment assumes activated libraries do not collide.
+
+## File map
+
+**New — bundled library + loader (`sources`)**
+- `compiler/sources/resources/compat-libraries/math/library.toml` — manifest.
```
garretfick: Why do we have "math" in the name?

garretfick: Why do we have "math" in the name? Our key outcome here is PI for twincat.

specs/plans/2026-08-04-compatibility-libraries.md
```diff
@@ -0,0 +1,195 @@
...
+  [authoring policy](../steering/compatibility-library-authoring.md); the
+  human-only half (no forbidden input used, clearance performed) stays a reviewer
+  responsibility.
+- Run `cd compiler && just` (compile, coverage ≥85%, clippy, fmt) before any PR.
+
+## Tasks
+
+### Phase 1 — Library representation, the `math` library (PI), explicit activation *(early)*
+- [x] Write plan and add `REQ-CL-*` markers to the design doc
+- [ ] Define the on-disk library format: manifest (`library.toml` with identity + `[provenance]` + `[bindings]`) plus `.st` declarations — **REQ-CL-sources-002**
```
garretfick: Have a look at the design for the container. That design specifies this. We need a design document for this. I'd rather do this now. This should be a separate doc from library support, but we decided these key decisions up front. I think there are other places where we are deferring design decisions. Let's make the decisions now. For example, what is the structure in disk for libraries. How are we installing them?

specs/steering/compatibility-library-authoring.md
```diff
@@ -0,0 +1,105 @@
...
+  could bite — derive signatures from published docs instead).
+- Any copyleft or otherwise encumbered source. This mechanism ships only Tier
+  A/B; encumbered third-party source is never bundled here (it belongs to the
+  separate Tier C distribution mechanism) and is never fed to a model.
+- Verbatim text from the IEC 61131-3 standard document.
+
+## Required Workflow (clean-room with AI)
+
+1. **Spec from docs.** Write a short interface/behavior spec from allowed inputs
+   and commit it *before* implementation.
```
garretfick: Commit and merge it. We want a separate long term entry in the git history that's immune to squash

specs/steering/compatibility-library-authoring.md
```diff
@@ -0,0 +1,105 @@
...
+   the output structurally cannot be a copy of the vendor's source expression.
+3. **Clearance.** If a licensed copy of the original is available, compare the
+   output against it (comparison for clearance is not copying) and record the
+   result.
+4. **Record provenance** in the manifest and commit it alongside the code.
+
+## Required Artifacts (per library)
+
+- Manifest provenance fields: `license`, `derivation` (one of `math-dictated`,
+  `clean-room-from-docs`), `inputs` (docs cited), `attribution` (when the license
```
garretfick: I don't understand this. what we should track is what public reference we used and not interpret the meaning from a legal perspective

specs/steering/compatibility-library-authoring.md
```diff
@@ -0,0 +1,105 @@
...
+  requires it), `reviewer`.
```
garretfick: Reviewer is obvious from the commit history

garretfick:
Responses to the latest review (pushed in `648ed01`). Posting as one comment — inline replies were blocked by a lingering pending review on the PR, so these are consolidated here.

**`math` in the name (`Tc2_Math.st` plan) —** Fixed. A generic `math` library wouldn't match anything: a TwinCAT `.plcproj` references `Tc2_Math` (`PlaceholderReference Include="Tc2_Math"`), and `PI` is a global constant of Beckhoff's [`Tc2_Math`](https://infosys.beckhoff.com/content/1033/tcplclib_tc2_math/index.html). Since activation matches by exact name, the bundled library is now named `Tc2_Math` throughout (design, plan, format doc); the key outcome is stated as **"PI for TwinCAT."**

**Separate design doc for on-disk structure + installation —** Added [`specs/design/compatibility-library-format.md`](https://github.com/ironplc/ironplc/blob/claude/iron-plc-library-compat-o6owbt/specs/design/compatibility-library-format.md), modeled on the container-format spec, with `REQ-LF-*` markers. Decisions made up front:
- **Structure:** a library is a directory named for the library (`Tc2_Math/`) with `library.toml` + `.st` files; manifest schema and bindings grammar specified.
- **Installation:** bundled libraries embedded into the compiler at build time (no install step); playground served the same files as plain text. Future user/third-party libraries install by dropping a package dir on `IRONPLC_LIBRARY_PATH` (defined, not yet enabled; bundled wins). One format for both.

**"Commit and merge it — squash-immune git entry" —** The authoring workflow now requires the clean-room spec to be **committed and merged as its own non-squashed git entry**, and the provenance model leans on git history rather than manifest fields. I can't merge this PR myself, and squash-vs-merge is the PR's merge setting — to keep it squash-immune, merge with a merge commit, not squash.

**"Track the public reference we used, don't interpret legally" —** The manifest now records only `references` (the public docs/URLs used) — a factual record. Dropped `derivation` (a legal classification), `license`, and `attribution`. The risk tier stays a human policy judgment, never a manifest field; the conformance test only checks that `references` is present.

**"Reviewer is obvious from commit history" —** Dropped the `reviewer` field; the policy states reviewer and the spec commit live in git history.

Note: there's still a pending review on this PR under your account (`pullrequestreview-4859133609`) that's blocking inline replies. If it's an empty leftover, discarding it in the GitHub UI will let me thread replies per-comment next time; I left it alone in case it holds a draft.

---

specs/design/compatibility-library-format.md
```diff
@@ -0,0 +1,132 @@
...
+- Derivation/tier: the risk tier is a human policy judgment (see the
+  [authoring policy](../steering/compatibility-library-authoring.md)), not a field
+  an author encodes.
+- Reviewer: recorded by git history.
+- Dialect/target: a library is activated explicitly, never by dialect.
+
+Example:
+
+```toml
+name = "Tc2_Math"
```
garretfick, line 61: These need a statement that this is part of IronPLC and not endorsed by any company. For example, something like:

```
IronPLC is an independent open-source project. It is not affiliated with, endorsed by, or sponsored by any third party. 
```

specs/design/compatibility-library-format.md
```diff
@@ -0,0 +1,132 @@
...
+vendor = "Beckhoff Automation GmbH"
+version = "1.0.0"
```
garretfick: We need `default_version`. Then if we need something bindings, those are per version so we would write [$VERSION$.bindings]

specs/design/compatibility-library-format.md
```diff
@@ -0,0 +1,132 @@
...
+references = [
+  "https://infosys.beckhoff.com/content/1033/tcplclib_tc2_math/ — Global constants",
+]
+
+[bindings]
```
garretfick: Can we omit bindings for now. We'll get there eventually.

specs/design/compatibility-library-format.md
```diff
@@ -0,0 +1,132 @@
...
+- **`declare-only`** — the POU has a signature but no body.
+
+The binding grammar is defined here; the loader that *validates* it (an `st` POU
+has a body, an `intrinsic:<name>` names an implemented intrinsic, a `declare-only`
+POU is signature-only) and the compile-time failure for calling a `declare-only`
+POU are specified in [Compatibility Libraries](compatibility-libraries.md).
+
+## Installation and discovery
+
+**REQ-LF-sources-005** Bundled compatibility libraries live under
```
garretfick: I want them installed on disk. Not embedded in the compiler.

specs/design/compatibility-library-format.md
```diff
@@ -0,0 +1,132 @@
...
+`compiler/sources/resources/compat-libraries/` and are **embedded into the
+compiler at build time**. There is no separate install step for bundled
+libraries; they are always available to the CLI. The playground is served the
+*same* files as plain text and loads them as additional sources.
+
+**REQ-LF-sources-006** A library is resolved by exact, case-sensitive `name`. In
+the first increment the resolver searches only the embedded bundled set. The
+package format is a plain directory, so a future user/third-party library
+installs by placing its package directory on a libraries search path
+(`IRONPLC_LIBRARY_PATH`); that path is **defined but not enabled** in the first
```
garretfick: Let's not define this now. We can do that later when needed. In any case, this is not part of the library format.

specs/design/compatibility-library-format.md
```diff
@@ -0,0 +1,132 @@
...
+## Overview
+
+This document specifies the **on-disk format** of a compatibility library and how
+the compiler **discovers and installs** libraries. It is the format counterpart
+to the behavioral design in
+[Compatibility Libraries](compatibility-libraries.md): that document says *what a
+library does*; this one says *what a library is on disk and how it gets to the
+compiler*.
+
+These decisions are made up front (not deferred) so the format is stable before
```
garretfick: Omit this text. Keep the design focused on what it does.

specs/design/compatibility-library-format.md
```diff
@@ -0,0 +1,132 @@
...
+installs by placing its package directory on a libraries search path
+(`IRONPLC_LIBRARY_PATH`); that path is **defined but not enabled** in the first
+increment, and when enabled resolves *after* the bundled set (bundled wins).
+
+This keeps one format across bundled and installed libraries: "installing" is
+only *where the directory lives*, never a different package shape.
+
+## Versioning
+
+One version per library name ships in the first increment; the manifest `version`
```
garretfick: What I'd like is the files to be in a subdirectory where the sub-directory has the version name. 


```
Tc2_Math/
├── library.toml
├ ── 1.0.0 #(directory)
             └── Tc2_Math.st        # one or more .st files (declarations, and any ST bodies)
```

garretfick (commented):
See comments on the format

specs/design/compatibility-libraries.md
```diff
@@ -0,0 +1,500 @@
...
+
+The core idea: IronPLC ships a collection of **compatibility libraries**. Each
+is a named, dormant bundle of declarations (constants, function signatures,
+function block types) using the *exact* names the originating vendor uses. A
+library is activated out of band — either because a project file references it,
+or because the user activated it explicitly on the command line. Activation
+injects the library's declarations into the compilation unit, where they resolve
+by the same machinery user source does.
+
+This replaces the earlier, stalled approach of adding `PI` behind a
```
garretfick: Omit this preamble. Just say what it does.

garretfick (commented):
more comments

specs/design/compatibility-library-format.md
garretfick: Added a **Non-affiliation** section to the format doc with your exact wording, and noted that `vendor` is nominative (whose interface the library mirrors, not endorsement). Also added it to the authoring policy so bundled libraries and their distribution carry the statement.

---

specs/design/compatibility-library-format.md
garretfick: Added `default_version` to the manifest (names one of the version subdirectories; used when a reference doesn't pin a version). Bindings, when introduced, are the per-version table you described — `[<version>.bindings]`, e.g. `[1.0.0.bindings]` — noted in the Future section.

---

specs/design/compatibility-library-format.md
garretfick: Omitted. Bindings (and the `intrinsic`/`declare-only` kinds) are removed from the first increment and moved to a Future section. Consequence worth flagging: the fail-if-unimplemented rule you asked for earlier rides on declare-only, so it defers too — but the first increment ships only `Tc2_Math`'s `PI` (a constant), so there are no callable POUs to be unimplemented. I removed `REQ-CL-analyzer-005` and `REQ-CL-sources-009` accordingly and kept the "Safety first" principle.

---

specs/design/compatibility-library-format.md
garretfick: Changed. Libraries are **installed on disk** and read at runtime — not embedded in the compiler binary. The playground is served the same on-disk files as plain text. Propagated to the support doc and plan (loader reads installed libraries at runtime).

---

specs/design/compatibility-library-format.md
garretfick: Removed. The install location and discovery/search mechanism (the `IRONPLC_LIBRARY_PATH` idea) are gone from the format doc — it now just states they're defined separately when needed and are not part of the library format.

---

specs/design/compatibility-library-format.md
garretfick: Removed the "made up front / like the container format" preamble. The overview now just states what the doc specifies.

---

specs/design/compatibility-library-format.md
garretfick: Done — the layout is now version-named subdirectories:

```
Tc2_Math/
├── library.toml
└── 1.0.0/
    └── Tc2_Math.st
```

`REQ-LF-sources-001` specifies exactly this, and `default_version` in the manifest selects the version when a reference doesn't pin one.

---

specs/design/compatibility-libraries.md
garretfick: Removed the `--allow-math-constants` preamble from the support doc's overview; it now just states what the mechanism does.

---

PR #1318, opened 2026-08-05
feat(libraries): compatibility libraries phase 1 — Tc2_Math PI

garretfick:
Implement the first increment of the Compatibility Libraries design: IronPLC
ships dormant, named bundles of vendor declarations that are activated out of
band and injected into the compilation unit so their symbols resolve under
their exact vendor names.

Key outcome: PI for TwinCAT. Activating the bundled Tc2_Math library (via
`--library Tc2_Math`) injects `VAR_GLOBAL CONSTANT PI : LREAL`, so a TwinCAT
initializer like `d2r : LREAL := PI/180.0;` resolves and constant-folds with
no new keyword and no per-symbol flag.

- On-disk package format + loader in `sources` (`libraries::{manifest, mod}`):
  a `library.toml` manifest (name, vendor, default_version, references) plus
  per-version subdirectories of `.st` declarations, read from disk at runtime
  (REQ-LF-sources-001/002/004, REQ-CL-sources-002).
- Bundled `Tc2_Math` library defining `PI`.
- Activated-library set carried on the project; injected ahead of user source
  in `run_semantic_analysis` (base stdlib -> library -> user), so a user
  declaration shadows a library one (REQ-CL-analyzer-001..004,006).
- `--library <name>` (repeatable) CLI option wired through check/echo/tokenize/
  compile (REQ-CL-sources-005/006).
- New problem codes P6010 (LibraryManifestInvalid) and P6011 (LibraryNotFound),
  with docs.
- Spec-conformance wiring for sources/analyzer/plc2plc/playground; deferred
  markers (phases 2-4) are `#[ignore]`d so the orphan guard and meta-tests pass.

Deferred to later phases: reading library references from `.plcproj` (phase 2),
plc2plc round-trip fidelity and playground activation (phase 3), and the
provenance walk over bundled manifests (phase 4).

Claude-Session: https://claude.ai/code/session_01VSQqtNdCPJhkASVs5bfToJ


compiler/ironplc-cli/bin/main.rs
```diff
@@ -292,21 +300,31 @@ pub fn main() -> Result<(), String> {
...
+            false,
+        ),
+        Action::Echo { file_args } => cli::echo(
+            &file_args.files,
+            file_args.compiler_options(),
+            &file_args.libraries,
+            false,
+        ),
+        Action::Tokenize { file_args } => cli::tokenize(
+            &file_args.files,
```
garretfick: Tokenize should not have libraries - I don't see why this would be needed

compiler/ironplc-cli/src/cli.rs
```diff
@@ -32,9 +32,10 @@ use ironplc_project::{FileBackedProject, Project};
 pub fn check(
     paths: &[PathBuf],
     compiler_options: CompilerOptions,
+    libraries: &[String],
```
garretfick: Can we make this LibraryName instead of string. That would make it clear if this is a path or a name. It is a name.

compiler/problems/resources/problem-codes.csv
```diff
@@ -99,6 +99,8 @@ P6006,UnsupportedEncoding,File encoding is not supported
 P6007,UnsupportedFileType,File type is not supported
 P6008,XmlUnsupportedVersion,PLCopen XML version is not supported
 P6009,OutputPathConflictsWithInput,Output path conflicts with an input file
+P6010,LibraryManifestInvalid,Compatibility library manifest is malformed or missing a required field
+P6011,LibraryNotFound,Referenced compatibility library is not bundled with the compiler
```
garretfick: Can we just say "not found". instead of bundled with compiler. We don't know about bundling

compiler/sources/resources/compat-libraries/Tc2_Math/library.toml
```diff
@@ -0,0 +1,6 @@
+name = "Tc2_Math"
+vendor = "Beckhoff Automation GmbH"
+default_version = "1.0.0"
+references = [
+  "https://infosys.beckhoff.com/content/1033/tcplclib_tc2_math/ — Tc2_Math global constants (PI)",
```
garretfick: This reference doesn't have PI defined. Where does this actually come from?

compiler/sources/src/libraries/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+impl LibraryRegistry {
+    /// The registry of libraries bundled with the compiler.
+    ///
+    /// Libraries are read from disk at runtime (not embedded in the binary):
+    /// the root is the crate's `resources/compat-libraries` directory. The
+    /// install location and search mechanism for a shipped binary are defined
+    /// separately (out of scope for the on-disk format; see the format design's
+    /// *Installation* section).
```
garretfick: The path is relative to the current application exe. We should get the path of the application, then look in "resources/libs". In there we find the libraries.

garretfick (commented):
See comments

compiler/ironplc-cli/src/cli.rs
garretfick: Done. Added a `LibraryName` newtype in `ironplc_sources` (with `Display`/`FromStr`/`From<&str>`/`From<String>`) and threaded it through the activation API (`SourceProject`, the registry `load`/`contains`) and the CLI (clap now collects `Vec<LibraryName>`), so a name is no longer interchangeable with a path.

---

compiler/ironplc-cli/bin/main.rs
garretfick: Agreed. Moved `--library` off the shared `FileArgs` and onto only the `check` and `compile` subcommands. `tokenize` (and `echo`, which likewise runs no analysis) no longer accept it — `ironplcc tokenize --library …` now errors with "unexpected argument".

---

compiler/problems/resources/problem-codes.csv
garretfick: Done. P6011's message is now just "Referenced compatibility library not found", and the runtime label reads ``compatibility library `X` not found``. I also softened P6010's doc wording away from "bundled" for consistency, since libraries are installed alongside the app rather than baked into the binary.

---

compiler/sources/resources/compat-libraries/Tc2_Math/library.toml
garretfick: Good catch — I'd cited the library *index* page, not where the constant lives. Two parts to the answer:

- **The value**: `PI` is the mathematical constant π. Under the authoring policy that's a Tier A, math-dictated fact, so it's authored to `LREAL` precision rather than copied from any vendor expression — the number itself carries no provenance concern.
- **The name/membership**: I repointed the reference to the specific Tc2_Math "Global constants" page (`…/tcplclib_tc2_math/68453899.html`), which is the right sub-page for `PI` as a Tc2_Math global.

Transparency: Beckhoff's InfoSys returns `403` to my automated fetches, so I couldn't machine-verify that exact page lists `PI` (vs. only the library-version constant) or a specific documented precision. Since you authored the design that names Tc2_Math as PI's home — could you confirm that's the correct citation, and whether Beckhoff documents a particular `PI` value/precision you'd want mirrored? Happy to adjust the reference or the literal to match.

---

compiler/sources/src/libraries/mod.rs
garretfick: Done. `LibraryRegistry::bundled()` now derives its root from the running executable — `std::env::current_exe()` → parent → `resources/libs` — and reads libraries from there. Renamed the on-disk dir from `resources/compat-libraries` to `resources/libs` to match.

One wrinkle I want to flag: in dev/test the executable lives under `target/` with no resources beside it, so `cargo run`/`cargo test` would find nothing. I kept a fallback to the crate's source `resources/libs` for exactly that case (only used when the exe-relative dir is absent). Also, nothing yet copies `resources/libs` next to the installed binary in the installer/packaging step, so an *installed* `ironplcc` won't find libraries until that's wired up. I left that as a follow-up rather than expanding this PR into packaging — let me know if you'd prefer it here.

---

compiler/sources/resources/libs/Tc2_Math/library.toml
```diff
@@ -0,0 +1,10 @@
+name = "Tc2_Math"
+vendor = "Beckhoff Automation GmbH"
+default_version = "1.0.0"
+# `PI` mirrors the TwinCAT Tc2_Math global constant of the same name. Its value
+# is the mathematical constant pi -- a Tier A, math-dictated fact rather than a
+# copied expression -- authored to LREAL precision. See the authoring policy at
+# specs/steering/compatibility-library-authoring.md.
+references = [
+  "https://infosys.beckhoff.com/content/1033/tcplclib_tc2_math/68453899.html -- Tc2_Math global constants documentation",
```
garretfick: The right file is https://infosys.beckhoff.com/english.php?content=../content/1033/tcplclib_tc2_system/31084171.html&id=

compiler/sources/resources/libs/Tc2_Math/1.0.0/Tc2_Math.st
```diff
@@ -0,0 +1,10 @@
+(* Tc2_Math compatibility library -- global constants.
+
+   This file reproduces the *interface* (the constant name PI and its value)
+   of Beckhoff's TwinCAT Tc2_Math library for interoperability. IronPLC is an
+   independent open-source project and is not affiliated with, endorsed by, or
+   sponsored by Beckhoff Automation GmbH. See
+   specs/steering/compatibility-library-authoring.md. *)
+VAR_GLOBAL CONSTANT
+    PI : LREAL := 3.14159265358979;
```
garretfick: 3.1415926535897932384626433832795

compiler/sources/resources/libs/Tc2_Math/1.0.0/Tc2_Math.st
```diff
@@ -0,0 +1,10 @@
+(* Tc2_Math compatibility library -- global constants.
```
garretfick: The right library is `Tc2_System`

compiler/sources/resources/libs/Tc2_Math/library.toml
```diff
@@ -0,0 +1,10 @@
+name = "Tc2_Math"
```
garretfick: This should be `Tc2_System`

compiler/sources/resources/libs/Tc2_Math/1.0.0/Tc2_Math.st
garretfick: We also need to fix the Path for these files (two places)

garretfick (commented):
Corrected the library information

compiler/sources/resources/libs/Tc2_Math/library.toml
garretfick: Done — renamed the library to `Tc2_System` throughout: the package directory and the `.st` file (now `resources/libs/Tc2_System/1.0.0/Tc2_System.st`), the manifest `name`, and every code/test/fixture reference. I also updated the plan and design docs to the corrected identity, while keeping the legitimate `Tc2_Math` mentions intact (the `FLOOR` signature example, the future Tc2_Math numerics note, and the real `.plcproj` appendix grounding). Thanks for the correction and the authoritative pointers.

---

compiler/sources/resources/libs/Tc2_Math/1.0.0/Tc2_Math.st
garretfick: Done — `PI` is now `3.1415926535897932384626433832795` (rounds to the nearest `LREAL`/`f64`, i.e. full double precision).

---

compiler/sources/resources/libs/Tc2_Math/library.toml
garretfick: Done — the manifest `references` entry now points to that Tc2_System page.

---

PR #1323, opened 2026-08-05
Replace coarse `allow_iec_61131_3_2013` with per-feature flags

garretfick:
## Summary

Replaces the monolithic `allow_iec_61131_3_2013` boolean flag with two granular descriptor-based flags that independently control Edition-3 features. This allows vendor dialects to adopt specific Edition-3 features without forcing adoption of the entire edition.

## Changes

- **Removed** the hand-written `allow_iec_61131_3_2013` field from `CompilerOptions` and its special-case handling in `from_dialect()`
- **Added** `allow_long_time_types` descriptor (tagged `[Iec61131_3Ed3]`) to gate the long-time-type keywords (`LTIME`, `LDATE`, `LTOD`, `LDT`)
- **Extended** `allow_ref_to` descriptor with `Iec61131_3Ed3` tag to gate reference keywords (`REF_TO`, `REF`, `NULL`)
- **Updated** `xform_demote_keywords.rs` to use the new independent flags:
  - `demote_time_types = !options.allow_long_time_types`
  - `demote_ref = !options.allow_ref_to` (no longer coupled to edition boolean)
- **Added** `--allow-long-time-types` CLI flag in `ironplc-cli`
- **Updated** all test helpers and assertions to use the new flags instead of the removed boolean
- **Preserved** behavior: `from_dialect(Iec61131_3Ed3)` still yields both flags plus partial-access syntax, identical to the old edition boolean

## Implementation Details

The Edition-3 dialect is now an ordinary preset assembled from descriptors tagged with `Iec61131_3Ed3`, using the same mechanism that `allow_partial_access_syntax` already employs. This eliminates the special-case handling in `describe_dialects()` and makes the edition's feature set explicit and discoverable.

The decoupling of long-time types from reference keywords allows the RuSTy dialect (and others) to enable `REF_TO` syntax while keeping `LDT` etc. available as identifiers—a use case that was previously impossible with the coarse flag.

https://claude.ai/code/session_01U6nwdg8ii72PFNySurbzPg

### DIFF
--- a/compiler/ironplc-cli/bin/main.rs
+++ b/compiler/ironplc-cli/bin/main.rs
@@ -99,9 +99,15 @@ struct FileArgs {
     #[arg(long)]
     allow_c_style_comments: bool,
 
-    /// Allow REF_TO, REF(), and NULL syntax without enabling full Edition 3.
-    /// This is useful for libraries like OSCAT that use references but also
-    /// use Edition 3 type names (LDT, LTIME) as identifiers.
+    /// Allow the IEC 61131-3:2013 long-time-type keywords (LTIME, LDATE, LTOD,
+    /// LDT). Enabled by `--dialect=iec61131-3-ed3`.
+    #[arg(long)]
+    allow_long_time_types: bool,
+
+    /// Allow REF_TO, REF(), and NULL syntax (standardized in IEC 61131-3:2013)
+    /// without enabling the rest of Edition 3. This is useful for libraries like
+    /// OSCAT that use references but also use Edition 3 type names (LDT, LTIME)
+    /// as identifiers. Enabled by `--dialect=iec61131-3-ed3`.
     #[arg(long)]
     allow_ref_to: bool,
 
@@ -211,6 +217,7 @@ impl FileArgs {
         options.allow_empty_var_blocks |= self.allow_empty_var_blocks;
         options.allow_time_as_function_name |= self.allow_time_as_function_name;
         options.allow_c_style_comments |= self.allow_c_style_comments;
+        options.allow_long_time_types |= self.allow_long_time_types;
         options.allow_ref_to |= self.allow_ref_to;
         options.allow_reference_to |= self.allow_reference_to;
         options.allow_ref_arithmetic |= self.allow_ref_arithmetic;
--- a/compiler/ironplc-cli/bin/main.rs
+++ b/compiler/ironplc-cli/bin/main.rs
@@ -100,14 +100,14 @@ struct FileArgs {
     allow_c_style_comments: bool,
 
     /// Allow the IEC 61131-3:2013 long-time-type keywords (LTIME, LDATE, LTOD,
-    /// LDT). Enabled by `--dialect=iec61131-3-ed3`.
+    /// LDT). Without this flag those words remain available as identifiers.
     #[arg(long)]
     allow_long_time_types: bool,
 
     /// Allow REF_TO, REF(), and NULL syntax (standardized in IEC 61131-3:2013)
     /// without enabling the rest of Edition 3. This is useful for libraries like
     /// OSCAT that use references but also use Edition 3 type names (LDT, LTIME)
-    /// as identifiers. Enabled by `--dialect=iec61131-3-ed3`.
+    /// as identifiers.
     #[arg(long)]
     allow_ref_to: bool,
 
--- a/compiler/parser/src/options.rs
+++ b/compiler/parser/src/options.rs
@@ -157,9 +157,6 @@ macro_rules! define_compiler_options {
     ) => {
         #[derive(Debug, Default, Clone, Copy)]
         pub struct CompilerOptions {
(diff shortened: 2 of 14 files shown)


compiler/ironplc-cli/bin/main.rs
```diff
@@ -99,9 +99,15 @@ struct FileArgs {
     #[arg(long)]
     allow_c_style_comments: bool,
 
-    /// Allow REF_TO, REF(), and NULL syntax without enabling full Edition 3.
-    /// This is useful for libraries like OSCAT that use references but also
-    /// use Edition 3 type names (LDT, LTIME) as identifiers.
+    /// Allow the IEC 61131-3:2013 long-time-type keywords (LTIME, LDATE, LTOD,
+    /// LDT). Enabled by `--dialect=iec61131-3-ed3`.
```
garretfick: This is not the place to say what dialect enables

compiler/ironplc-cli/bin/main.rs
```diff
@@ -99,9 +99,15 @@ struct FileArgs {
...
+    #[arg(long)]
+    allow_long_time_types: bool,
+
+    /// Allow REF_TO, REF(), and NULL syntax (standardized in IEC 61131-3:2013)
+    /// without enabling the rest of Edition 3. This is useful for libraries like
+    /// OSCAT that use references but also use Edition 3 type names (LDT, LTIME)
+    /// as identifiers. Enabled by `--dialect=iec61131-3-ed3`.
```
garretfick: Also not the place to say what dialect enables this

compiler/parser/src/options.rs
```diff
@@ -256,9 +250,14 @@ define_compiler_options! {
     [Rusty, Codesys, TwinCat],
     allow_time_as_function_name,
 
-    "Allow REF_TO, REF(), and NULL without full Edition 3",
+    "Allow IEC 61131-3:2013 long-time-type keywords (LTIME, LDATE, LTOD, LDT)",
+    "--allow-long-time-types",
+    [Iec61131_3Ed3],
+    allow_long_time_types,
```
garretfick, line 256: Is this right. I thought others support this

compiler/parser/src/options.rs
```diff
@@ -412,24 +406,31 @@ mod tests {
...
-        assert!(!CompilerOptions::from_dialect(Dialect::Iec61131_3Ed2).allow_iec_61131_3_2013);
         assert_enabled_flags(Dialect::Iec61131_3Ed2, &[]);
     }
 
-    /// IEC 61131-3 Ed. 3 turns on the Edition-3 keyword set and, among dialect
-    /// extensions, only partial-access syntax (standardized in Edition 3).
+    /// IEC 61131-3 Ed. 3 is a preset assembled from the descriptors tagged with
+    /// `Iec61131_3Ed3`: the long-time-type keywords, the `REF_TO`/`REF`/`NULL`
+    /// reference keywords, and partial-access syntax. There is no longer a
+    /// separate coarse edition boolean -- the edition is exactly its descriptor
```
garretfick: Don't write history as a comment. Just describe the behavior

garretfick (commented):
See comments

compiler/parser/src/options.rs
garretfick: Good catch — you're right that CODESYS and TwinCAT support the long-time types. I've changed the descriptor from `[Iec61131_3Ed3]` to `[Iec61131_3Ed3, Codesys, TwinCat]` so those dialects now recognize `LTIME`/`LDATE`/`LTOD`/`LDT` as keywords.

RuSTy stays off the list on purpose: it deliberately keeps those words as identifiers for OSCAT (the `RuSTy` variant doc and the demote logic both call this out), so enabling long-time types there would break OSCAT parsing.

This is a deliberate behavior change from the old coarse flag (which was Ed3-only), so I also updated the consequences: the codesys/twincat `assert_enabled_flags` sets and their comments, the LSP codesys/twincat assertions, the docs **Enables:** lists and prose, and the two codegen end-to-end tests that asserted `LDT` is usable as a variable name under CODESYS/TwinCAT now assert it is a reserved keyword there.

---

compiler/ironplc-cli/bin/main.rs
garretfick: Done — removed the "Enabled by `--dialect=...`" clause from this flag's help so it describes only what the flag does.

---

compiler/ironplc-cli/bin/main.rs
garretfick: Done — dropped the dialect-enabling clause from `--allow-ref-to`'s help too.

---

compiler/parser/src/options.rs
garretfick: Done — rewrote the comment to describe the current behavior (Ed3 is the preset of descriptors tagged `Iec61131_3Ed3`) rather than narrating the removed boolean.

---

PR #1340, opened 2026-08-08
docs: ADR-0042 (library functions over intrinsics) + plan for compatibility-library bindings

garretfick:
## Summary

Docs-only PR: the decision rule and implementation plan for the support primitives needed to resolve #1217 (`LTRUNC`/`LMOD`), #1218 (`MODABS`), and #1246 (`BOOL_TO_STRING`/`LREAL_TO_FMTSTR`/`ADR`) per review — vendor functions externalized as compatibility libraries, native code in the compiler only where ST cannot express the semantics.

**ADR-0042 — Library Functions Over Compiler Intrinsics.** Compiler-provided names are limited to the IEC 61131-3 standard surface, and native implementations to behavior that cannot be a function. Every vendor-defined name ships as a compatibility library function: ST body when expressible; when not, the library's manifest binds the POU to an *unnamed* VM builtin (new func_ids are fine per ADR-0008, but they add no name to any scope — the library is the only way to reach them). Function-like operators that require compiler cooperation (`SIZEOF`, `ADR`) remain the dialect axis behind `--allow-*` flags.

**Plan — compatibility-library bindings.** Builds the deferred pieces of the compatibility-libraries design: the `["1.0.0".bindings]` manifest table (`intrinsic` / `declare-only`), the fail-if-unimplemented compile error (new P4046; declare-only calls still pass `check`, matching the corpus-check motivation of the original PRs), two new LREAL builtins (`trunc_lreal` 0x03A3, `fmod_lreal` 0x03A4), and the `Tc2_Math`/`Tc2_Utilities` libraries. Under the rule:

| Function | Mechanism |
|---|---|
| `LTRUNC`, `LMOD` | `Tc2_Math` POUs bound to the two new builtins (LREAL-preserving trunc / float modulo are inexpressible in ST) |
| `MODABS` | `Tc2_Math` ST body calling `LMOD` — no builtin |
| `BOOL_TO_STRING` | plain library ST function — trivially expressible, so no intrinsic and no func_id |
| `LREAL_TO_FMTSTR` | `Tc2_Utilities`, declare-only — surface lands, calls fail compile with P4046; native formatting is a follow-up |
| `ADR` | non-goal — dialect axis, deferred to the `POINTER TO` family work |

Open item flagged for review in the plan: where `BOOL_TO_STRING` lives (recommend bundled `Tc2_System`, since TwinCAT treats it as a compiler operator and real projects reference `Tc2_System` by default).

## Test plan
- [x] Docs only — no code changes
- [x] Full CI (`cd compiler && just`): compile, coverage ≥85%, clippy, fmt, dupes — all pass

https://claude.ai/code/session_01GECQRVfdE7uzpe5VaxzLx5

---
_Generated by [Claude Code](https://claude.ai/code/session_01GECQRVfdE7uzpe5VaxzLx5)_

### DIFF
--- a/specs/plans/2026-08-08-compatibility-library-bindings.md
+++ b/specs/plans/2026-08-08-compatibility-library-bindings.md
@@ -6,7 +6,7 @@ Deliver the support primitives that let the stalled vendor-function PRs
 ([#1217](https://github.com/ironplc/ironplc/pull/1217) `LTRUNC`/`LMOD`,
 [#1218](https://github.com/ironplc/ironplc/pull/1218) `MODABS`,
 [#1246](https://github.com/ironplc/ironplc/pull/1246)
-`BOOL_TO_STRING`/`LREAL_TO_FMTSTR`/`ADR`) be resolved the way review asked:
+`BOOL_TO_STRING`/`LREAL_TO_FMTSTR`) be resolved the way review asked:
 vendor functions **externalized as compatibility libraries**, with native code
 in the compiler only where IEC 61131-3 source cannot express the semantics.
 
@@ -51,7 +51,6 @@ Under the ADR-0042 rule each function lands as:
 | `FRAC(IN: LREAL): LREAL` | `Tc2_Math` POU with a math-dictated **ST body**: `FRAC := IN - LTRUNC(IN);` — no builtin (reviewer-requested addition; not in the original PRs) |
 | `BOOL_TO_STRING(IN: BOOL): STRING` | library **ST body** (`IF IN THEN BOOL_TO_STRING := 'TRUE'; ELSE ... 'FALSE'`) — trivially expressible, so no intrinsic and no func_id |
 | `LREAL_TO_FMTSTR(in: LREAL, iPrecision: INT, bRound: BOOL): STRING(255)` | `Tc2_Utilities` POU, **declare-only** — surface lands now, calls fail compile with the new code; native formatting is a follow-up plan |
-| `ADR` | **non-goal** — a function-like operator on the dialect axis, deferred to the future `POINTER TO` family work (interacts with ADR-0017/ADR-0021; a declared signature today would have to lie about its return type) |
 
 The mechanism is not TwinCAT-specific: bindings, declare-only, and the new
 problem code are generic library-format capabilities. The Tc2 libraries are
@@ -59,8 +58,6 @@ the first consumers because TwinCAT is a target.
 
 ## Non-goals
 
-- The `ADR` address operator and the `POINTER TO` family (dialect axis; its
-  own future design).
 - A native `LREAL_TO_FMTSTR` implementation (declare-only now; follow-up
   plan adds a formatting builtin and flips the binding).
 - Collision/precedence between libraries or vs. the base stdlib (design
@@ -228,5 +225,5 @@ declare-only compatibility library function", naming the library and POU.
   so it becomes a Phase 2 general capability (benefits every future library),
   and `BOOL_TO_STRING` ships declare-only until it lands — never a builtin.
 - Once the phases land, PRs #1217/#1218/#1246 are superseded: same names,
-  same behavior, delivered per ADR-0042 (`ADR` deferred to the `POINTER TO`
-  work; native `LREAL_TO_FMTSTR` as a follow-up plan).
+  same behavior, delivered per ADR-0042 (native `LREAL_TO_FMTSTR` as a
+  follow-up plan).
--- a/specs/plans/2026-08-08-compatibility-library-bindings.md
+++ b/specs/plans/2026-08-08-compatibility-library-bindings.md
@@ -24,7 +24,7 @@ The primitives, from the deferred *Future Goals* of the
 
    ```toml
    ["1.0.0".bindings]
-   LTRUNC = { intrinsic = "trunc_lreal" }
+   LTRUNC = { intrinsic = "trunc_f64" }
    ```
 
    or **declare-only** (`POU = "declare-only"`): the signature exists so the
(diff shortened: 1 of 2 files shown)


garretfick:
**Note: four pending review comments on ADR-0042 were accidentally deleted** while I was trying to reply to them (they were part of an unsubmitted review, and a `delete pending review` call removed them). Restoring them verbatim here so nothing is lost, with responses. All four are addressed in commit c1658e9.

---

> **On the "No IronPLC dialect" decision driver (line 29):** Don't confuse dialect with vendor libraries. We don't have a dialect, we haven't said IronPLC won't have libraries. However we have said compatibility.

Fixed — the driver is now **Compatibility**: a name is in scope exactly when the corresponding real-world library is activated, so code that compiles in IronPLC also compiles in the environment it targets. It now states explicitly that dialect (syntax, ADR-0036) is a separate axis, and that IronPLC may ship its own libraries (`vendor = "IronPLC"` is already in the manifest format) — what's decided is that such names arrive *as libraries* with the same activation model, never as unconditional compiler names.

> **On the "Scalability" driver (line 31):** Different vendors might have different behavior for the same library function

Added as its own decision driver (**Per-vendor behavior divergence**): a library captures each vendor's behavior in that vendor's package (e.g. TwinCAT `FLOOR` taking `LREAL`); a single compiler intrinsic would have to pick one vendor's behavior for everyone.

> **On the "Behavioral fidelity" driver (line 39):** Are there others that we cannot represent? This is actually a key question that I want to answer.

Answered in a new section, *What IEC 61131-3 source cannot express*. The genuinely inexpressible set falls into four categories:
1. **Arithmetic primitives absent from the operator/stdlib surface** — LREAL-preserving truncation and floating modulo. These two builtins essentially close the category: MODABS, FRAC, CEIL, FLOOR, round-half-away are all math-dictated compositions of them.
2. **Bit-level reinterpretation across type families** (LREAL bits ↔ LWORD without numeric conversion) — checksum/serialization/float-inspection functions.
3. **Compile-time knowledge** — SIZEOF, ADR/BITADR/INDEXOF, layout introspection; not functions at all, dialect-axis operators.
4. **Runtime-service access** — time, I/O, files, persistence, task control; FB-shaped ones follow ADR-0003 intrinsics, function-shaped ones need bindings.

Everything else is expressible — "laborious in ST" is not inexpressible. One borderline case is recorded: `LREAL_TO_FMTSTR` is expressible in principle, but numerically-faithful float formatting is subtle, so the plan lands it declare-only and the follow-up decides ST vs. a formatting builtin on fidelity grounds.

> **On the "Simplicity of contribution" driver (line 42):** Also difficulty of testing. Intrinsics are easier to test. Libraires are harder

Added — a **Testability** decision driver acknowledging the trade-off runs the other way (intrinsics unit-test directly in Rust; library bodies only end-to-end), and a matching "Bad" consequence with the mitigation (the `codegen/tests/it/` end-to-end harness plus conformance tests that load every bundled library).

---

specs/plans/2026-08-08-compatibility-library-bindings.md
```diff
@@ -0,0 +1,214 @@
...
+[#1218](https://github.com/ironplc/ironplc/pull/1218) `MODABS`,
+[#1246](https://github.com/ironplc/ironplc/pull/1246)
+`BOOL_TO_STRING`/`LREAL_TO_FMTSTR`/`ADR`) be resolved the way review asked:
+vendor functions **externalized as compatibility libraries**, with native code
+in the compiler only where IEC 61131-3 source cannot express the semantics.
+
+The primitives, from the deferred *Future Goals* of the
+[Compatibility Libraries design](../design/compatibility-libraries.md):
+
+1. **Bindings** — a per-version `library.toml` table mapping a library POU to
```
garretfick: What is this?

garretfick: "Bindings" is the mechanism the compatibility-libraries design defers under *Future Goals* — the name comes from there. Today the library loader can only compile fully-defined ST, so a POU like `LTRUNC` (whose semantics can't be written in ST) can't ship in a library at all. A binding is a per-version entry in the library's `library.toml` that marks a POU as either:

- **backed by a native VM builtin** — the `.st` file still declares the full signature (so `check`/type-checking work unchanged), and codegen lowers calls to the named builtin instead of a `CALL`:
  ```toml
  ["1.0.0".bindings]
  LTRUNC = { intrinsic = "trunc_lreal" }
  ```
- **declare-only** (`POU = "declare-only"`) — the signature exists so a library's surface can land ahead of its implementation, and *calling* it is a compile error (the fail-if-unimplemented rule, item 2).

Expanded inline at this spot in e9ae8e4 so the plan defines it on first use.

---

specs/plans/2026-08-08-compatibility-library-bindings.md
```diff
@@ -0,0 +1,214 @@
...
+   a **VM builtin** or to **declare-only**.
+2. **Fail-if-unimplemented** — a call to a declare-only POU is a dedicated
+   compile error (new problem code), never silently-wrong codegen and never a
+   runtime trap. `check` still passes (the declaration resolves), matching the
+   corpus-check motivation of the original PRs.
+3. **Two new BUILTIN func_ids** for semantics inexpressible in ST: LREAL
```
garretfick, line 38: I think we should add `FRAC`.

garretfick: Added in a65400c — as a `Tc2_Math` ST body rather than a third builtin: once `trunc_lreal` exists, `FRAC` is math-dictated (`FRAC := IN - LTRUNC(IN);`), the same composition argument as `MODABS`. It's in the function table, package contents, the clean-room spec list (pinning that `FRAC` keeps its input's sign, `FRAC(-3.7) = -0.7`), and the end-to-end VM assertions.

---

specs/plans/2026-08-08-compatibility-library-bindings.md
```diff
@@ -0,0 +1,214 @@
...
+Under the ADR-0042 rule each function lands as:
+
+| Function | Mechanism |
+|---|---|
+| `LTRUNC(IN: LREAL): LREAL` | `Tc2_Math` POU bound to new `trunc_lreal` VM builtin — `TRUNC` clamps to `ANY_INT`, so LREAL-preserving truncation is inexpressible in ST |
+| `LMOD(IN1: LREAL, IN2: LREAL): LREAL` | `Tc2_Math` POU bound to new `fmod_lreal` VM builtin — no floating modulo exists in ST (`MOD` is integer) |
+| `MODABS(IN: LREAL, IM: LREAL): LREAL` | `Tc2_Math` POU with a math-dictated **ST body** calling `LMOD` — no builtin |
+| `BOOL_TO_STRING(IN: BOOL): STRING` | library **ST body** (`IF IN THEN BOOL_TO_STRING := 'TRUE'; ELSE ... 'FALSE'`) — trivially expressible, so no intrinsic and no func_id |
+| `LREAL_TO_FMTSTR(in: LREAL, iPrecision: INT, bRound: BOOL): STRING(255)` | `Tc2_Utilities` POU, **declare-only** — surface lands now, calls fail compile with the new code; native formatting is a follow-up plan |
+| `ADR` | **non-goal** — a function-like operator on the dialect axis, deferred to the future `POINTER TO` family work (interacts with ADR-0017/ADR-0021; a declared signature today would have to lie about its return type) |
```
garretfick: Omit this entirely from the plan since it isn't something we are doing in this work.

garretfick: Removed in bd59fca — the `ADR` address operator no longer appears anywhere in the plan (function table, non-goals, wrap-up, or the intro's function list).

---

Issue #1292, opened 2026-08-02
TwinCAT discovery: resolve projects via .sln → .tsproj → PrjFilePath instead of recursive .plcproj walk


garretfick:
## Background

Follow-up from PR #1279 ([review discussion](https://github.com/ironplc/ironplc/pull/1279#discussion_r3699746896)), part of the broader TwinCAT vendor dialect support (#1199).

TwinCAT project discovery in `compiler/sources/src/discovery/mod.rs` (`detect_twincat`, currently around lines 170–194) works by recursively walking a directory tree and collecting every `.plcproj` file. PR #1279 improved this to merge multiple sub-projects into one compilation unit, but the underlying approach — finding projects by globbing for `.plcproj` files — is a heuristic rather than how Visual Studio / TwinCAT actually models a solution.

The authoritative structure is:

- The **`.sln`** file lists the projects in the solution.
- Each TwinCAT project **`.tsproj`** file names its live PLC project via `PrjFilePath` (e.g. `PrjFilePath="MONETNRuntime\MONETNRuntime.plcproj"`).

So the correct resolution chain is **`.sln` → `.tsproj` → `PrjFilePath` → `.plcproj`**.

## Why the current approach is fragile

The recursive `.plcproj` walk can silently pick the wrong project. A concrete real-world example from the discussion: a `MONETNRuntime` folder containing both `MONETNRuntime.plcproj` and a stale `MONETRuntime.plcproj` left over from a project rename. The current per-directory dedup keeps only the right one purely because `"MONETNRuntime"` happens to sort before `"MONETRuntime"` alphabetically — that is luck, not correctness. The `.tsproj`'s `PrjFilePath` is what unambiguously identifies the live `.plcproj`.

## Proposed work

Add `.sln`/`.tsproj`-based discovery for TwinCAT projects:

1. Parse the `.sln` file (custom line-oriented format, not XML) to enumerate the projects in the solution.
2. Parse each `.tsproj` (XML) to read `PrjFilePath` and resolve the live `.plcproj` for that project.
3. Merge the resulting set of `.plcproj` files into one compilation unit (reusing the merging logic already added in #1279).
4. Keep the current recursive `.plcproj` walk as a **fallback** for directory trees that have no `.sln` (so bare-`.plcproj` cases and non-solution directories still work).

This decouples correctness from filename sort order and matches how the toolchain actually resolves a solution.

## Acceptance criteria

[...]

PR #1342, opened 2026-08-10
fix(twincat): resolve projects via .sln -> .tsproj -> PrjFilePath

thusser:
## Summary

Follow-up from #1279's review, filed as #1292. Part of #1199 (TwinCAT vendor dialect support).

`detect_twincat` in `compiler/sources/src/discovery/mod.rs` found `.plcproj` files by recursively globbing for the extension, keeping only the first (sorted) per directory. That's a heuristic, not how TcXaeShell actually resolves a solution -- and it can silently pick the wrong project.

Real example from a private TwinCAT checkout: a directory containing both `MONETNRuntime.plcproj` (current) and a stale `MONETRuntime.plcproj` (leftover from a project rename). The old dedup picked the right one only because `"MONETNRuntime"` sorts before `"MONETRuntime"` -- alphabetical luck, not correctness.

The authoritative chain is `.sln` -> `.tsproj` -> `PrjFilePath`:
- The `.sln` (a line-oriented format, not XML) lists a solution's projects; only entries pointing at a `.tsproj` are relevant, other VS project types are ignored.
- Each `.tsproj` (XML) names its live `.plcproj` via one or more nested `<Project PrjFilePath="...">` elements.

This PR adds that resolution path ahead of the existing recursive walk, which becomes the fallback when there's no unambiguous `.sln` (so bare-`.plcproj` directories keep working exactly as before). `.splcproj` (TwinSAFE safety projects) entries are explicitly skipped -- different compilation model, out of scope. No changes to `.plcproj` parsing itself or to the existing multi-sub-project merge logic; this only changes which `.plcproj` paths feed into it.

## Test plan
- [x] New unit tests: `.sln`-driven resolution, non-`.tsproj` `.sln` entries ignored, `.splcproj` entries skipped, ambiguous multi-`.sln` falls back to the walk, multi-sub-project merge via one `.tsproj`
- [x] Regression test modeled directly on the real stale-rename case (named so it would sort first, so it fails under the old glob-only logic and passes under this fix)
- [x] Verified against the real TwinCAT solution that exhibits this exact bug: before the fix, discovery would depend on alphabetical luck; after the fix, all 23 discovered files correctly come from the live `MONETNRuntime/` project, none from the stale `MONETRuntime/` directory
- [x] Full CI (`cd compiler && just`): build, coverage ≥85%, clippy, fmt, dupes check -- all pass


compiler/sources/src/discovery/mod.rs
```diff
@@ -158,27 +158,175 @@ fn walk_files(dir: &Path, out: &mut Vec<PathBuf>) {
...
+    Some(merge_plcproj_projects(dir, plcproj_paths))
+}
+
+/// Find the single `.sln` file directly in `dir` (not recursive -- a
+/// `.sln` is always the file a user/tool points at directly, unlike
+/// `.plcproj`, which real layouts commonly nest several levels deep).
+///
+/// Returns `None` if there is no `.sln`, or more than one: ambiguous,
+/// callers should fall back to the recursive walk rather than guess.
+fn find_sln(dir: &Path) -> Option<PathBuf> {
```
garretfick: Key issue - this file is now more than 1000 files. I've normally set a maximum file length of 1000 lines.

I would like SLN parsing to be in a separate file, such as `sln.rs`. We normally do not write implementation code in the `mod.rs` file.

garretfick (commented):
See inline comment

garretfick:
The `.sln` → `.tsproj` → `PrjFilePath` resolution itself looks right, and the stale-rename regression test is a good catch. The thing I'd like resolved before this lands is the **fallback to the recursive walk**.

## The concern

`resolve_plcproj_via_sln` returns an empty `Vec` for several distinct conditions, and `detect_twincat` treats all of them identically as "fall back to the walk":

- no `.sln` in the directory — the intended fallback case
- **more than one `.sln`** — `find_sln` returns `None` on `sln_files.len() != 1`
- **`.sln` unreadable** — `parse_sln`'s `let Ok(content) = fs::read_to_string(...) else { return Vec::new() }`
- **`.tsproj` unreadable or not well-formed XML** — same pattern in `resolve_plcproj_via_tsproj`

In the last three an authoritative manifest *was* found and simply couldn't be resolved. Dropping to `collect_plcproj_via_walk` there silently reinstates the exact failure this PR sets out to fix: that function still does `candidates.sort()` + first-per-directory, so the stale `MONETRuntime.plcproj` case goes back to being decided by alphabetical order. The multi-`.sln` case is arguably worse, since it merges `.plcproj` files across solutions the user never asked for.

## Where the walk's justification comes from

The doc comment justifies recursion with "real TwinCAT layouts commonly nest `.plcproj` files several levels below the directory a user would naturally point the tool at." True, but it's compensating for an unstated convention about *which* directory to open. State the convention — **open the folder containing the manifest** — and the nesting is traversed by reference (`.sln` → `.tsproj` → `PrjFilePath`), never by search.

A bare-`.plcproj` project is the same story one rung down: open the folder containing the `.plcproj`, and a non-recursive `read_dir` finds it — the same call shape as `find_sln`. That covers the LSP too, which can only ever hand `discover()` a directory (`lsp.rs:133` → `initialize_many`), so it can't take a file argument.

## Proposed rule

Manifests in priority order `.sln` > `.tsproj` > `.plcproj`, so a solution folder with a stray `.plcproj` still resolves via the `.sln`:

| Input | Behavior |
|---|---|
| file argument: `.sln` / `.tsproj` / `.plcproj` | authoritative, use it |
| folder, exactly one manifest at the top tier present | authoritative; resolution failure → diagnostic, never a heuristic |
| folder, multiple manifests at the same tier | diagnostic naming them; user passes one explicitly |
| folder, no manifest, one found nested below | diagnostic pointing at it |
| folder, no manifest anywhere | unstructured: enumerate loose source files (unchanged) |

Rows 1 and 3 are a pair — don't add the ambiguity diagnostic without also teaching `enumerate_files` (`cli.rs:259`) to accept a manifest as a file argument. Today a `.sln` passed on the command line isn't parsed at all; `metadata.is_file()` loads it as source text.

Bare-`.st` directories are unaffected: they never reach the TwinCAT detector, and `detect_fallback` stays recursive — with no manifest format in play, enumeration *is* the project definition, so there's nothing to guess between.

Row 4's hint is the only remaining use of `walk_files` in the TwinCAT path, and it's purely for the error message — it never selects sources. It should fire only when `detect_fallback` also comes back empty, so a directory of loose `.st` files that happens to contain an unrelated solution in some subfolder doesn't get a spurious warning.

## The type change

`detect_twincat`'s current `Option<Result<DiscoveredProject, Diagnostic>>` can't express that rule. There's no way to say "not my project, but here's something worth telling the user," and it doesn't distinguish "no manifest here" from "manifest found but unresolvable" — which is precisely the distinction that decides whether falling through to the next detector is legitimate. Something like:

```rust
/// The outcome of running one detector against a directory.
enum Detection {
    /// No manifest of this kind here. Try the next detector.
    NotDetected,
    /// No manifest here, but one exists deeper in the tree. Still not
    /// this detector's project, so the chain continues -- the hint is
    /// surfaced only if nothing else claims the directory.
    NotDetectedWithHint(Diagnostic),
    /// Detected and fully resolved.
    Detected(Box<DiscoveredProject>),
    /// Detected but unresolvable: ambiguous manifests, unreadable
    /// `.sln`, malformed `.tsproj`. Authoritative -- never falls
    /// through to another detector.
    Failed(Diagnostic),
}
```

`Failed` never falls through, `NotDetected*` always does, and `discover` owns the policy about when to surface a hint — the detector doesn't need to know the fallback's outcome. (If the "not detected, but with data" variant feels awkward, `NotDetected(Vec<Diagnostic>)` with an empty vec in the common case collapses the two and reads about as well.)

New diagnostics need a row in `compiler/problems/resources/problem-codes.csv` — next free is P6012 — and a page under `docs/reference/compiler/problems/`.

## Breaking change worth calling out

Dropping the recursive walk means anyone whose setup currently works by accident — opened the repo root, walk happened to find the right thing — gets a diagnostic instead of results. I'm OK with that change in behavior - I prefer to go with correctness.

---

garretfick:
I implemented a different approach to get rid of the general "walk all files" if we know what the format is.

PR #1361, opened 2026-08-12
feat(twincat): METHOD declarations + static dispatch resolution (ADR-0041 phase 1, slice 1)

thusser:
## Summary

First slice of ADR-0041 Phase 1 (static method/property dispatch). Adds:

- `METHOD ... END_METHOD` declarations on a function block, gated by `--allow-fb-inheritance` (same flag as the other OOP keywords: `EXTENDS`/`IMPLEMENTS`/`INTERFACE`/`ABSTRACT`)
- `instance.Method(args)` call statements
- Static dispatch resolution: a method call is resolved against the instance's own methods, then its `EXTENDS` base, then that base's base, and so on. New diagnostic **P4046 `MethodNotFound`** when nothing in the chain declares the method; existing function/FB-invocation diagnostics reused for arity/type mismatches
- plc2plc round-trip rendering for both declarations and calls (also fixes a pre-existing gap: `FbCall` rendering had no trailing `;`, breaking round-trip whenever a call statement was followed by another statement — `MethodCall` shares the fix)

This means `ironplcc check` against real TwinCAT code that uses methods now produces real diagnostics instead of a parse error.

## Explicitly out of scope (follow-up slices)

- **Codegen** — method calls parse and resolve, but don't yet execute. `compile_stmt.rs` returns `Diagnostic::todo` for `StmtKind::MethodCall`. This is the largest remaining piece (receiver-pointer calling convention, symbol mangling, FB instance memory layout) and is its own slice.
- `THIS^`/`SUPER^`
- `PROPERTY ... GET ... SET ... END_PROPERTY`
- Method calls in expression position (`IF fb.IsMoving() THEN`) — statement-only for now, a deliberate scope decision
- `.TcPOU` XML method wiring (TwinCAT stores each method as a separate `<Method>` XML element; `transform.rs` currently always sets `methods: vec![]`)

Full design and remaining task list: `specs/plans/2026-08-12-oop-method-declarations-static-dispatch.md`.

## Test plan

- [x] `cd compiler && just` (compile + coverage + lint) clean
- [x] New parser tests (`parser/src/tests/methods.rs`, 7 tests)
- [x] New plc2plc round-trip tests (`plc2plc/src/tests/methods.rs`, 4 tests)
- [x] New analyzer tests (`analyzer/src/rule_method_call_declared.rs`, 5 tests: own type, one/two levels of `EXTENDS`, not-found, wrong arg count)


docs/reference/compiler/problems/P4046.rst
```diff
@@ -0,0 +1,51 @@
+=====
+P4046
+=====
+
+.. problem-summary:: P4046
+
+This error occurs when a method call (``instance.MethodName(args)``, the
+CODESYS/TwinCAT dialect extension) refers to a method that is not
```
garretfick: This is not just CODESYS/TwinCAT. OOP is also part of edition 3.

compiler/sources/src/discovery/mod.rs
```diff
@@ -173,27 +173,175 @@ fn walk_files(dir: &Path, out: &mut Vec<PathBuf>) {
...
 }
 
-/// Detect a TwinCAT project by searching for `.plcproj` files.
+/// Detect a TwinCAT project by resolving `.plcproj` files.
 ///
-/// Searches recursively, since real TwinCAT layouts commonly nest
-/// `.plcproj` files several levels below the directory a user would
-/// naturally point the tool at (e.g. a Visual-Studio-style
-/// solution/project structure).
+/// Tries `.sln` -> `.tsproj` -> `PrjFilePath` resolution first
```
garretfick: This looks like a stacked PR. I've requested changes on the SLN part that would apply here.

compiler/dsl/src/common.rs
```diff
@@ -2765,6 +2765,41 @@ pub struct FunctionBlockDeclaration {
...
+/// See ADR-0041 ("Staged Method/Property Dispatch and Interface Values"),
+/// Phase 1: static dispatch. Resolution of calls against this declaration
+/// (own methods first, then the `EXTENDS` chain) is implemented outside
+/// the AST, in `ironplc-analyzer`.
+#[derive(Clone, Debug, PartialEq, Recurse, Located)]
+pub struct MethodDeclaration {
+    pub name: Id,
+    pub return_type: Option<FunctionReturnType>,
+    pub variables: Vec<VarDecl>,
+    pub edge_variables: Vec<EdgeVarDecl>,
```
garretfick, line 2800: I didn't see in the twincat docs where edge var dec is allowed. Can you point me to where the docs say edge variables are allowed?

garretfick (commented):
see inline comments

compiler/plc2plc/src/tests/methods.rs
```diff
@@ -0,0 +1,120 @@
+//! OOP extension: METHOD declarations and instance.Method(args) calls,
+//! round-trip. See
+//! specs/plans/2026-08-12-oop-method-declarations-static-dispatch.md.
+
+use super::common::*;
+
+#[test]
+fn write_to_string_when_method_has_params_and_return_then_round_trips() {
```
garretfick: Please see the corpus tests. Those have shorthand that remove a lot of duplicated code here.

garretfick (commented):
additional comments

garretfick:
This was built on top of a change for project discovery which I've taken a different direction. I merged something equivalent.

PR #1362, opened 2026-08-12
feat(twincat): codegen for method calls (ADR-0041 phase 1, slice 1)

thusser:
## Summary

Adds codegen so `instance.Method(args)` actually executes. #1361 (declarations, call parsing, static-dispatch resolution) was closed in favor of the maintainer's own equivalent implementation, merged as #1386 -- this PR is rebased directly onto that.

- New opcode `METHOD_CALL` reuses `FB_CALL`'s existing copy-in/copy-out machinery (confirmed by reading the VM: there's no heap/pointers at all, ADR-0021's flat variable table — each FB type shares one scratch region across instances, swapped via the instance's persistent `data_region` storage). A method's own params/locals are allocated contiguously after its owning type's field region so the VM's single-range `VariableScope` bounds check covers both self-field access and the method's own params/locals in one window.
- Proven end-to-end via real VM execution, not just opcode-level unit tests: `codegen/tests/it/end_to_end_methods.rs` compiles actual `.st` source through the full pipeline and asserts the FB instance's field value after the scan — including a method reading another field of the same instance, two sequential calls with different args, and named-argument resolution.

## Explicitly out of scope

- **Methods reached only via `EXTENDS`.** Already accepted by the analyzer (`rule_method_call_declared`, from #1386), but not by codegen: derived FB types don't currently get runtime storage for inherited fields at all (confirmed — `inherited_fields.rs` is used only for semantic name-duplication checks, never to flatten storage onto the derived type). That's a separate, pre-existing gap in `EXTENDS` support, not something to fix here. Calling an inherited method hits `Diagnostic::todo`.
- `THIS^`/`SUPER^`, `PROPERTY`, expression-position method calls — same as #1386.

Full design notes (VM calling-convention research, opcode layout) are in `specs/plans/2026-08-12-oop-method-declarations-static-dispatch.md`.

## Test plan

- [x] `cd compiler && just` (compile + coverage ≥85% + clippy + fmt + dupes) clean
- [x] New opcode pinned in `codegen/tests/it/wire_format.rs` (byte `0xF8`, completeness guard included)
- [x] New end-to-end tests (`codegen/tests/it/end_to_end_methods.rs`, 4 tests) proving real field mutation through a compiled-and-executed method call
- [x] Full existing test suite (workspace-wide) passes with zero regressions


garretfick:
Flagging one issue that I think is worth fixing before this lands — a `RETURN;` inside a void method leaves a value on the operand stack.

`compile_method.rs:183` sets `CurrentFunctionReturn::Scalar` unconditionally, including when `has_return_value` is false. `compile_stmt.rs:450` then compiles `RETURN;` as `LOAD_VAR + RET`, pushing a value — but the call site at `compile_stmt.rs:655` only pops twice `if has_return_value`, so `fb_ref` is left behind on every call.

This is worse than it first looks because the VM's operand stack is never reset: it's constructed once at start (`vm/src/vm.rs:82`, held at `:114`/`:307`) and `vm/src/stack.rs` has no `clear`/`reset`, nor does `run_round` truncate. So the leak accumulates *across scan rounds*, and since the stack buffer is sized to exactly `max_stack_depth` (`vm/src/buffers.rs:48`), a void method with an early `RETURN` called once per scan traps with `StackOverflow` after roughly `max_stack_depth` scans — seconds into a run on a fast task, with a fault that points nowhere near the cause.

The fix looks like one line: only set `Scalar` when `method.return_type.is_some()`, and leave void methods with `current_function_return = None` so `RETURN;` emits `RET_VOID`.

The tests don't catch it because no test in the PR puts a `RETURN;` inside a `METHOD` — the four e2e bodies are straight-line assignments — and `parse_and_run` runs a single scan round and asserts only on variable values, which a leaked stack slot doesn't affect. A regression test needs either the call in a loop within one scan, or a multi-round `parse_and_run_rounds` case.

Worth noting the underlying gap isn't yours: there's no stack-balance check anywhere in the compiler today, even though the emitter already tracks `current_stack_depth` (`emit.rs:46`) and discards it at `finalize_function`. That's being looked at separately on `main`, so nothing to do here beyond the one-line fix.

---

garretfick:
I did commit changes to address the stack issue.

thusser:
Good catch, thanks. Fixed in 1f24a0a7: `current_function_return` is now only set to `Scalar` when `method.return_type.is_some()`, so a void method's `RETURN;` falls through to the existing `None` branch and emits `RET_VOID`.

Added a regression test (`end_to_end_when_void_method_has_early_return_then_repeated_calls_do_not_overflow_stack`) that calls a void method with an early `RETURN;` in a loop within one scan round — confirmed it reproduces `StackOverflow` on the old code and passes with the fix. Full CI clean.

garretfick:
Merged in https://github.com/ironplc/ironplc/pull/1449

PR #1364, opened 2026-08-13
Implement Layer-1 debug-info swap for DAP breakpoints, variables, and stack traces

garretfick:
## Summary

Replaces the passthrough debug-info resolvers in `dap/debug_info.rs` with real implementations that use the container's line map, VAR_NAME, FUNC_NAME, STRING layouts, and source file table. Breakpoints now resolve by source line (with snap-to-next-executable behavior), variables render with their declared names and types, and stack frames show POU names with source locations.

## Key Changes

- **Breakpoint resolution** (`resolve_breakpoint`): 
  - Returns `Option<ResolvedBreakpoint>` with the snapped line and bytecode locations
  - Filters line-map entries by source file (exact path match, then basename match for absolute-vs-relative differences)
  - Snaps forward to the nearest executable line
  - Arms the statement-start offset for each function with code on the bound line

- **Variable rendering** (`render_variables`):
  - Looks up VAR_NAME entries by slot index to get source names and IEC type tags
  - Formats values using the shared `debug_format::format_variable_value` helper
  - Reads STRING values from the data region using the STRING layout table
  - Renders STRING literals with IEC 61131-3 escaping (`$T`, `$L`, `$P`, `$R`, `$$`, `$'`, `$XX` hex)
  - Falls back to `var[i]` / signed-i32 for unnamed slots or missing debug info

- **Stack frame resolution** (new `resolve_frame`):
  - Looks up POU names from FUNC_NAME, falling back to `function {id}`
  - Uses `DebugSection::lookup_source_location` to find the enclosing line for the paused PC
  - Returns source file name and recorded path from the source file table
  - Reports line 0 and no source when the line map has no hit

- **Server integration** (`server.rs`):
  - `set_breakpoints` now echoes the snapped line in verified breakpoints so the editor moves the marker to the actual bind location
  - `stack_trace_body` gains a `debug` parameter and maps `FrameInfo` into `StackFrame` with POU name and source location
  - `variables_body` passes the data region to `render_variables` for STRING reads

- **Test fixtures**: Updated to include line-map, source-file, and FUNC_NAME entries; breakpoint tests now set by source line instead of raw offset

## Implementation Details

- File matching is case-insensitive on Windows (ASCII) to absorb path separator and case differences
- Out-of-bounds STRING layout offsets/lengths degrade to `<not available>` rather than panicking
- WSTRING rendering is a v1 placeholder (same as the playground)
- Containers without a source file table are treated as single-source (no file filtering)
- The `FrameInfo` struct carries both file name (for display) and recorded path (for editor navigation)

## Testing

Comprehensive unit tests in `debug_info.rs` cover:
- Snap-to-next-line, exact-line hits, line past end
- Basename path matching and non-matching files
- Multiple offsets and functions on one line
- Variable rendering for BOOL/DINT/REAL, STRING from data region, out-of-bounds layouts, missing entries
- Frame resolution with and without FUNC_NAME/line-map hits
- IEC string literal escaping

Integration tests in `server.rs` verify end-to-end breakpoint setting, stack traces, and variable inspection against fixture containers with full debug info.

https://claude.ai/code/session_01UbYjygkU7ko7pnoBTEoZHU


compiler/vm-cli/src/dap/debug_info.rs
```diff
@@ -1,93 +1,561 @@
...
+        locations,
+    })
+}
+
+/// Whether a recorded source path refers to the same file as a requested
+/// path. Exact match first; otherwise the file names are compared (ASCII
+/// case-insensitively, for Windows) to absorb absolute-vs-relative and
+/// separator differences between what the compiler recorded and what the
+/// editor sends.
+fn file_matches(recorded: &str, requested: &str) -> bool {
```
garretfick, line 129: 1. is there not a Path structure we can use that will do the right comparison? I suspect not, but want to validate.
2. The eq ignore case is correct on Windows but wrong on Linux. I think we should be using a correct file comparison function so that this is right

compiler/vm-cli/src/dap/debug_info.rs
```diff
@@ -1,93 +1,561 @@
...
+    pub locations: Vec<(FunctionId, usize)>,
+}
+
+/// A stack frame resolved against FUNC_NAME and the line map.
+#[derive(Debug, PartialEq, Eq)]
+pub struct FrameInfo {
+    /// The POU name from FUNC_NAME, or `function {id}` when unnamed.
+    pub name: String,
+    /// 1-based source line of the paused instruction; `0` when unknown.
+    pub line: i64,
```
garretfick: I understand that JSON doesn't have unsigned types. That said, I'd like the conversion to be at serialization and not carried throughout. I'd like the type to be an alias so that we cannot confuse line, column and position values.

compiler/vm-cli/src/dap/debug_info.rs
```diff
@@ -1,93 +1,561 @@
...
+    // source file table predates per-file tracking: every entry is eligible.
+    let file_ids: Option<Vec<u16>> = if debug.source_files.is_empty() {
+        None
+    } else {
+        let ids: Vec<u16> = debug
+            .source_files
+            .iter()
+            .enumerate()
+            .filter(|(_, sf)| file_matches(&sf.path, source_path))
+            .map(|(id, _)| id as u16)
```
garretfick: If we convert early, then we have this `as` earlier

compiler/vm-cli/src/dap/debug_info.rs
garretfick: Both addressed in e565044.

**1. Is there a `Path` structure that does the right comparison?** Your suspicion is correct — there isn't. I tested it rather than assume, on this Linux host:

```
Path::new(r"C:\work\demo.st").file_name()  => Some("C:\\work\\demo.st")  // whole string
Path::new("Demo.st") == Path::new("demo.st")  => false
Path::new("./demo.st") == Path::new("demo.st")  => false
```

`Path` parses and compares by the *host* platform's rules, so a Windows-recorded path is one opaque component on Unix. It's also always case-sensitive, even on Windows — `std` deliberately doesn't fold case, since it can't know the filesystem's semantics. And it doesn't fold a leading `./` (it does normalize interior `/./` and `//`, but not `..`, which needs symlink knowledge).

`fs::canonicalize` (or the `same_file` crate) would give true identity, but both require the paths to exist on this machine — not guaranteed for a path recorded at compile time.

`Path` does offer one genuinely useful primitive: `ends_with` is component-wise, so `/w/demo.st` ends_with `demo.st` is true while `/w/xdemo.st` is false. That solves suffix matching but not the case question, which is the half that was actually wrong. I've recorded these findings in a comment on `file_matches` so the next reader doesn't have to re-derive them.

**2. `eq_ignore_ascii_case` is wrong on Linux.** Agreed, and this was a real bug, not just a style issue — folding case there would bind a breakpoint set in `Demo.st` to code in `demo.st`, a genuinely different file. Case sensitivity now follows the host:

```rust
#[cfg(any(target_os = "windows", target_os = "macos"))]
fn file_names_match(recorded: &str, requested: &str) -> bool {
    recorded.eq_ignore_ascii_case(requested)
}

#[cfg(not(any(target_os = "windows", target_os = "macos")))]
fn file_names_match(recorded: &str, requested: &str) -> bool {
    recorded == requested
}
```

macOS is grouped with Windows because APFS/HFS+ are case-insensitive by default. That's a default, not a guarantee — a case-sensitive APFS volume exists — but there's no compile-time way to know, and matching the platform default is what other debug adapters do. The tests are now split by the same `cfg`, asserting a match on case-insensitive hosts and `None` on case-sensitive ones.

One thing I kept deliberately: `file_name` still splits on both `/` and `\` regardless of host, so a container compiled on Windows still resolves when debugged elsewhere. The tradeoff is that a Unix filename legitimately containing a backslash would mis-split — pathological enough that I favored the cross-platform case, but happy to flip it if you'd rather use host-only separators.

---

compiler/vm-cli/src/dap/debug_info.rs
garretfick: Done in e565044 — and rather than a type alias, I used the container's existing newtypes directly, which gets you the "same types" guarantee more strongly than an alias would.

`ResolvedBreakpoint.line` and `FrameInfo.line`/`.column` are now `SourceLine` / `SourceColumn` (the `u16` newtypes from `id_types.rs`) instead of `i64`. The wire value is narrowed once, on entry to this module:

```rust
fn source_line_from_dap(line: i64) -> Option<SourceLine> {
    u16::try_from(line).ok().map(SourceLine::new)
}
```

`u16::try_from` covers both out-of-range directions, so the old `if line < 0` guard is gone and a too-large line is rejected rather than truncated. That mattered: previously `65546` would have been cast down to `10` and bound as a valid-looking line. There's a test pinning both `65_546` and `i64::MAX` to `None`.

Widening back on the way out is `i64::from(...)`, which is lossless and infallible — no `as` involved:

```rust
line: Some(i64::from(resolved.line.raw())),
```

`SourceLine(0)` is already documented as "unknown" in `id_types.rs`, and that lines up exactly with the DAP convention of line 0 meaning no source location, so the unmapped-frame case now falls out of `SourceLine::default()` rather than a bare `0`.

---

compiler/vm-cli/src/dap/debug_info.rs
garretfick: Right — that `as u16` is gone in e565044, along with the others it was standing in for.

The file-id collection now stays in the container's type end to end, with a checked narrowing instead of a cast:

```rust
let ids: Vec<SourceFileId> = debug
    .source_files
    .iter()
    .enumerate()
    .filter(|(_, sf)| file_matches(&sf.path, source_path))
    .filter_map(|(index, _)| u16::try_from(index).ok().map(SourceFileId::new))
    .collect();
```

so the membership test is `ids.contains(&e.file_id)` on `SourceFileId` rather than on a bare `u16`.

Converting the line early removed the rest of them: `(e.source_line.raw() as i64) >= line` became `e.source_line.raw() >= requested.raw()`, and the bound-line comparison is now `e.source_line == bound` on `SourceLine` directly. Two more that were nearby:

- `entry.bytecode_offset as usize` → `usize::from(...)` (lossless, infallible)
- `entry.file_id.raw() as usize` → `usize::from(...)` for the source-table index

The one remaining narrowing is the frame's `pc`, which arrives from the VM as `usize`. It was `u16::try_from(pc).unwrap_or(u16::MAX)`, which silently clamped — a `pc` past `u16::MAX` would have resolved to whatever the last line-map entry happened to be. It now uses `u16::try_from(pc).ok()` and resolves to the "unknown" location instead, which is the honest answer given the line map stores offsets as `u16` and so cannot describe that offset at all.

The only `as` left in the module is `index as i64` for DAP frame ids in `server.rs`, which is an unrelated synthetic handle.

---

compiler/vm-cli/src/dap/debug_info.rs
```diff
@@ -1,93 +1,561 @@
...
+                },
+            }
         })
         .collect()
 }
 
+/// Format one variable's value per its IEC type tag. STRING values live in
+/// the data region (the slot is unused); everything else renders from the
+/// raw slot via the shared `debug_format` helper.
+fn variable_value(
```
garretfick, line 246: Ok - we largely have the same code for the playground. Can we make this the same code?

garretfick (commented):
see inline

compiler/vm-cli/src/dap/debug_info.rs
garretfick: Agreed — it was a near-copy of the playground's. Shared in 3f10b19.

`read_string_value`, `format_iec_string_literal`, and `StringReadError` now live in `container/src/debug_format.rs`, which is already the home for exactly this — its module doc opens with "These were previously duplicated across `ironplc-cli` (the LSP runner), `vm-cli`, and the playground WASM crate." This is the same class of duplication, so it belongs in the same place rather than in a new module.

The playground's shape won, not mine. It returned `Result<String, StringReadError>` distinguishing `OffsetOutOfBounds` from `LengthOutOfBounds`, which it needs to set `valid: false` and render differently from real string content; mine collapsed both to `Option`. The shared version keeps the richer errors and the DAP side just does `.ok()`, since it renders one placeholder either way.

Also moved the playground's seven tests for these into the container crate (they were more thorough than mine — named escapes, zero length, both error variants) and deleted my redundant one. The `dupes` check confirms it: exact duplicated lines went 3320 → 3284, and one duplicate group disappeared.

Worth noting this fixed a latent bug in my copy. I had written the header layout as `[max_len][cur_len][bytes…]` and hardcoded a 4-byte skip in a test fixture, but ADR-0035 defines it as `[max_length: u16][cur_length: u16][char_width: u16][data…]` — 6 bytes. The code was right because it used `STRING_HEADER_BYTES`, but the doc comment and fixture were wrong, and that would have misled the next person. Sharing one implementation means one place to get it right.

---

compiler/vm-cli/src/dap/debug_info.rs
garretfick: You're right — my previous pass converted at the module boundary, not at serialization, so the `i64` was still travelling one hop further than it needed to. Moved into the serde layer in 3f10b19.

The protocol structs in `dap/types.rs` now carry the container's newtypes, and serde does the conversion:

```rust
pub struct SourceBreakpoint {
    #[serde(deserialize_with = "source_coords::deserialize_opt_line")]
    pub line: Option<SourceLine>,
    #[serde(default, deserialize_with = "source_coords::deserialize_opt_column")]
    pub column: Option<SourceColumn>,
}

pub struct StackFrame {
    #[serde(serialize_with = "source_coords::serialize_line")]
    pub line: SourceLine,
    #[serde(serialize_with = "source_coords::serialize_column")]
    pub column: SourceColumn,
    ...
}
```

`i64` now appears in exactly one place — inside the private `source_coords` module. `resolve_breakpoint` takes an already-validated `SourceLine`, `server.rs` performs no conversion at all (the `i64::from(...)` calls I added last round are gone), and `source_line_from_dap` is deleted since serde does that job now.

On "alias": I went with the existing `SourceLine`/`SourceColumn` newtypes rather than `type Line = u16`, because a `type` alias in Rust is transparent — `type Line = u16; type Column = u16;` would still let you pass a column where a line is expected. The newtypes make that a compile error, which I read as the actual goal of your comment. If you did mean a literal alias, say so and I'll switch.

One judgement call worth flagging: deserialization is deliberately **lenient**, not a hard error. An out-of-range line yields `None`, which becomes one unverified breakpoint, rather than failing the whole `setBreakpoints` request and taking the valid breakpoints down with it. There's a test asserting that `[{-1}, {65546}, {10}]` leaves line 10 verified while the other two come back unverified. If you'd rather a malformed line be a protocol error, that's a one-line change to return `Err` from the deserializer.

---

PR #1392, opened 2026-08-21
document debugger support on the docs website

garretfick:
Proposes the page set, quadrant mapping, phasing, and open questions for
adding debugger documentation to the docs website. The debugger shipped
across #1364, #1385, and #1388, but the website says nothing about it and
the landing page states the opposite.

Claude-Session: https://claude.ai/code/session_01C2DMiwE8kfNB92xofG3N2r


docs/how-to-guides/getting-started/debug-a-program.rst
```diff
@@ -0,0 +1,194 @@
...
+
+Breakpoints still work in your source files. The container records which
+file each line came from, so the debugger matches your breakpoints to the
+right file.
+
+Recompile after every source change. The debugger runs the container as it
+finds it.
+
+--------------------------------------
+Debug a TwinCAT POU
```
garretfick: Can we omit this section about TwinCAT. It is true, but it belongs in the TwinCAT section, not here. Let's not write the TwinCAT section today.

docs/quickstart/debugging.rst
```diff
@@ -0,0 +1,240 @@
...
+.. code-block:: text
+
+   scanCount  ULINT   0
+
+``scanCount`` is the number of scan cycles the program has completed. It is
+0, because none have. This is the number to watch.
+
+.. note::
+
+   ``PulseTimer`` is a function block instance. Its fields --- ``IN``,
```
garretfick: Omit this note. We said debugging is early.

docs/quickstart/debugging.rst
```diff
@@ -0,0 +1,240 @@
...
+``PrevBuzzer`` holds what ``Buzzer`` was on the previous scan. The count now
+changes only on the scan where the buzzer turns on --- the **rising edge**.
+
+Press :kbd:`F5`, then :kbd:`F5` again to leave the entry pause. The program
+stops on the counter line once, at ``scanCount`` 500, exactly as before.
+
+Press :kbd:`F5` again. Nothing stops. The program keeps scanning, the
+condition is never true again, and ``RingCount`` stays where it is.
+
+--------------------------------------
```
garretfick: Other examples don't have What You Learned. Omit this for now.

docs/quickstart/debugging.rst
```diff
@@ -0,0 +1,240 @@
...
+
+The debugger makes that visible. A breakpoint stops the program on every
+scan, and ``scanCount`` tells you which scan you are on, so you can watch a
+value accumulate across cycles instead of guessing why it is wrong.
+
+--------------------------------------
+Next Steps
+--------------------------------------
+
+* :doc:`/explanation/debugging-a-scan-cycle` --- what pausing means when the
```
garretfick: this points to the next step in the tutorial.

docs/reference/editor/debugging.rst
```diff
@@ -0,0 +1,312 @@
...
+Debugging
+=========
+
+.. include:: /includes/debugging-in-development.rst
+
+The IronPLC extension debugs Structured Text programs: set breakpoints on
+source lines, step through the code, and inspect variables while the program
+is paused.
+
+Debugging is driven by :program:`ironplcdap`, which installs alongside the
```
garretfick, line 11: I think we should rename the program to `ironplcvmd`

garretfick: Agreed, but doing it here would turn a docs-only PR into a compiler + extension + installer change. `ironplcdap` appears 135 times across 32 files: the `[[bin]]` target, the shipped-binaries guard test, the NSIS and Homebrew installers, the extension's discovery code and `package.json`, the `IRONPLCDAP` environment variable, the `E0007` problem text, and the design and plan docs.

Plan: keep `ironplcdap` in this PR, then do the rename as a follow-up once this merges — code, installers, tests, and these pages in one commit, so the docs are never wrong. The `ironplc.dapServerPath` setting and the `IRONPLCDAP` environment variable get renamed with it, with no compatibility fallback.

Leaving this thread open to track the follow-up.

---

docs/quickstart/debugging.rst
```diff
@@ -0,0 +1,219 @@
...
+:guilabel:`Program` holds the variables you declared:
+
+.. code-block:: text
+
+   Button     BOOL    FALSE
+   Buzzer     BOOL    FALSE
+   RingCount  DINT    0
+   PulseTimer TON     0
+
+:guilabel:`Runtime` holds one value the virtual machine owns:
```
garretfick: Let's not say "one value". It might hold other values in the future. Include this, but write this as an example, put `...` to indicate that there can be additional values.

docs/quickstart/debugging.rst
```diff
@@ -0,0 +1,219 @@
...
+.. code-block:: text
+
+   Buzzer     BOOL    TRUE
+   RingCount  DINT    0
+   scanCount  ULINT   500
+
+Two things to notice.
+
+``scanCount`` is 500. The program ran 500 complete scan cycles before the
+buzzer turned on, because ``PulseTimer`` waits ``T#500ms`` and the debugger
```
garretfick: Why does the debugger advance 1 ms per scan? It should advance based on the program configuration.

docs/reference/compiler/ironplcc.rst
```diff
@@ -52,6 +52,11 @@ Build Commands
       generation does not yet support produces a code generation error
       rather than incorrect bytecode.
 
+   Every container carries debug information --- source line maps, variable
```
garretfick: This will not be true forever. We want to enable compiling without this information. Let's not make this claim now because it will be false in the future.

docs/reference/editor/debugging.rst
```diff
@@ -0,0 +1,316 @@
...
+
+Before You Start
+================
+
+Debugging needs three things:
+
+* **The IronPLC compiler.** The extension discovers :program:`ironplcc` and
+  finds :program:`ironplcdap` beside it. See :doc:`problems/E0007` if the
+  server cannot be found.
+* **A program that compiles on its own.** The debugger compiles the single
```
garretfick: I think we should omit this middle "program that compiles on its own". That's saying we only support one program instance, and that's covered below.

docs/reference/editor/debugging.rst
```diff
@@ -0,0 +1,316 @@
...
+  file you point it at. A file that references POUs or types defined in
+  sibling files does not compile in isolation --- compile the project first
+  and debug the container instead. See `Debug a Compiled Container`_.
+* **A single program instance.** The configuration must declare exactly one
```
garretfick, line 23: The reason doesn't matter. We would eventually like to support more than one program.

docs/reference/editor/debugging.rst
```diff
@@ -0,0 +1,316 @@
...
+   * - :guilabel:`Step Out`
+     - :kbd:`Shift+F11`
+     - Run to the end of the current POU and stop in its caller.
+   * - :guilabel:`Stop`
+     - :kbd:`Shift+F5`
+     - End the session.
+
+.. note::
+
+   :guilabel:`Pause` and :guilabel:`Restart` are not supported. The debug
```
garretfick: We have a message about early support. Should we just omit this?

docs/reference/editor/debugging.rst
```diff
@@ -0,0 +1,316 @@
...
+       you continue, which is how you tell one scan from the next.
+
+.. note::
+
+   Values are read-only. Setting a variable while paused, forcing a value, and
+   watch expressions are not supported, and expressions typed into the Debug
+   Console are not evaluated. To change what the program does, edit the source
+   and launch again.
+
+   Forcing is absent rather than merely missing. In a PLC, forcing a variable
```
garretfick: Can we omit this paragraph about "forcing is absent"

docs/reference/editor/debugging.rst
```diff
@@ -0,0 +1,316 @@
...
+   means holding it at a value across scans, overriding what the program
+   writes. A value set only while paused is overwritten on the next scan, a
+   few milliseconds later, which looks like a debugger that does not work.
+
+.. note::
+
+   Two kinds of value do not render yet. A ``WSTRING`` shows
```
garretfick: Omit this as future work.

docs/reference/editor/debugging.rst
```diff
@@ -0,0 +1,316 @@
...
+
+Program Time
+============
+
+The debugger drives the clock itself. Program time advances by one
+millisecond for every scan cycle, rather than following the real clock, so a
+``TON`` declared ``PT := T#500ms`` elapses on scan 500 no matter how long you
+spend paused.
+
+This makes a debug session repeatable: the same program stops at the same
```
garretfick: Something is very wrong here. We should run one cycle per the program's configuration.

garretfick (commented):
see comments

docs/quickstart/debugging.rst
garretfick: You are right that it should follow the configuration, and it does not. This is what the debug server does today, not a docs choice.

`dap/server.rs` drives the VM itself rather than through the scheduler:

```rust
running.run_round_debug(current_time_us, &mut hook)
...
current_time_us = current_time_us.saturating_add(1000);
```

A flat 1000 µs per round, and `run_round_debug` bypasses the scheduler, so `INTERVAL := T#100ms` has no effect under the debugger — every round executes the program body and the clock moves 1 ms.

Measured it against the tutorial program (`TASK plc_task(INTERVAL := T#100ms)`, `PulseTimer` with `PT := T#500ms`): the timer elapsed at `scanCount` 500. Honoring the task interval it would be scan 5.

Removed the claim from the reference and the debug server page, and rewrote the tutorial to describe how the values move rather than naming scan numbers, so it stays correct once the clock follows the configuration. Same for the `Program Time` section you flagged separately.

Leaving this thread open — the fix belongs in the debug server, not here.

---

docs/reference/editor/debugging.rst
garretfick: Tracked in #1399, with the decisions from this thread recorded there: the setting and environment variable are renamed with the binary, no compatibility fallback, and the docs pages change in the same commit. Blocked on this PR merging first.

Resolving so this does not hold up the merge.

---

docs/quickstart/debugging.rst
garretfick: Tracked in #1397, with the code pointers and the measured numbers. The docs on this branch make no claim about how time advances, so they stay correct whichever way it is fixed.

Resolving so this does not hold up the merge.

---

PR #1401, opened 2026-08-22
Rename ironplcdap to ironplcvmd

garretfick:
Records the rename and the deduplication that goes with it: one constant
per name in the extension, and guards for the copies that live in a
manifest, a CSV, or a Cargo target and so cannot import one.

Claude-Session: https://claude.ai/code/session_01EKmSJBfmtx77QhqwLVtW3U

### DIFF
--- a/compiler/test/tests/shipped_binaries_guard.rs
+++ b/compiler/test/tests/shipped_binaries_guard.rs
@@ -1,9 +1,10 @@
 //! Workspace-level guard that every binary we build is a binary we ship.
 //!
-//! `ironplcdap` was declared with `required-features = ["dap"]` while every
-//! packaging manifest listed only the other three binaries. Nothing compared the
-//! two, so the debug adapter was neither built by `cargo build` nor shipped by
-//! any installer, and the failure surfaced only as "the debugger doesn't work".
+//! `ironplcvmd` (then named `ironplcdap`) was declared with `required-features
+//! = ["dap"]` while every packaging manifest listed only the other three
+//! binaries. Nothing compared the two, so the debug adapter was neither built
+//! by `cargo build` nor shipped by any installer, and the failure surfaced only
+//! as "the debugger doesn't work".
 //! See `specs/plans/2026-08-16-always-build-ship-dap-server.md`.
 //!
 //! Like `spec_conformance_guard.rs`, both sides are recovered from files already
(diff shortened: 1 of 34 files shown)


compiler/test/tests/shipped_binaries_guard.rs
```diff
@@ -1,9 +1,10 @@
 //! Workspace-level guard that every binary we build is a binary we ship.
 //!
-//! `ironplcdap` was declared with `required-features = ["dap"]` while every
-//! packaging manifest listed only the other three binaries. Nothing compared the
-//! two, so the debug adapter was neither built by `cargo build` nor shipped by
-//! any installer, and the failure surfaced only as "the debugger doesn't work".
+//! `ironplcvmd` (then named `ironplcdap`) was declared with `required-features
```
garretfick: Can we remove the historical information. I didn't want that in code.

compiler/vm-cli/Cargo.toml
```diff
@@ -11,15 +11,19 @@ repository.workspace = true
...
 path = "src/main.rs"
 
-# The Debug Adapter Protocol server. Built unconditionally, like every other
-# shipped binary. It was once gated behind a `dap` feature to keep the DAP layer
-# out of `ironplcvm`, but that gate bought nothing: `mod dap` is declared only in
-# `dap_main.rs`, and separate `[[bin]]` targets are separate compilation units,
-# so `ironplcvm` never compiled the DAP layer either way. What the gate did do
-# was make `cargo build` skip this binary, leaving whatever stale copy was last
-# built by hand in `target/` for editors to launch.
+# The debug daemon: `ironplcvm` under a debugger. It is named for what it is
```
garretfick: There is some historical information here referring to an old name. Just document the behavior, not history

specs/design/debugger-support.md
```diff
@@ -94,9 +94,11 @@ The debugger is four layers:
 
 ### Why a separate DAP binary?
 
-The DAP server is its own binary, `ironplcdap`, built from `dap_main.rs` in the `vm-cli` crate. An earlier draft made it an `ironplcvm debug` subcommand to avoid duplicating the VM embedding code; sharing the crate achieves that without coupling the two entry points, and keeps the DAP protocol loop out of the production VM's argument parsing and lifecycle.
+The DAP server is its own binary, `ironplcvmd`, built from `dap_main.rs` in the `vm-cli` crate. An earlier draft made it an `ironplcvm debug` subcommand to avoid duplicating the VM embedding code; sharing the crate achieves that without coupling the two entry points, and keeps the DAP protocol loop out of the production VM's argument parsing and lifecycle.
```
garretfick: Document what we have, not how we got there. Don't put history here. History can be in plans, not in final code.

compiler/test/tests/shipped_binaries_guard.rs
garretfick: Removed. The header now states the failure mode the guard covers — a binary that is built but not packaged fails no compiler test, so it surfaces only as a missing program in an installed IronPLC — instead of narrating the bug that prompted it.

---

compiler/vm-cli/Cargo.toml
garretfick: Removed. The old name there was the `dap` feature, which no longer exists — the comment explained a gate that was already gone. It is now present tense and states only what holds today: the target is built unconditionally, and `ironplcvm` does not compile the DAP layer because separate `[[bin]]` targets are separate compilation units.

I made the same edit in `src/dap/mod.rs`, which carried a naming rationale for the same reason.

---

specs/design/debugger-support.md
garretfick: Removed both bits of history in that section: the note I had added about the rename, and the pre-existing sentence describing a rejected `ironplcvm debug` subcommand. The section now answers its own question directly — sharing the crate reuses the VM embedding code without coupling the two entry points.

One I left alone, since it is outside this change and a larger rewrite: the Phase 5 section (~line 1360) has a **Packaging decision (revised 2026-08-16)** paragraph and a "This supersedes the original decision to feature-gate DAP" paragraph. Same principle applies — happy to clean those up here or in a follow-up, whichever you prefer.

---

PR #1403, opened 2026-08-22
THIS^ and SUPER^ parsing (ADR-0041 Phase 1)

garretfick:
## Summary

This is an implementation plan for parsing `THIS^` and `SUPER^` — the self-reference and base-reference forms used by CODESYS/TwinCAT OOP — behind the existing `--allow-fb-inheritance` flag. The plan covers front-end parsing and round-trip rendering only, with no code generation semantics. Programs using these constructs will be reported as not-yet-implemented (P9999) rather than syntax errors.

This is Phase 1 of ADR-0041 and item 2 of Phase 2 in the Beckhoff TwinCAT dialect design document.

## Key Changes

The plan specifies:

- **Tokens**: Add `THIS` and `SUPER` keyword tokens, demoted to identifiers when `allow_fb_inheritance` is off (same gating as `METHOD`, `EXTENDS`, etc.)

- **AST**: Introduce `SelfRefKind` enum and `SelfRefVariable` struct as a new head variant of `SymbolicVariableKind`, allowing `THIS^` and `SUPER^` to appear in both read and write positions. The caret is folded into the node since these are not pointers in IronPLC's type system.

- **Parser**: Add `self_ref()` rule accepting `THIS _^ ` and `SUPER _^` (with optional whitespace before the caret), integrate as head alternative in `symbolic_variable()`, and add receiver alternative in `method_invocation()` for statement-position method calls.

- **Method calls**: Introduce `MethodReceiver` enum to distinguish `Instance(Id)` from `SelfRef(SelfRefVariable)` receivers, narrowing the change to avoid silently accepting `p^.M()` or `a.b.M()`.

- **Analyzer**: Implement `LanguageExtension` for `SelfRefVariable` and add explicit not-implemented arms (never silent skips) at all match sites. This forces the compiler to enumerate the work and ensures future semantics additions don't silently miscompile.

- **Codegen**: Add `Diagnostic::todo` backstop arms in variable lowering and statement compilation.

- **plc2plc**: Render `THIS^` / `SUPER^` with `write("^")` (no space) and support round-trip parsing.

- **Testing**: Parser tests (flag on/off, whitespace cases), plc2plc round-trip tests, analyzer rejection + P9999 presence assertion, and compile-fails-with-P9999 test.

## Notable Implementation Details

- **Whitespace before caret is accepted** (`THIS ^` is legal) to match vendor behavior and avoid rejecting real code, while whitespace after the caret keeps today's behavior (`THIS^ . x` still fails).

- **The caret is mandatory** — `THIS.count` and bare `THIS` are not accepted, matching CODESYS/TwinCAT semantics where these are pointers.

- **No quiet analyzer arms** — every new match site must return an explicit diagnostic, never a silent skip. This prevents future semantic additions from silently miscompiling if an earlier pass was quietly ignoring the construct.

- **Redundant diagnostics are expected** — multiple passes will report the construct as unimplemented before `rule_unsupported_extension` runs. This is the accepted cost of ensuring no pass silently skips the construct.

- **Optional add-ons** identified for future work: a dedicated diagnostic for missing caret, `SUPER^` outside an extending FB, and fixing an unrelated `ExprKind::Deref` rendering bug.

## Files Modified

The plan details changes to ~20 files across parser, DSL, analyzer, codegen, plc2plc, and documentation, with a comprehensive task checklist for implementation.

https://claude.ai/code/session_01GGyTDF3y1QSec861eiJsj4

### DIFF
--- a/compiler/analyzer/src/rule_unsupported_extension.rs
+++ b/compiler/analyzer/src/rule_unsupported_extension.rs
@@ -97,8 +97,7 @@ impl Visitor<Diagnostic> for RuleUnsupportedExtension {
     ) -> Result<Self::Value, Diagnostic> {
         // A SelfRefVariable only exists when THIS^/SUPER^ was written, so
         // it is always an extension. Parsed and rendered, but neither
-        // analyzed nor executed -- see
-        // specs/plans/2026-08-22-oop-this-super-parsing.md.
+        // analyzed nor executed.
         self.flag(node);
         node.recurse_visit(self)
     }
--- a/compiler/analyzer/src/rule_unsupported_extension.rs
+++ b/compiler/analyzer/src/rule_unsupported_extension.rs
@@ -91,6 +91,18 @@ impl Visitor<Diagnostic> for RuleUnsupportedExtension {
         node.recurse_visit(self)
     }
 
+    fn visit_self_ref_variable(
+        &mut self,
+        node: &ironplc_dsl::textual::SelfRefVariable,
+    ) -> Result<Self::Value, Diagnostic> {
+        // A SelfRefVariable only exists when THIS^/SUPER^ was written, so
+        // it is always an extension. Parsed and rendered, but neither
+        // analyzed nor executed -- see
+        // specs/plans/2026-08-22-oop-this-super-parsing.md.
+        self.flag(node);
+        node.recurse_visit(self)
+    }
+
     fn visit_interface_declaration(
         &mut self,
         node: &InterfaceDeclaration,
@@ -157,6 +169,36 @@ END_FUNCTION_BLOCK";
 
         let (input, _context) =
             parse_and_resolve_types_with_options(program, &opts_with_fb_inheritance());
+        let context = SemanticContextBuilder::new().build().unwrap();
+        let result = apply(&input, &context, &opts_with_fb_inheritance());
+
+        let errors = result.unwrap_err();
+        assert_eq!(errors.len(), 1);
+        // P9999 == Problem::NotImplemented; the enum variant is #[deprecated]
+        // (must be constructed via Diagnostic::not_implemented), so assert on
+        // the stable code string rather than referencing the variant.
+        assert_eq!("P9999", errors[0].code);
+    }
+
+    #[rstest::rstest]
+    #[case::this("    THIS^.count := 1;")]
+    #[case::super_("    count := SUPER^.count;")]
+    #[case::this_method_call("    THIS^.Start();")]
+    fn apply_when_self_ref_then_p9999(#[case] body: &str) {
+        let program = format!(
+            "
+FUNCTION_BLOCK FB_Motor
+VAR
+    count : INT;
+END_VAR
+METHOD Run
+{body}
+END_METHOD
+END_FUNCTION_BLOCK"
+        );
+
+        let (input, _context) =
+            parse_and_resolve_types_with_options(&program, &opts_with_fb_inheritance());
         let context = SemanticContextBuilder::new().build().unwrap();
         let result = apply(&input, &context, &opts_with_fb_inheritance());
 
--- a/compiler/analyzer/src/stages.rs
+++ b/compiler/analyzer/src/stages.rs
@@ -618,4 +618,80 @@ END_FUNCTION_BLOCK";
             "unexpected spurious P2011: {codes:?}"
         );
     }
+
+    // ---------------------------------------------------------------------
+    // THIS^ / SUPER^ (parsed, not analyzed or executed).
+    // See specs/plans/2026-08-22-oop-this-super-parsing.md.
+    // ---------------------------------------------------------------------
+
+    /// A program using `THIS^` is rejected, and P9999 is among the reasons.
+    ///
+    /// Deliberately asserts presence rather than an exact diagnostic set:
+    /// several passes meet the construct and each says so, and pinning the
+    /// set would turn every later improvement into a test edit. What must
+    /// hold is that no pass quietly accepts it.
+    #[rstest::rstest]
+    #[case::this_field_write("    THIS^.count := 1;")]
+    #[case::super_field_read("    count := SUPER^.count;")]
+    #[case::this_method_call("    THIS^.Start();")]
+    fn analyze_when_self_ref_then_rejected_with_not_implemented(#[case] body: &str) {
+        let options = CompilerOptions {
+            allow_fb_inheritance: true,
+            ..CompilerOptions::default()
+        };
+        let program = format!(
+            "
+FUNCTION_BLOCK FB_Motor
+VAR
+    count : INT;
+END_VAR
+METHOD Start
+    count := 1;
+END_METHOD
+METHOD Run
+{body}
+END_METHOD
+END_FUNCTION_BLOCK"
+        );
+        let lib = parse_program(&program, &FileId::default(), &options).unwrap();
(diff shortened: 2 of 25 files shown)


specs/plans/2026-08-22-oop-this-super-parsing.md
```diff
@@ -0,0 +1,420 @@
...
+Two related decisions that follow from the same reasoning:
+
+- **The `^` stays mandatory.** `THIS.count` and a bare `THIS` are not
+  accepted. That matches CODESYS/TwinCAT, where these are pointers.
+- **Whitespace *after* the caret keeps today's behavior** — `THIS^.x`
+  parses, `THIS^ . x` does not, exactly as `p^.x` / `p^ . x` behave now.
+  Relaxing member access generally is a separate change; this slice
+  should not make `THIS^` more permissive than `p^` past the caret.
+
+Also worth fixing near this work (separately — it is a pre-existing bug,
```
garretfick: Create an issue for this, then remove this from the plan.

garretfick (commented):
See comments

specs/plans/2026-08-22-oop-this-super-parsing.md
garretfick: Filed as #1404 (cause, the measured whitespace table, why the existing test misses it, and the suggested fix), and removed from the plan — both here and the "Recommended as its own commit" entry under Optional add-ons. What's left in its place is a one-sentence pointer to the issue, since the whitespace discussion above it is what surfaced the bug.

---

compiler/parser/src/tests/this_super.rs
```diff
@@ -0,0 +1,200 @@
...
+END_VAR
+
+THIS := 1;
+SUPER := 2;
+END_FUNCTION_BLOCK
+";
+    let result = parse_program(program, &FileId::default(), &CompilerOptions::default());
+    assert!(
+        result.is_ok(),
+        "THIS/SUPER must remain valid identifiers in standard mode: {:?}",
```
garretfick: We don't need the extra part for the message and err on the assert. That's basically dead code that never executes. Just write the first part. Same in multiple places here.

compiler/parser/src/parser.rs
```diff
@@ -849,9 +849,27 @@ parser! {
...
-      // Start by assuming that the top is just a named variable
-      let mut head = SymbolicVariableKind::Named(NamedVariable { name });
+    // OOP extension: `THIS^` / `SUPER^` (ADR-0041 Phase 1). The caret is
```
garretfick: We don't need to mention the ADR here.

compiler/parser/src/parser.rs
```diff
@@ -849,9 +849,27 @@ parser! {
...
+    // mandatory -- `THIS` and `SUPER` are pointers to an instance in the
+    // dialects that define them -- and optional whitespace/comments are
+    // accepted between the keyword and the caret, since nothing in the
+    // grammar joins them into one token. The `This`/`Super` tokens only
+    // reach the parser when `allow_fb_inheritance` is set (they demote to
+    // identifiers otherwise), so this rule is gated at the token level.
+    // See specs/plans/2026-08-22-oop-this-super-parsing.md.
```
garretfick: Definitely don't mention the plans here.

garretfick (commented):
See comments inline

compiler/parser/src/tests/this_super.rs
garretfick: Done — bare asserts everywhere in the new tests:

- this file: the standard-mode `is_ok` assert and the no-caret `is_err` assert
- `parse_in_method` no longer asserts `is_ok` before `unwrap()`; it just unwraps
- `plc2plc/src/tests/this_super.rs`: the re-parse is a plain `.unwrap()` again
- `codegen/tests/it/compile_this_super.rs`: the `is_err` assert

Kept one message: the `rendered.contains(...)` assert in the plc2plc test, where the message prints the rendered output — that isn't recoverable from the assert expression, and it matches `reference_to.rs`. Say the word if you want that one bare too.

---

compiler/parser/src/parser.rs
garretfick: Removed. The comment now just describes the rule: the caret is mandatory because THIS/SUPER are pointers, whitespace before it is accepted, and the tokens only reach the parser under `allow_fb_inheritance`.

---

compiler/parser/src/parser.rs
garretfick: Removed. Also took the plan reference out of the `THIS`/`SUPER` token comment in `token.rs` and pointed it at `xform_demote_keywords.rs` instead, matching the comment on the OOP keywords right above it.

I left plan references in place where the immediate neighbours already have them — the `dsl/textual.rs` node doc (the `MethodCall` doc two structs down cites its plan) and the new test-file headers (`parser/src/tests/methods.rs` and friends do the same). Happy to sweep those too if you'd rather the convention go the other way.

---

compiler/analyzer/src/rule_unsupported_extension.rs
```diff
@@ -91,6 +91,18 @@ impl Visitor<Diagnostic> for RuleUnsupportedExtension {
...
     }
 
+    fn visit_self_ref_variable(
+        &mut self,
+        node: &ironplc_dsl::textual::SelfRefVariable,
+    ) -> Result<Self::Value, Diagnostic> {
+        // A SelfRefVariable only exists when THIS^/SUPER^ was written, so
+        // it is always an extension. Parsed and rendered, but neither
+        // analyzed nor executed -- see
+        // specs/plans/2026-08-22-oop-this-super-parsing.md.
```
garretfick: Remove the plan reference.

compiler/analyzer/src/stages.rs
```diff
@@ -618,4 +618,80 @@ END_FUNCTION_BLOCK";
             "unexpected spurious P2011: {codes:?}"
         );
     }
+
+    // ---------------------------------------------------------------------
+    // THIS^ / SUPER^ (parsed, not analyzed or executed).
+    // See specs/plans/2026-08-22-oop-this-super-parsing.md.
```
garretfick: Remove reference to plans

garretfick (commented):
I see multiple places referring to plans. Please remove those

compiler/analyzer/src/rule_unsupported_extension.rs
garretfick: Removed. Took your answer to my earlier question as the general rule and swept every plan reference out of this diff, not just the two flagged:

- `analyzer/src/rule_unsupported_extension.rs` (here)
- `analyzer/src/stages.rs` test-section comment
- `dsl/src/textual.rs` — the `SelfRefVariable` doc, both the plan reference and the ADR mention
- the three new test-file headers (`parser`, `plc2plc`, `codegen`)

`grep specs/plans` over the diff is now empty. Pre-existing ADR-0041 references elsewhere are untouched — including one in a `renderer.rs` comment I extended, which shows as a changed line. Happy to sweep those repo-wide as a separate change if you want them gone too.

---

compiler/analyzer/src/stages.rs
garretfick: Removed — the test-section header is just `THIS^ / SUPER^ (parsed, not analyzed or executed)` now. Swept the rest of the diff for plan references at the same time; details on the `rule_unsupported_extension.rs` thread.

---

Issue #1370, opened 2026-08-13
Finish adopting shared test fixtures in mcp, and make them shareable (Stage C2)


garretfick:
Stage C2 of [the cross-crate test-duplication plan](specs/plans/2026-08-06-reduce-cross-crate-test-duplication.md). Test-infrastructure only — no production code changes.

## Current state (re-measured on main, after Stage A)

The original survey found the same program fixtures inlined as string literals ~17× in `mcp`, plus copies in `project` and `playground`. Stage A (#1339) added `compiler/mcp/src/tools/test_support.rs` with the canonical fixtures and converted the tool modules it touched, so **the cross-crate half of this is already done** — `playground/src` and `project/src` now contain zero copies.

What remains is uneven adoption *within* `mcp`. Counting `"PROGRAM p\nEND_PROGRAM"` on current main:

| File | Occurrences |
|---|---|
| `compiler/mcp/src/spec_conformance.rs` | 11 |
| `compiler/mcp/tests/cli.rs` | 7 |
| `compiler/mcp/src/feature_flag_conformance.rs` | 7 |
| `compiler/mcp/src/tools/types_all.rs` | 6 |
| `compiler/mcp/src/tools/parse.rs` | 4 |
| `compiler/mcp/src/tools/symbols.rs` | 3 |
| `compiler/mcp/src/tools/project_manifest.rs` | 3 |
| `compiler/mcp/tests/scenarios.rs` | 1 |
| `compiler/mcp/src/tools/pou_scope.rs` | 1 |
| `compiler/mcp/src/tools/pou_lineage.rs` | 1 |

37 total, of which 1 is the canonical definition in `test_support.rs`. The semantic-error snippet (`x := y;`) appears 13× in `mcp/src`.

Existing fixtures in `compiler/mcp/src/tools/test_support.rs`: `VALID_PROGRAM`, `SYNTAX_ERROR_PROGRAM`, `SEMANTIC_ERROR_PROGRAM`, `ed2_options()`, `source()`, `unnamed_source()`.

## Work

1. **Adopt `test_support` across the remaining `mcp` modules** listed above, extending it with any fixture shape that recurs (a program with a declared var, an enum type, etc.) rather than adding one-off constants per file.
[...]

PR #1408, opened 2026-08-22
Fix dereference operator spacing to enable round-trip parsing

garretfick:
## Summary

Fixes the renderer to emit dereference operators (`^`) without a separating space from their operand. The parser's `unary_expression` rule does not allow whitespace between the operand and `^`, so the previous output (`myRef ^`) would fail to re-parse.

## Changes

- **renderer.rs**: Changed `self.write_ws("^")` to `self.write("^")` in the `Deref` expression handler, with a comment explaining the parser constraint
- **Test expectations**: Updated all reference test files to expect `myRef^` instead of `myRef ^`:
  - `ref_rendered.st`
  - `reference_to_rendered.st`
  - `adr_rendered.st`
  - `pointer_to_rendered.st`
- **New round-trip test**: Added `write_to_string_when_deref_expression_then_round_trips()` in `reference_to.rs` that verifies rendered dereference expressions can be re-parsed successfully, preventing future regressions

## Implementation Details

The fix ensures that the plc2plc renderer produces syntactically valid output that can be re-parsed. The new test explicitly validates round-trip parsing (render → parse → compare ASTs) for dereference expressions, catching any future spacing regressions at the parser boundary.

https://claude.ai/code/session_01PmzDHbWkF8vqJ9uXqxLE1s

### DIFF
--- a/compiler/plc2plc/src/tests/reference_to.rs
+++ b/compiler/plc2plc/src/tests/reference_to.rs
@@ -80,7 +80,7 @@ VAR
 END_VAR
     value := myRef^;
 END_PROGRAM",
-    "myRef ^"
+    "myRef^"
 )]
 #[case::ref_expression(
     "PROGRAM main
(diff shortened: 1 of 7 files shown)


compiler/plc2plc/src/tests/reference_to.rs
```diff
@@ -138,3 +138,30 @@ fn write_to_string_when_ref_to_type_decl_then_preserves() {
...
+#[test]
+fn write_to_string_when_deref_expression_then_round_trips() {
+    let source = "PROGRAM main
+VAR
+    myRef : REF_TO INT;
+    value : INT;
+END_VAR
+    value := myRef^;
+END_PROGRAM";
+    let options = CompilerOptions::from_dialect(Dialect::Iec61131_3Ed3);
```
garretfick: Can we not remove some of this boiler plate code?

garretfick: Done in 1bad75d. Extracted `assert_round_trips(source, &options)` into `tests/common.rs`; the test body is now one call instead of six lines of setup.

The helper also panics with the rendered text when the re-parse fails, which the hand-rolled copies don't — a re-parse failure now shows the offending output rather than just a diagnostic.

Worth noting the scope: this same parse → render → re-parse block is hand-rolled at **20 call sites across 14 files** in `plc2plc/src/tests/`, so the helper is reusable well beyond this test. I deliberately didn't convert the other 19 here — this is a 6-file bugfix PR and that sweep would dominate the diff. Two of the existing sites also re-parse with *different* (stricter) options than they parsed with, deliberately, so they'd need an `assert_round_trips_with(source, parse_opts, reparse_opts)` variant rather than a mechanical swap.

Happy to do that sweep in this PR if you'd prefer it here, or file it as a follow-up alongside #1370/#1371 — your call.

---

Issue #1383, opened 2026-08-16
Top-level ARRAY OF &lt;struct&gt; variables are rejected at declaration time


garretfick:
## Summary

An array variable whose element type is a user-defined structure fails in codegen at *declaration* time, before any access is compiled:

```st
TYPE Item : STRUCT
    Flag : BOOL;
END_STRUCT;
END_TYPE

PROGRAM main
VAR
    Arr : ARRAY[1..6] OF Item;
END_VAR
    Arr[1].Flag := TRUE;
END_PROGRAM
```

```
error[P9999]: Capability is not implemented (yet!)
  │     Arr : ARRAY[1..6] OF Item;
  │     ^^^ Unsupported array element type
  = ... file=codegen/src/compile_array.rs&line=520
```

`ironplcc check` accepts the program; only codegen rejects it.

## Cause

Array registration resolves its element type through `compile_setup::resolve_type_name`, which maps only elementary and generic type names (`INT`, `BOOL`, `REAL`, …) to a `VarTypeInfo`. A structure type name yields `None`, producing the diagnostic above:

```rust
super::compile_setup::resolve_type_name(&spec.element_type_name).ok_or_else(|| {
    Diagnostic::not_implemented(Label::span(span.clone(), "Unsupported array element type"))
})?
```

Because the failure is at declaration, both reads and writes are unreachable.

## Notes

The analyzer already models this correctly — `IntermediateType::slot_count()` computes array-of-struct layout as `total_elements * element_slots` — so this is a codegen-side gap only.

The equivalent access *through a struct field* (`holder.items[i].field`) is supported: structures are laid out as a contiguous run of slots, and the array-of-struct case reuses the struct's flat-slot descriptor with strides scaled by the element slot count. A top-level array-of-struct has no enclosing struct to borrow a descriptor and data region from, so it needs its own allocation path: register the variable with a slot-typed descriptor sized `total_elements * element_slots`, then resolve `Arr[i].field` against it.

Not required by #1376, which reaches its array-of-struct only through struct fields.

Relevant code:
- `compiler/codegen/src/compile_array.rs` — element type resolution during registration
- `compiler/codegen/src/compile_setup.rs` — `resolve_type_name`

PR #1415, opened 2026-08-22
Support top-level ARRAY OF <struct> variables (issue #1383)

garretfick:
## Summary

Implements support for declaring and accessing top-level `ARRAY OF <struct>` variables, allowing code like `Arr[i].field` where `Arr` is an array whose elements are user-defined structures. Previously, `ironplcc check` accepted such declarations but codegen rejected them at declaration time.

## Key Changes

- **New module `compile_array_struct.rs`**: Handles declaration detection, registration, and field access resolution for arrays of structures. Structures occupy contiguous slot runs, so an array of them is a flat slot array rather than one slot per element.

- **Registration path**: Top-level array-of-struct variables are allocated like structure variables—one contiguous data region run of `total_elements * element_slots` slots, plus a slot-typed array descriptor covering the entire region. The variable slot holds the data-region byte offset.

- **Access path reuse**: Field access (`arr[i].field`) reuses the existing `ResolvedAccess::StructFieldArrayElement` variant by scaling array dimension strides by the element slot count, making the emitted flat index count slots directly. No new opcodes or emission paths required.

- **Moved code**: The array-of-struct element field resolution logic previously in `compile_array.rs` is now in the new module, shared between struct-field arrays (`holder.items[i].field`) and top-level arrays (`items[i].field`).

- **Declaration detection**: `struct_array_declaration()` identifies arrays with structure element types (both inline `ARRAY[..] OF Item` and named array types), distinguishing them from `ARRAY OF REF_TO <struct>` which remains on the reference path.

- **Global support**: Functions and function blocks can access global array-of-struct variables through re-inserted global metadata, matching the existing pattern for structure and array globals.

- **Validation**: Rejects unsupported shapes (whole-element access, STRING fields, composite fields) with clear diagnostics at declaration or first access time.

## Implementation Details

- Element initialization is deliberately not emitted, matching existing array-of-struct-field behavior; the data region starts zeroed.
- Bounds checks validate against unscaled element counts, not slot counts.
- Data region byte offsets are stored in variable slots via `LOAD_CONST_I32`, with overflow checks for the 2 GiB limit and 32768-slot maximum.
- Comprehensive end-to-end tests cover nominal reads/writes, bounds shapes (zero-based, negative lower bounds, multidimensional), element shapes (nested structures, 64-bit fields), named array types, global declarations, and rejected shapes.

## Testing

- 20+ end-to-end tests in new `end_to_end_array_of_struct.rs` covering read/write, bounds, element types, globals, and error cases
- Container-shape and bytecode-emission tests in `compile_array.rs`
- All existing tests continue to pass

https://claude.ai/code/session_01Phh6cZqjqR7tSVpe2UUCX2

### DIFF
--- a/compiler/codegen/src/compile_array.rs
+++ b/compiler/codegen/src/compile_array.rs
@@ -411,7 +411,12 @@ pub(crate) fn array_spec_for_declaration(
                 dimensions,
             } = array_type
             else {
-                unreachable!("resolve_array_type guarantees Array variant");
+                // `resolve_array_type` returns only the Array variant, so this
+                // is a compiler invariant rather than anything the program did.
+                return Err(Diagnostic::internal_error_at(Label::span(
+                    type_name.span(),
+                    "Array type resolved to a non-array representation",
+                )));
             };
             array_spec_from_named(element_type, dimensions)
         }
--- a/compiler/codegen/src/compile_array.rs
+++ b/compiler/codegen/src/compile_array.rs
@@ -208,6 +208,17 @@ pub(crate) fn resolve_access<'ctx, 'ast>(
                     SymbolicVariableKind::Structured(structured) => {
                         levels.reverse();
                         let all_subscripts: Vec<&Expr> = levels.into_iter().flatten().collect();
+                        // `a[i].values[j]` -- the record is itself an array
+                        // element, so the subscripts collected here index the
+                        // *field*, and the element index is still inside the
+                        // record. The array-of-struct path combines the two.
+                        if matches!(structured.record.as_ref(), SymbolicVariableKind::Array(_)) {
+                            return crate::compile_array_struct::resolve_struct_array_element_field(
+                                ctx,
+                                structured,
+                                all_subscripts,
+                            );
+                        }
                         return resolve_struct_field_array(ctx, structured, all_subscripts);
                     }
                     other => {
@@ -222,7 +233,11 @@ pub(crate) fn resolve_access<'ctx, 'ast>(
         Variable::Symbolic(SymbolicVariableKind::Structured(structured))
             if matches!(structured.record.as_ref(), SymbolicVariableKind::Array(_)) =>
         {
-            crate::compile_array_struct::resolve_struct_array_element_field(ctx, structured)
+            crate::compile_array_struct::resolve_struct_array_element_field(
+                ctx,
+                structured,
+                Vec::new(),
+            )
         }
         _ => {
             // Fall through to existing resolve_variable() for scalars.
@@ -403,8 +418,14 @@ pub(crate) fn array_spec_for_declaration(
             array_spec_from_inline(subranges, span)
         }
         ironplc_dsl::common::SpecificationKind::Named(type_name) => {
+            // The caller reaches this arm only for a declaration the type
+            // environment already resolved to an array, so a miss here is a
+            // compiler invariant rather than anything the program did.
             let array_type = types.resolve_array_type(type_name).ok_or_else(|| {
-                Diagnostic::not_implemented(Label::span(type_name.span(), "Unknown array type"))
+                Diagnostic::internal_error_at(Label::span(
+                    type_name.span(),
+                    "Array type is absent from the type environment",
+                ))
             })?;
             let IntermediateType::Array {
                 element_type,
--- a/compiler/codegen/src/compile_array.rs
+++ b/compiler/codegen/src/compile_array.rs
@@ -158,10 +158,22 @@ pub(crate) fn resolve_access<'ctx, 'ast>(
                     SymbolicVariableKind::Named(named) => {
                         levels.reverse();
                         let all_subscripts: Vec<&Expr> = levels.into_iter().flatten().collect();
-                        let info = ctx
-                            .array_vars
-                            .get(&named.name)
-                            .ok_or_else(|| Diagnostic::todo_with_span(named.name.span()))?;
+                        let info = ctx.array_vars.get(&named.name).ok_or_else(|| {
+                            // An element of an array of structures spans
+                            // several slots, so it has no single-value load or
+                            // store; only its fields are addressable.
+                            if ctx.struct_array_vars.contains_key(&named.name) {
+                                Diagnostic::not_implemented(Label::span(
+                                    named.name.span(),
+                                    format!(
+                                        "Whole-element access of '{}' -- select a field of the element instead",
+                                        named.name
+                                    ),
+                                ))
+                            } else {
+                                Diagnostic::todo_with_span(named.name.span())
+                            }
+                        })?;
                         return Ok(ResolvedAccess::ArrayElement {
                             info,
                             subscripts: all_subscripts,
@@ -210,7 +222,7 @@ pub(crate) fn resolve_access<'ctx, 'ast>(
         Variable::Symbolic(SymbolicVariableKind::Structured(structured))
             if matches!(structured.record.as_ref(), SymbolicVariableKind::Array(_)) =>
         {
-            resolve_struct_array_element_field(ctx, structured)
+            crate::compile_array_struct::resolve_struct_array_element_field(ctx, structured)
         }
         _ => {
             // Fall through to existing resolve_variable() for scalars.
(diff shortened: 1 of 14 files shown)


compiler/codegen/src/compile_array.rs
```diff
@@ -523,6 +388,36 @@ pub(crate) fn array_spec_from_inline(
...
+        ironplc_dsl::common::SpecificationKind::Named(type_name) => {
+            let array_type = types.resolve_array_type(type_name).ok_or_else(|| {
+                Diagnostic::not_implemented(Label::span(type_name.span(), "Unknown array type"))
+            })?;
+            let IntermediateType::Array {
+                element_type,
+                dimensions,
+            } = array_type
+            else {
+                unreachable!("resolve_array_type guarantees Array variant");
```
garretfick: This should be an internal error diagnostic.

compiler/codegen/src/compile_array_struct.rs
```diff
@@ -0,0 +1,414 @@
...
+    ctx.data_region_offset = ctx
+        .data_region_offset
+        .checked_add(total_bytes)
+        .ok_or_else(|| {
+            Diagnostic::not_implemented(Label::span(span.clone(), "Data region overflow"))
+        })?;
+
+    // The offset is stored in the variable slot via LOAD_CONST_I32.
+    if ctx.data_region_offset > i32::MAX as u32 {
+        return Err(Diagnostic::not_implemented(Label::span(
```
garretfick: This need a new code. Not implemented indicated something we want to implement but haven't gotten to. This is something not currently planned. Can you create a new problem code for this (along with the short hand such as this). It should be in the P9xxx range, at the end, so maybe P9997.

garretfick (commented):
See comments

compiler/codegen/src/compile_array.rs
garretfick: Done in 6f1edf0 — now `Diagnostic::internal_error_at` with the type name's span:

```rust
else {
    // `resolve_array_type` returns only the Array variant, so this
    // is a compiler invariant rather than anything the program did.
    return Err(Diagnostic::internal_error_at(Label::span(
        type_name.span(),
        "Array type resolved to a non-array representation",
    )));
};
```

Worth noting this arm is unreachable today for a second reason: named array specifications are expanded to inline ones before codegen, so `SpecificationKind::Named` never arrives here at all. I confirmed that by panicking in the arm and running the suite — no hit, and the same holds for the equivalent arm I added in `compile_array_struct.rs`. So it is doubly defensive, which is all the more reason not to panic the process over it.

---

compiler/codegen/src/compile_array_struct.rs
garretfick: Added **P9997 / NotSupported** — "Capability is not supported" — in 6f1edf0.

- `compiler/problems/resources/problem-codes.csv`
- `Diagnostic::not_supported(primary: Label)` in `compiler/dsl/src/diagnostic.rs`, mirroring `not_implemented`
- `docs/reference/compiler/problems/P9997.rst`, spelling out the P9999 contrast: P9999 promises "not yet" and a later release should accept the program; P9997 does not, so the program has to change

I also added `NotSupported` to the `compiler_located` list in `problems/build.rs`, so the enum variant is `#[deprecated]` like `NotImplemented` and `InternalError`. That makes the constructor the only way to build one and keeps the compiler `file#Lline` flowing to the P9xxx dashboards — tell me if you would rather it stay constructible directly.

**Scope applied.** I used it for all eight fixed-capacity checks in the new array-of-struct code, not just the flagged line, since they are one class: element-count/slot-count/byte-size overflow, the 32768-slot cap, and the 2 GiB cap. The genuinely "not yet" diagnostics keep P9999 — a STRING leaf, a composite leaf, an initializer on an array of structures, and whole-element assignment.

Test is `compile_when_top_level_array_of_struct_exceeds_slot_limit_then_not_supported`, asserting P9997 on `ARRAY[1..40000] OF Item`; it uses the exact program from the P9997 doc page, and both carry a note to keep them in step.

**One thing left for you to decide.** The pre-existing equivalents still report P9999 — `register_array_variable` in `compile_array.rs` and `allocate_struct_variable` in `compile_struct.rs` have the same "Array too large" / "exceeds maximum 32768 slots" / "Data region overflow" / "2 GiB limit" checks. So today `ARRAY[1..40000] OF Item` gives P9997 while `ARRAY[1..40000] OF INT` gives P9999 for the identical reason. I left them alone rather than widening this PR, but it is an odd seam. Happy to sweep them here, or file a follow-up — your call.

---

compiler/codegen/src/compile_array.rs
```diff
@@ -158,10 +158,22 @@ pub(crate) fn resolve_access<'ctx, 'ast>(
...
-                        let info = ctx
-                            .array_vars
-                            .get(&named.name)
-                            .ok_or_else(|| Diagnostic::todo_with_span(named.name.span()))?;
+                        let info = ctx.array_vars.get(&named.name).ok_or_else(|| {
+                            // An element of an array of structures spans
+                            // several slots, so it has no single-value load or
+                            // store; only its fields are addressable.
+                            if ctx.struct_array_vars.contains_key(&named.name) {
+                                Diagnostic::not_implemented(Label::span(
```
garretfick, line 171: I think this should be caught at analysis time. If that's the case, then this should be an internal error rather than "not implemented".

compiler/codegen/src/compile_array.rs
```diff
@@ -523,6 +388,41 @@ pub(crate) fn array_spec_from_inline(
...
+    spec: &ironplc_dsl::common::SpecificationKind<ironplc_dsl::common::ArraySubranges>,
+    span: &SourceSpan,
+) -> Result<ArraySpec, Diagnostic> {
+    match spec {
+        ironplc_dsl::common::SpecificationKind::Inline(subranges) => {
+            array_spec_from_inline(subranges, span)
+        }
+        ironplc_dsl::common::SpecificationKind::Named(type_name) => {
+            let array_type = types.resolve_array_type(type_name).ok_or_else(|| {
+                Diagnostic::not_implemented(Label::span(type_name.span(), "Unknown array type"))
```
garretfick: This should also be an internal error. We should not get this far if we cannot generte the code.

compiler/codegen/src/compile_array_struct.rs
```diff
@@ -0,0 +1,414 @@
...
+        ArrayOfStructBase::Field(base) => {
+            let (root_name, field_slot_offset, field_type) =
+                crate::compile_struct::walk_struct_chain(ctx, &base.record, &base.field, 0)?;
+
+            let IntermediateType::Array {
+                element_type,
+                dimensions: array_dims,
+            } = &field_type
+            else {
+                return Err(Diagnostic::not_implemented(Label::span(
```
garretfick, line 95: Should this also be internal error?

compiler/codegen/tests/it/end_to_end_array_of_struct.rs
```diff
@@ -0,0 +1,714 @@
...
+//! `specs/plans/2026-08-22-top-level-array-of-struct.md`.
+//!
+//! Field access through an array-of-struct *field* (`h.items[i].a`) is covered
+//! by `end_to_end_struct.rs`; these tests own the case where the variable
+//! itself is the array.
+//!
+//! Assertion indices are variable-table slots, assigned in declaration order
+//! starting at 0. Each test notes its mapping.
+
+use crate::common::try_parse_and_compile;
```
garretfick, line 13: We should have a test that is array of structure containing an array, so it is something like:

`a[i].value[j]`

garretfick (commented):
See comments

compiler/codegen/tests/it/end_to_end_array_of_struct.rs
garretfick: Good catch — `a[i].value[j]` didn't work at all. The subscript chain bottomed out in `walk_struct_chain`, which has no array case, so it failed with a bare P9999. Implemented in f597b93 rather than just tested.

Both indices fit one flat-index computation. The element index is already scaled by the element's slot count; the field's own dimensions append after it at their natural strides, because the field's elements are single slots sitting side by side inside the element. `emit_flat_index` then produces `i * element_slots + j` in one pass, and the field's offset within the element stays the compile-time part — so it still reuses `StructFieldArrayElement`, no new opcode and no new emission path.

Five tests:

| Test | Covers |
|---|---|
| `..._element_has_array_field_then_reads_back` | literal indices, three distinct `(element, field)` pairs |
| `..._array_field_indexed_by_variables_then_correct_element` | both subscripts runtime, not constant-folded |
| `..._array_field_preceded_by_scalar_then_no_alias` | a scalar ahead of the array field, so the field offset must be added — a wrong offset aliases `lead` with `values[1]` |
| `..._array_field_written_in_nested_loops_then_all_set` | nested FOR over both axes |
| `..._struct_field_array_of_struct_has_array_field_then_reads_back` | the same through a struct field, `h.items[i].values[j]` |

Verified on the real VM as well as in tests: `a[2].value[3] := 7` then reading it back gives 7.

---

compiler/codegen/src/compile_array.rs
garretfick: Agreed — changed to `internal_error_at` in f597b93:

```rust
// The caller reaches this arm only for a declaration the type
// environment already resolved to an array, so a miss here is a
// compiler invariant rather than anything the program did.
let array_type = types.resolve_array_type(type_name).ok_or_else(|| {
    Diagnostic::internal_error_at(Label::span(
        type_name.span(),
        "Array type is absent from the type environment",
    ))
})?;
```

Both arms of this `match` are now internal errors, which is consistent: the whole `SpecificationKind::Named` branch is unreachable today anyway, since named array specifications are expanded to inline ones before codegen.

---

compiler/codegen/src/compile_array.rs
garretfick: I checked, and the premise doesn't hold — so I've left this as `not_implemented` rather than making the change. Happy to be overruled.

This branch fires for `arr[1] := one` (assigning a whole element of an array-of-struct). Analysis accepts it:

```
$ ironplcc check p.st     # TYPE Item : STRUCT a : DINT; ... arr[1] := one;
$                         # no output — passes
$ ironplcc compile ...
error[P9999]: Capability is not implemented (yet!)
  │  ^^^ Whole-element access of 'arr' -- select a field of the element instead
```

And it *should* pass analysis: whole-element assignment of a structure is valid IEC 61131-3 (§7.3.3.1, assignment over a "multi-element variable"). So this is a genuine codegen gap that we'd want to close — exactly what P9999 means. Reporting it as an internal error would tell a user they'd hit a compiler bug when they wrote correct ST.

The `else` arm just below is different but lands the same way. It fires for `x[1] := 5` where `x : DINT` — subscripting a scalar. That *is* an invalid program, and analysis doesn't catch it either (it reaches codegen and comes out as a generic P9999). It's an analyzer gap, not a compiler-invariant violation, so an internal error would misattribute it too.

So there are two real gaps here, both in the analyzer rather than codegen:

1. subscripting a non-array variable is not rejected
2. whole-element assignment of a structure array element is accepted but not codegen-able

Fixing (1) with a new P4xxx rule would then justify making the `else` arm an internal error, which I think is the end state you're after. I'd rather not add analyzer rules and problem codes inside this PR — want me to file that as a follow-up issue, or fold it in here?

---

compiler/codegen/src/compile_array_struct.rs
garretfick: Same answer as the `compile_array.rs` one, for the same reason — I've left this as `not_implemented`, but say the word and I'll change it.

This fires when a subscripted struct field turns out not to be an array, e.g. `h.n[1]` where `n : DINT`. Analysis lets it through:

```
$ ironplcc check p.st     # TYPE H : STRUCT n : DINT; ... r := h.n[1];
$                         # no output — passes
$ ironplcc compile ...
error[P9999]: Capability is not implemented (yet!)
  │  ^ Field 'n' is not an array type
```

So the compiler is not at fault here; the program is. An internal error would tell the user they found a bug in the compiler when they subscripted a scalar field. The right fix is an analyzer rule that rejects subscripting a non-array — the same missing rule as in the other thread — and *then* this arm becomes a genuine invariant and should be an internal error.

One nuance worth noting: after f597b93 this arm has a second caller. `a[i].values[j]` reuses it to check that `values` really is an array before appending its dimensions, so the message now covers both "you subscripted a scalar field" and "you subscripted a scalar field of an array element". Both are program errors, so the reasoning is unchanged.

I'd group this with the analyzer work rather than doing it here — same question as the other thread: follow-up issue, or fold it in?

---

compiler/codegen/src/compile_array.rs
garretfick: Tracked as follow-ups so this PR can move: **#1470** for the analyzer rule (subscripting a non-array is accepted today), and **#1471** for whole-element assignment of an array-of-struct element.

#1470 explicitly notes that once the analyzer rejects the shape, this arm and the `compile_array_struct.rs` one become genuine invariants and should be `internal_error_at` — so your change lands there rather than being lost.

---

compiler/codegen/src/compile_array_struct.rs
garretfick: Tracked as **#1470**, together with the `compile_array.rs` arm — same missing analyzer rule, and the issue records that both should become `internal_error_at` once it exists.

---

compiler/codegen/src/compile_array_struct.rs
garretfick: The open question here — whether to sweep the pre-existing capacity checks in `register_array_variable` and `allocate_struct_variable` from P9999 to P9997 — is now tracked as **#1473**, so it need not hold this PR.

The issue records the care it needs: only the capacity checks move (unsupported element types stay P9999), and since P9999 is public API, a program that produced it will start producing P9997, which is worth a release note.

---

Issue #1407, opened 2026-08-22
plc2plc renders array subscripts as `refs [ 0 ]`, which the parser then rejects


garretfick:
## Summary

`plc2plc` renders an array subscript with a space between the variable and the opening bracket (`refs [ 0 ]`). The parser rejects that spelling, so **any** rendered program containing an array subscript does not re-parse — in both expression and assignment-target position.

Same shape as #1404 (`myRef ^`), but a different rule and a wider blast radius: #1404 only affected dereference in expression position.

## Cause

`visit_array_variable` (`compiler/plc2plc/src/renderer.rs:1775`) uses `write_ws("[")`:

```rust
fn visit_array_variable(
    &mut self,
    node: &dsl::textual::ArrayVariable,
) -> Result<Self::Value, Diagnostic> {
    self.visit_symbolic_variable_kind(&node.subscripted_variable)?;

    self.write_ws("[");
    visit_comma_separated!(self, node.subscripts.iter(), Expr);
    self.write_ws("]");

    Ok(())
}
```

`write_ws` inserts a separating space when the buffer does not already end in one (`renderer.rs:77`), producing `refs [ 0 ]`.

## Why the parser rejects it

`symbolic_variable` (`compiler/parser/src/parser.rs:852`) chains its elements — `.field`, `[i]`, `^` — with no `_` (optional whitespace/comment) rule between them, so nothing may separate `refs` from its `[`.

The space *inside* the brackets is fine: `subscript_list` (`parser.rs:919`) is `'[' _ subscript ++ (_ ',' _) _ ']'`, so it already permits whitespace after `[` and before `]`.

Measured against `--dialect iec61131-3-ed3` on `main`:

| Source | Result |
|---|---|
| `value := refs[0];` | parses |
| `value := refs[ 0 ];` | parses |
| `value := refs[0 ];` | parses |
| `value := refs [0];` | **P0002 syntax error** |
| `value := refs [ 0 ];` | **P0002 syntax error** |
| `refs[ 1 ] := 5;` | parses |
| `refs [1] := 5;` | **P0002 syntax error** |

So only the space the renderer emits *before* `[` is the problem.

[...]

Issue #1404, opened 2026-08-22
plc2plc renders `ExprKind::Deref` as `myRef ^`, which the parser then rejects


garretfick:
## Summary

`plc2plc` renders a dereference in *expression* position with a space before the caret (`myRef ^`). The parser rejects that spelling, so the rendered output does not re-parse. Dereference in *variable* position renders correctly (`myRef^`).

## Cause

`compiler/plc2plc/src/renderer.rs:1679` uses `write_ws("^")` for `ExprKind::Deref`:

```rust
dsl::textual::ExprKind::Deref(expr) => {
    self.visit_expr(expr)?;
    self.write_ws("^");
    Ok(())
}
```

`write_ws` inserts a separating space when the buffer does not already end in one (`renderer.rs:77`), producing `myRef ^`. The variable-position path, `visit_deref_variable` (`renderer.rs:1694`), correctly uses `write("^")`.

## Why the parser rejects it

`symbolic_variable` (`compiler/parser/src/parser.rs:852`) chains its elements — `.field`, `[i]`, `^` — with no `_` (optional whitespace/comment) rule between them, and `unary_expression` (`parser.rs:1754`) likewise consumes trailing carets with no `_`. Measured against `--dialect codesys` on `main`:

| Source | Result |
|---|---|
| `v := p^;` | parses |
| `v := p ^;` | **P0002 syntax error** |
| `p^ := 1;` | parses |
| `p ^ := 1;` | parses (the assignment rule at `parser.rs:1817` *does* have `_` around its caret) |
| `p^.x := 1;` | parses |
| `p ^.x := 1;` | **P0002 syntax error** |

So the strictness is an accident of two rules disagreeing, not a deliberate rule.

## Why it has gone unnoticed

`parse_and_render_edition3` (`compiler/plc2plc/src/tests/common.rs:43`) renders but never re-parses, and `compiler/plc2plc/src/tests/reference_to.rs:83` asserts the buggy spelling directly:

```rust
#[case::deref_expression(
    "...    value := myRef^;\nEND_PROGRAM",
    "myRef ^"
)]
```

## Suggested fix

1. `renderer.rs:1679`: `write_ws("^")` → `write("^")`.
2. Update the expectation at `reference_to.rs:83` from `"myRef ^"` to `"myRef^"`.
3. Consider adding a re-parse step to the round-trip assertion for this case (as `plc2plc/src/tests/case.rs:33` already does), so a future stray space fails the test rather than passing silently.

[...]

Issue #1416, opened 2026-08-23
A TYPE alias to an elementary type with no initializer is a syntax error (`MY_ALIAS : INT;`)


garretfick:
## Summary

A type declaration whose base type is an **elementary** type name and which has **no** initializer and **no** subrange is rejected as a syntax error:

```
TYPE
   MY_ALIAS : INT;
END_TYPE
```

```
error[P0002]: Syntax error
2 │    MY_ALIAS : INT;
  │                  ^ Expected ' ' (space) | '\t' (tab) | '(' | '(* ... *)' (comment) | ':=' | '\n' (new line) | '{ ... }' (pragma). Found text ';' that matched token ';'
```

Adding either an initializer or a subrange makes it parse, and a user-defined base type name parses. Only the "bare elementary alias" spelling fails.

Dialect-independent — reproduces identically under the default dialect and `--dialect iec61131-3-ed3`.

## Measured on `main`

| Source | Result |
|---|---|
| `TYPE MY_ALIAS : INT; END_TYPE` | **P0002 syntax error** |
| `TYPE MY_ALIAS : BOOL; END_TYPE` | **P0002 syntax error** |
| `TYPE MY_ALIAS : REAL; END_TYPE` | **P0002 syntax error** |
| `TYPE MY_ALIAS : TIME; END_TYPE` | **P0002 syntax error** |
| `TYPE MY_ALIAS : STRING; END_TYPE` | **P0002 syntax error** |
| `TYPE MY_ALIAS : INT := 0; END_TYPE` | parses |
| `TYPE MY_ALIAS : INT (0..10); END_TYPE` | parses |
| `TYPE MY_ALIAS : STRING [10]; END_TYPE` | parses |
| `TYPE MY_ALIAS : ARRAY[0..3] OF INT; END_TYPE` | parses |
| `TYPE MY_BASE : (A, B); MY_ALIAS : MY_BASE; END_TYPE` | parses |
| `VAR x : INT; END_VAR` (for contrast) | parses |

## Cause

`type_declaration` (`compiler/parser/src/parser.rs:523`) tries its alternatives in order. For `MY_ALIAS : INT;`:

- `simple_type_declaration__with_constant` (`:552`) goes through `simple_spec_init__with_constant` (`:577`), which **requires** the `:=` — hence `':='` in the expected set;
- `subrange_type_declaration__with_range` requires the `(` — hence `'('`;
[...]

PR #1417, opened 2026-08-23
plc2plc: re-parse by default and fix six rendering bugs

garretfick:
## Summary

Refactored plc2plc renderer tests to **re-parse every rendering by default**, catching and fixing six latent bugs that produced text the parser rejects. The re-parse is now the core assertion: a renderer test proves not just that output is produced, but that the output is valid input.

## Problem

The plc2plc test suite compared rendered output against golden `*_rendered.st` files without re-parsing. This meant broken spellings were silently recorded as "expected" — issues #1404 and #1407 (dereference and array subscript spacing) survived in committed fixtures because the golden-file comparison couldn't distinguish correct output from output the parser rejects.

### Test Infrastructure (`compiler/plc2plc/src/tests/common.rs`)

- **`assert_round_trips(source, options)`** — parse → render → re-parse, requiring same AST. Returns rendered text for additional assertions.
- **`assert_resource_renders_to(source_name, rendered_name, options)`** — round-trip plus golden-file comparison.
- **`assert_round_trips_idempotently(source, options)`** — for renderings that normalize to different AST spellings; still re-parses but compares text-to-text.
- **`assert_library_renders_to_parseable_text(library, options)`** — for hand-built libraries.
- **`edition3()`** helper for Edition-3 dialect options.

All test files migrated to use these helpers, eliminating boilerplate parse/render/assert patterns.

### Renderer Fixes (`compiler/plc2plc/src/renderer.rs`)

| Bug | Rendered | Fix |
|---|---|---|
| Array subscript (#1407) | `refs [ 0 ]` | Use `write("[")` not `write_ws("[")` in `visit_array_variable` |
| Whole-valued real | `LREAL#11` | Preserve decimal point: `11.0` unless text already has `.`/`e`/`E` |
| Negative integer | `INT (- 10.. 10 )` | Write sign tight: `write_ws("-")` then digits, no space between |
| STRING/WSTRING quotes | `STRING [5] := 'v'` → `:="v"` | Swap quotes in `visit_string_type_declaration` (single for STRING, double for WSTRING) |
| Array initializer | `:= 2 , 7 , 18` | Restore enclosing `[ ]` brackets |
| Repeated element | `3 2` | Add `visit_repeated` override rendering `3 ( 2 )` |
| `VAR_CONFIG` blocks | One per initializer | Emit single block in `visit_configuration_declaration`; item visitors render just their line |

### Parser Fixes (`compiler/parser/src/parser.rs`)

- **`semisep_oneplus`**: Add `_` before terminating semicolon (was eating trailing space as separator)
- **`array_initial_elements`**: Allow whitespace inside repeat parentheses

### Test Coverage

- New regression cases in `reference_to.rs` for subscripts in expression/assignment-target position, `refs[0]^`, `PT^[0]`, `s.items[0]`, `grid[1, 2]`
- `this_super.rs` regains `THIS^.values[2]` case previously excluded due to #1407
- `declarations.rs` fixed hand-built `LateBoundDeclaration` to use user-defined base type (parser never produces elementary-type aliases)
- All test files simplified: boilerplate parse/render/assert replaced with single helper call

### Documentation

- Updated `specs/steering/syntax-support-guide.md` to require re-parsing in renderer tests
- Added plan document `specs/plans/2026-08-23-plc2plc-round-trip-reparse-by-default.md`

### Golden Files

Twelve `*_rendered.st` fixtures regenerated with corrected spellings:
- `reference_to_rendered.st`, `var

https://claude.ai/code/session_015v7NXFxgVBaiYXRKprVxyB


compiler/plc2plc/src/renderer.rs
```diff
@@ -507,9 +520,13 @@ impl Visitor<Diagnostic> for LibraryRenderer {
         self.write_ws("]");
 
         if let Some(init) = &node.init {
+            // Single quotes delimit a STRING, double quotes a WSTRING (IEC
+            // 61131-3 2.2.2) -- the same mapping visit_string_initializer
+            // uses. These two were swapped, so a STRING initializer rendered
```
garretfick: Don't describe history. You can probably omit most of this comment.

compiler/plc2plc/src/renderer.rs
```diff
@@ -843,12 +860,32 @@ impl Visitor<Diagnostic> for LibraryRenderer {
         if !node.initial_values.is_empty() {
             self.write_ws(":=");
 
+            // The list is bracketed: `:= [2, 7, 18]`. Without the brackets
```
garretfick: I'm unclear why you propose this comment. Just fix the bug.

compiler/plc2plc/src/renderer.rs
```diff
@@ -843,12 +860,32 @@ impl Visitor<Diagnostic> for LibraryRenderer {
...
+            // the rendering does not parse.
+            self.write_ws("[");
             visit_comma_separated!(self, node.initial_values.iter(), ArrayInitialElementKind);
+            self.write_ws("]");
         }
 
         Ok(())
     }
 
+    /// A repeated array element renders as `count(value)`. Without the
```
garretfick: Again, don't write history.

compiler/plc2plc/src/renderer.rs
```diff
@@ -1803,7 +1836,12 @@ impl Visitor<Diagnostic> for LibraryRenderer {
     ) -> Result<Self::Value, Diagnostic> {
         self.visit_symbolic_variable_kind(&node.subscripted_variable)?;
 
-        self.write_ws("[");
+        // No separating space before `[`: the parser's `symbolic_variable`
+        // rule chains `.field`, `[i]` and `^` with no optional-whitespace
```
garretfick: Is what the parser allows correct or is that a defect?

compiler/plc2plc/src/tests/fb_inheritance.rs
```diff
@@ -23,49 +31,37 @@ VAR
...
-        allow_fb_inheritance: true,
-        ..CompilerOptions::default()
-    };
-    let library_original = parse_program(source, &FileId::default(), &options).unwrap();
-    let rendered = write_to_string(&library_original).unwrap();
+",
+        &inheritance_options(),
+    );
 
     assert!(rendered.contains("EXTENDS FB_Motor"));
```
garretfick: Do we need these asserts if we are now doing round trips?

compiler/plc2plc/src/tests/declarations.rs
```diff
@@ -8,10 +8,15 @@ fn write_to_string_late_bound_declaration() {
         DataTypeDeclarationKind, LateBoundDeclaration, Library, LibraryElementKind, TypeName,
     };
 
-    // Create a library with a late bound declaration in code
+    // A late-bound declaration is what the parser produces when the base
+    // type is a name it cannot resolve yet, so the base must be a
```
garretfick: I think you are documenting history. We don't document history.

compiler/plc2plc/src/renderer.rs
```diff
@@ -175,7 +175,12 @@ impl Visitor<Diagnostic> for LibraryRenderer {
         node: &ironplc_dsl::common::SignedInteger,
     ) -> Result<Self::Value, Diagnostic> {
         if node.is_neg {
-            self.write("-");
+            // The separating space goes *before* the sign, not between the
+            // sign and the digits: the parser's signed-integer rule joins
+            // them into one token, so `- 10` does not parse.
```
garretfick: Is what the parser allows correct or is that a defect?

garretfick (commented):
See comments inline

compiler/plc2plc/src/renderer.rs
garretfick: Cut to one line stating the rule: single quotes delimit a STRING, double quotes a WSTRING.

Applied the same pass to every comment this change added, not just the flagged ones — the real-literal, VAR_CONFIG, subscript and `semisep_oneplus` comments were all describing what used to be wrong. They now state the constraint the code has to satisfy, or are gone where the code already says it.

---

compiler/plc2plc/src/renderer.rs
garretfick: Comment removed. `write_ws("[")` around the comma-separated list says it.

---

compiler/plc2plc/src/renderer.rs
garretfick: Cut to `/// Renders a repeated array element, `count(value)`.` — the syntax it produces, nothing about what it did before.

---

compiler/plc2plc/src/tests/declarations.rs
garretfick: Trimmed to the live constraint — three lines, all present tense:

```rust
// The base type must be a user-defined name: an elementary base resolves
// to a simple type declaration and never reaches this node. Building the
// library directly is the only way to reach it from one source file.
```

Both facts are why the test looks the way it does rather than what it used to look like: `LateBoundDeclaration` is unreachable with an elementary base, and unreachable from parsed source at all.

That constraint is also the bug I filed as #1416 — `TYPE MY_ALIAS : INT; END_TYPE` is rejected outright, so the base type in this test couldn't stay `INT` once the rendering had to re-parse.

---

compiler/plc2plc/src/tests/fb_inheritance.rs
garretfick: No — removed, here and in five other files. You're right that the round trip subsumes them: if `EXTENDS FB_Motor` were dropped or mangled, the re-parsed AST would differ and `assert_round_trips` fails.

Dropped 20 assertions in total: EXTENDS/IMPLEMENTS/INTERFACE/ABSTRACT (`fb_inheritance.rs`), enum explicit values, base-type suffix and `COLOR#RED` (`enums.rs`), `AND_THEN` (`short_circuit.rs`), `SCALE`/`180.5` (`constant_initializers.rs`), `pDevice`/`Delta` (`struct_init_expressions.rs`), the FB call-style initializer (`declarations.rs`), and the three names in `mixed_vars.rs` that the structural checks below them already cover.

Three kinds survive, because AST equality genuinely cannot make them:

1. **Identifier casing.** `Id`'s `PartialEq` compares `lower_case` (`dsl/src/core.rs:222`), so a renderer that lower-cased every identifier would round-trip clean. That is exactly what the `Tc2_Math`/`Tc2_Utilities` tests exist to prevent — `LREAL_TO_FMTSTR`, `iPrecision`, `bRound` have to come back with vendor casing intact (REQ-CL-plc2plc-001). Those `contains` calls now say so at the call site.
2. **Spellings the AST does not record.** `STRING [ 255 ]` in `declarations.rs` — the DSL keeps no bracket/paren marker, so both spellings parse to the same AST and only a text check pins which one is emitted.
3. **Tests asserting idempotency rather than AST equality.** `case.rs`'s bit-string literals decimalize to a different variant, so there is no AST comparison to lean on.

Wrote the rule into `syntax-support-guide.md` so the next test doesn't re-add the redundant kind.

---

compiler/plc2plc/src/renderer.rs
garretfick: **A parser defect, I think** — but one this change doesn't need to settle, and I'd rather not settle it here.

The repo already states the test, in the `self_ref` comment two rules up (`parser.rs:851`): whitespace is accepted "since nothing in the grammar joins them into one token." Nothing joins `refs` and its `[` into one token, so by that rule `refs [ 0 ]` should parse. `subscript_list` is already inconsistent with rejecting it — it allows `_` *inside* the brackets, so `refs[ 0 ]` is fine and only the space *before* the `[` is refused. Same for `.field`: `s . x` is rejected while `s.x` is accepted, and there's no principle separating those two.

Two reasons I left the parser alone:

1. The renderer change is right either way. `refs[ 0 ]` is the canonical spelling regardless of what the parser tolerates; emitting a gratuitous space there was never intended.
2. Widening `symbolic_variable` to allow `_` between every element is a language-surface change that wants its own diff and its own tests — it affects `.field`, `[i]` and `^` together. #1407 and #1404 both explicitly scoped it out for that reason.

Happy to do it as a follow-up if you want it — say the word and I'll file it or write it.

(For contrast, the sign-vs-digits case on line 180 is *not* the same question — see that thread.)

---

compiler/plc2plc/src/renderer.rs
garretfick: **Correct, not a defect** — this one is genuinely different from the subscript case on line 1840.

`signed_integer__negative` is `Minus Digits` (`parser.rs:365`). The `-` there is part of a *literal*, not a delimiter between two tokens, and the free-format rule permits whitespace between delimiters but not inside a literal, identifier or keyword. `-10` is one lexical unit that happens to be spelled with two tokens; `- 10` is not a signed-integer literal. So the parser is doing the right thing and the renderer is what was wrong.

That is the opposite conclusion to line 1840, and deliberately so — the distinction is whether whitespace would fall *inside* a lexical unit or *between* two of them. `refs` and `[` are two units; `-` and `10` are one.

Flagging my confidence honestly: the subscript answer follows from the grammar as written and I'm sure of it. This one rests on reading `-10` as a single literal rather than a unary minus applied to `10`, which is a judgement about intent. If you read it the other way — sign as an operator — then allowing `- 10` would be consistent and this would be a defect too. The renderer emitting `-10` is correct under either reading, so nothing in this PR depends on which way you call it.

---

garretfick: Correcting my previous reply — I said `- 10` "is not a signed-integer literal, so the parser is doing the right thing," which implied the text is illegal. It isn't. You're right: it's legal, via a different production.

Measured on this branch:

| Context | `-10` | `- 10` |
|---|---|---|
| expression (`x := ...;`) | parses | parses |
| subrange (`INT(-10..10)`) | parses | **P0002** |
| array bound (`ARRAY[-10..10]`) | parses | **P0002** |

So the two positions behave differently, and correctly:

- **Expression position** has `unary_expression ::= unary_operator? _ primary_expression` (`parser.rs:1771`), which allows the space. Both spellings are legal there. In fact both produce the *same* AST here — `UnaryOp(Neg, Const(IntegerLiteral(10)))` with `is_neg: false` — so the unary route wins even for `-10`, and no signed-integer literal is involved at all.
- **`signed_integer` position** — subrange bounds, array bounds, enum values — has no expression production to fall back on. `signed_integer__negative ::= Minus Digits` is the only route, so only `-10` is legal there.

Which means `visit_signed_integer` with `is_neg: true` only ever fires in the second kind of position. That's what the fixture shows: `INT ( -10.. 10 )` in `type_decl_rendered.st`. Rendering `- 10` there is a syntax error, full stop — no unary-operator reading is available to rescue it.

Conclusion unchanged, reasoning fixed: the parser is right, the renderer was wrong. But the reason is "a subrange bound admits no expression", not "whitespace can't appear inside a literal."

---

compiler/plc2plc/src/renderer.rs
garretfick: Re-answering against the standard rather than against our parser, and being explicit about what I can and can't support.

**What I can't do:** cite a clause. I don't have the IEC 61131-3 text, so I can't quote the free-format rule verbatim, and I'm not going to invent a clause number.

**What I can argue.** In Annex A, `array_variable ::= subscripted_variable subscript_list` and `subscript_list ::= '[' subscript {',' subscript} ']'`. Those are two productions joined by juxtaposition — `[` is a delimiter, `subscripted_variable` reduces to an identifier. Nothing binds them into one token, so free-format whitespace should be permitted between them. `refs [ 0 ]` looks legal to me.

**The asymmetry with the dash, which is the useful part.** `- 10` has a competing reading — unary operator versus signed literal — so whitespace there genuinely changes which production applies. `refs [ 0 ]` has no competing reading. A `[` following a variable can only begin a subscript; there is no other production it could start. So there's no meaning to get wrong, only text to accept or reject.

**Internal evidence that we already treat it as legal.** We accept a space before `[` in every declaration context and reject it only in use:

| | |
|---|---|
| `TYPE A : ARRAY [ 0..3 ] OF INT; END_TYPE` | parses |
| `TYPE S : STRING [ 10 ]; END_TYPE` | parses |
| `v := refs[0];` | parses |
| `v := refs [ 0 ];` | **P0002** |

Same delimiter, same shape, opposite answers. And we already render `ARRAY [ 1.. 8 ]` with that space in the committed fixtures. That inconsistency is hard to defend as intentional.

So: probably legal, and a gap on our side — but a gap the renderer shouldn't rely on either way, since `refs[ 0 ]` is the canonical spelling regardless. Still worth its own diff, since widening `symbolic_variable` affects `.field` and `^` at the same time.

If you want it settled properly rather than argued from structure, I can check the spelling against a reference implementation's grammar (matiec derives its parser directly from Annex A) — say the word.

---

PR #1435, opened 2026-08-23
Show VM system uptime in the debugger Runtime scope

garretfick:


compiler/vm/src/vm.rs
```diff
@@ -220,6 +220,7 @@ impl<'a> VmReady<'a> {
             frames: self.frames,
             shared_globals_size,
             scan_count: 0,
+            system_time_us: 0,
```
garretfick: Can this be a duration object?

compiler/vm/src/vm.rs
```diff
@@ -485,11 +493,12 @@ impl<'a> VmRunning<'a> {
         let task_id = self.program_instances[0].task_id;
 
         if !resuming {
-            // Fresh scan: inject system uptime, then reset the resume state so
+            // Fresh scan: record the clock (and inject the uptime system
+            // variables), then reset the resume state so
             // run_instance starts with an empty frame stack (the dispatch loop
             // pushes the entry frame). When resuming, the preserved frame count
             // is non-zero and the paused frames survive in place.
-            self.inject_system_uptime(current_time_us);
```
garretfick, line 492: I'm a bit concerned about the name `current_time_us`. This is showing how long the VM has been running, not the current time. Those are different things. The behavior is right, just the variable name is wrong.

garretfick (commented):
See comments

compiler/vm/src/vm.rs
garretfick: Agreed, and it was wrong everywhere, not only here — every caller passes elapsed-since-start: `start.elapsed()` in `ironplcvm`, `scan_count() * cycle_time_us` in the playground and LSP runner, a simulated clock in the MCP runner, and the per-round bump in the debug driver. Nothing ever passes a wall clock.

Renamed `current_time_us` -> `uptime_us` across `run_round`, `run_round_debug`, `run_instance`, `execute`, and the scheduler, plus each caller's local (`current_us` -> `uptime_us`). Parameters are positional, so this is a rename with no call-site churn.

Fixed in d2c1cde.

---

compiler/vm/src/vm.rs
garretfick: Yes for the stored value — done in d2c1cde. The field is now `uptime: Duration` (`core::time::Duration`), initialized `Duration::ZERO`, and the accessor is `VmRunning::uptime() -> Duration` instead of `uptime_ms() -> i64`. Choosing the unit moves to the caller, so the DAP server renders `running.uptime().as_millis()` for the `LINT` it reports, and the tests read `Duration::from_millis(5)` rather than a bare `5`.

I stopped short of typing the clock *parameters* (`run_round(uptime_us: u64)` and friends), for two reasons — happy to go further if you'd rather:

1. They flow straight into the scheduler's `next_due_us` / `interval_us` / `watchdog_us` comparisons, and those are `u64` microseconds decoded from the container's binary task table. A `Duration` parameter would convert back to micros at each comparison, so the type would be decoration rather than enforcement.
2. It is 157 call sites across eight crates (`vm`, `vm-cli`, `playground`, `ironplc-cli`, `mcp`, `codegen` tests, `benchmarks`, and the VM's own suite) — `run_round(0)` becoming `run_round(Duration::ZERO)` everywhere.

Worth noting the call sites would get *nicer* if we did: `ironplcvm` currently writes `start.elapsed().as_micros() as u64`, and `Instant::elapsed()` already hands back a `Duration`. If you want that, I would rather do it as its own mechanical PR than fold it into this one.

---

PR #1436, opened 2026-08-23
Assumed cycle time for freewheeling tasks under simulated time

garretfick:
The run MCP tool executes for a simulated duration derived from task
intervals, which a freewheeling task does not have, so the round loop
breaks after one cycle and reports a clean completion.

Plan an assumed freewheeling cycle time: defaulted to 100 ms, caller
overridable, and always reported back, applied by rewriting the task
table before the VM loads it.

Claude-Session: https://claude.ai/code/session_01PYHirGM4Z2N6jCb6EdpWTW


compiler/mcp/src/tools/compile.rs
```diff
@@ -257,10 +257,13 @@ fn extract_task_program_metadata(library: &Library) -> (Vec<TaskMeta>, Vec<Progr
...
                 .unwrap_or_else(|| "default".to_string());
 
+            // No CONFIGURATION: codegen keeps the freewheeling task that
+            // `ContainerBuilder` synthesizes, so that is what `run` will
+            // schedule and what the caller needs to see.
             let tasks = vec![TaskMeta {
                 name: program_name.clone(),
                 priority: 0,
-                kind: "event".to_string(),
+                kind: "freewheeling".to_string(),
```
garretfick: I realize this is beyond the scope of this change, but why is this a string value? We would normally represent this as an enumeration. Can we do that and then have this converted for the JSON representation by a converter?

compiler/mcp/src/tools/run.rs
```diff
@@ -122,6 +131,12 @@ pub struct RunSummary {
     pub final_values: Map<String, Value>,
     pub completed_cycles: Map<String, Value>,
     pub terminated_reason: String,
+    /// The cycle time assumed for the container's freewheeling tasks, in
+    /// milliseconds (REQ-TOL-mcp-049). Absent when the container declared an
+    /// `INTERVAL` for every task, so every timestamp in the trace follows from
+    /// the program itself.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub assumed_freewheeling_interval_ms: Option<f64>,
```
garretfick: Let's drop the prefix "assumed_freewheeling_". Can we just report the interval we used?

compiler/mcp/src/tools/run.rs
```diff
@@ -303,11 +326,33 @@ fn build_success_response(
...
+/// duration a task could actually run at, which `interval_us_from_ms` decides
+/// — the debugger takes the same input and applies the same rule.
+fn resolve_freewheeling_interval_us(interval_ms: Option<f64>) -> Result<u64, Vec<Diagnostic>> {
+    let Some(ms) = interval_ms else {
+        return Ok(DEFAULT_FREEWHEELING_INTERVAL_US);
+    };
+
+    interval_us_from_ms(ms).ok_or_else(|| {
+        vec![validation(&format!(
+            "freewheeling_interval_ms ({ms}) must be greater than 0 and at most \
```
garretfick: This says a maximum, but I don't see how we enforce that at all.

compiler/mcp/src/runner.rs
```diff
@@ -186,6 +186,10 @@ pub struct RunOutcome {
     pub truncated: bool,
     /// Diagnostic message when `terminated_reason == Error`.
     pub error_message: Option<String>,
+    /// The cycle time assumed for the container's freewheeling tasks, in
+    /// microseconds. `None` when the container declared an `INTERVAL` for
+    /// every task, so nothing was assumed.
+    pub assumed_freewheeling_interval_us: Option<u64>,
```
garretfick: I want this to be just be "freewheeling_interval_us".

Why are we not using `core::time::Duration;`. That is no-std compatible and it removes the need to have `us` everywhere for variable names.

compiler/vm-cli/src/dap/server.rs
```diff
@@ -462,6 +481,45 @@ fn launched_session<R: BufRead, W: Write>(
 
 /// Whether the launch's `scanLimit` bound (if any) has been reached, so the
 /// session should terminate rather than start another cycle.
+/// How far program time advances per scan for a program that declares an
+/// `INTERVAL`. A flat 1 ms, pending #1397.
+const DEBUG_SCAN_ADVANCE_US: u64 = 1_000;
```
garretfick: As in other places, I think this should be a Duration type.

compiler/vm-cli/src/dap/server.rs
```diff
@@ -462,6 +481,45 @@ fn launched_session<R: BufRead, W: Write>(
...
+}
+
+/// The console line telling the user which cycle time the session assumed.
+///
+/// Every timer in a freewheeling program elapses on this assumption, so a
+/// session that did not state it would leave the user unable to explain what
+/// they are watching.
+fn freewheeling_notice(interval_us: u64) -> String {
+    format!(
+        "This program declares no INTERVAL, so it has no scan cycle time of its own. \
```
garretfick, line 507: Let's not say specifically how to change the behavior. Just point to change "launch configuration". Otherwise we are sure to break this text in the future and not update the text.

garretfick (commented):
See comments

compiler/mcp/src/tools/compile.rs
garretfick: Done. `TaskKind` is now an enum in `cache.rs` with `#[serde(rename_all = "lowercase")]` doing the conversion, so `TaskMeta` and `TaskInfo` both carry variants and only the wire format deals in strings. The JSON is unchanged.

Added a test pinning the spelling — `cyclic` / `single` / `freewheeling` is the published `compile` contract (REQ-TOL-mcp-032), so it shouldn't be free to drift with a variant rename.

---

compiler/mcp/src/tools/run.rs
garretfick: Renamed to `summary.interval_ms`, and I took "just report the interval we used" literally: it is now always present and reports the cycle time the run executed at, whether the program declared it or the caller supplied it. The response no longer distinguishes assumed from declared at all — a caller who wants to know which can see it from `compile`'s task `kind`.

Flagging the one thing that isn't a pure rename: a field called `interval_ms` that disappeared whenever the program declared its own interval felt like the wrong shape, so I made it unconditional rather than optional. That's exact today because the VM schedules one task; a container with several tasks at different rates would need this reported per task, which I noted in REQ-TOL-mcp-049. Happy to make it optional again if you meant the narrower change.

---

compiler/mcp/src/tools/run.rs
garretfick: It was enforced — in `interval_us_from_ms` over in the vm crate — but you couldn't see that from this line, and a message restating a bound it doesn't itself apply is exactly how the two drift apart.

Restructured so the enforcement is what produces the text: `interval_from_ms` now returns `Result<Duration, IntervalError>`, and `IntervalError::TooLong`'s `Display` names the bound from `MAX_FREEWHEELING_INTERVAL` itself. The diagnostic here just interpolates that error, so it can't claim a limit the check doesn't have. The debugger takes the same rule, and there's a test asserting the rendered bound.

---

compiler/mcp/src/runner.rs
garretfick: Switched to `Duration` — `core::time::Duration` in the vm crate, so `freewheeling` stays no-std-compatible. Microseconds now appear only at the two boundaries that genuinely count in them: the container's `interval_us` field and the VM clock, via a single `interval_us()` helper.

One deviation worth calling out: the field ended up as `RunOutcome::interval` rather than `freewheeling_interval`. Once the summary reports the interval the run actually used (your other comment), this field means "the cycle time we ran at", which is wrong to call freewheeling for a program that declares an `INTERVAL`. The runner *parameter* is still `freewheeling_interval` — that one really is the assumption. If you'd rather the outcome stay freewheeling-specific and optional, that follows from the other thread.

---

compiler/vm-cli/src/dap/server.rs
garretfick: Done — `DEBUG_SCAN_ADVANCE` and `assumed_freewheeling_interval` are both `Duration` now, and the conversion to microseconds happens only where the VM clock needs it.

---

compiler/vm-cli/src/dap/server.rs
garretfick: Agreed. Now: "This program declares no INTERVAL, so it has no scan cycle time of its own. Assuming 100 ms per scan; change it in the launch configuration." The test asserts on "launch configuration" rather than the setting name, so it won't pin the old spelling either.

The name still appears in `docs/reference/runtime/ironplcvmd.rst`, in the launch-arguments table — that page has to name it to document it, and it's the page anyone renaming the setting would already be editing.

---

compiler/mcp/src/runner.rs
```diff
@@ -227,16 +231,33 @@ impl TerminatedReason {
 /// a clean finish. `limits.max_duration_ms` is the sandbox ceiling
 /// (REQ-ARC-mcp-030); being cut short by it is not, and reports
 /// [`TerminatedReason::Duration`].
+///
+/// `freewheeling_interval` is the cycle time to assume for tasks that declare
+/// no `INTERVAL` (REQ-TOL-mcp-049). A freewheeling task has no rate of its own,
```
garretfick: There should still be the special case of "run as fast as you can". I think the parameter should be optional. When set, it defines the freewheeling interval. When not set, then run as fast as you can.

garretfick: Done — `freewheeling_interval` is `Option<Duration>` now. `Some` assumes that rate; `None` leaves the tasks freewheeling, and since the scheduler reports a freewheeling task ready every round, the loop keeps scanning with the clock standing still until a sandbox limit ends the run. That also tightened the bug this PR is about: the round loop's `None` arm now only fires when the container has nothing that can ever be scheduled, which is the case it was always meant to catch.

There's a test driving `execute` directly — 25 cycles against a tightened sample cap, `interval == Duration::ZERO`, every sample at `time_ms: 0`.

Two things I left for you rather than deciding:

**The `run` tool still passes `Some` on every call**, defaulting to 100 ms when the caller omits `freewheeling_interval_ms` — that's the default you asked for earlier, and I didn't want to reverse it on a comment aimed at the runner's parameter. So `None` is currently reachable only from the runner. If the mode should be callable from `run`, the natural spelling is `freewheeling_interval_ms: 0`, which matches codegen — `INTERVAL := T#0s` already compiles to a freewheeling task. Today 0 is rejected as out of range.

**Such a run ends at `sample_cap` or `fuel`, so it reports `ok: false`.** Honest — you asked for an unbounded spin and we stopped you at a limit — but odd for a deliberately requested mode, and `ok` is REQ-TOL-mcp-047's contract, so I didn't want to bend it without your say.

---

garretfick: Replaced the spin with a rejection, per your follow-up — `963be85`.

A container with a freewheeling task now requires `freewheeling_interval_ms`. Without it the run returns `ok: false` before the VM starts:

> Task 'Main' declares no INTERVAL, so it has no scan cycle time and the run has nothing to advance simulated time by. Set `freewheeling_interval_ms` to the scan time to run it at, or declare `INTERVAL` on the task in a CONFIGURATION.

No 100 ms default in `run` any more. `duration_ms` was already required by the schema, so the interval was the missing half of the terminal conditions.

The runner still takes `Option<Duration>`, but `None` now means "the caller supplied nothing", which is only legal when every task declares an `INTERVAL`; `None` against a freewheeling container returns `Err`. The tool rejects it first, so that check is a guard keeping the invariant next to the code that depends on it. The round loop's `None` arm went back to meaning "nothing is schedulable", which is the case it was written for.

The debugger keeps the 100 ms default and its console notice — a launch that refuses to start is a poor answer to a missing setting. REQ-TOL-mcp-049 and the docs page both state that split now, so the asymmetry is on the record rather than accidental.

I built the spin because "when not set, then run as fast as you can" read as a literal instruction. I flagged the `ok: false` oddity but should have said the louder thing: a mode whose only possible ending is a sandbox limit isn't useful to an agent. Sorry for the round trip.

---

Issue #1414, opened 2026-08-22
Whole-array assignment aliases the source instead of copying it


garretfick:
## Summary

Assigning one array variable to another copies the data-region *offset* rather
than the elements, so the destination becomes an alias of the source. Every
later read **and write** through the destination lands in the source's storage,
and the destination's own region is orphaned. No diagnostic is produced.

```st
PROGRAM main
VAR
    x : ARRAY[1..2] OF DINT;
    y : ARRAY[1..2] OF DINT;
    rx : DINT;
    ry : DINT;
END_VAR
    y[1] := 5;
    x := y;
    x[1] := 99;
    rx := x[1];   (* 99 *)
    ry := y[1];   (* 99 -- should be 5 *)
END_PROGRAM
```

Observed against the VM:

| Target type | `x[1]` | `y[1]` | |
|---|---|---|---|
| `STRUCT` | 99 | 5 | correct |
| `ARRAY[1..2] OF DINT` | 99 | **99** | aliased |
| `ARRAY[1..2] OF <struct>` | 99 | **99** | aliased |

Structures are unaffected — they already have a copy protocol.

## Cause

`compile_statement`'s assignment path special-cases a whole-struct target
(`compile_stmt.rs:222`) but has no equivalent for arrays. An array target
therefore falls through to the scalar arm (`compile_stmt.rs:278`), which loads
the RHS variable and stores it. Since an array variable's slot holds its
data-region byte offset, that copies the offset:

```
x := y   =>   LOAD_VAR_I32  y
              STORE_VAR_I32 x
```

## Expected

IEC 61131-3 §7.3.3.1 defines the assignment statement over a "single or
multi-element variable" — arrays and structures alike — as a value copy. There
is no reference semantics for a non-`REF_TO` variable, so this should be a deep
copy of the elements.

## Notes

The struct copy protocol (`compile_stmt.rs:230`) cannot simply be reused as-is:
it pushes every slot of the value onto the operand stack, then pops them back.
That is fine for a handful of struct fields, but a variable may occupy up to
`MAX_DATA_REGION_SLOTS` = 32768 slots (`compile.rs:107`), so applying it to
arrays would blow the operand stack. The same latent limit applies to a
sufficiently large struct today.

Two ways out:

- **Emitted loop** — a bounded copy loop in bytecode (index variable,
  `LOAD_ARRAY`/`STORE_ARRAY`, `CMP_BR`). No VM change, but O(n) bytecode per
[...]

PR #1443, opened 2026-08-25
Whole-aggregate assignment copies the value instead of aliasing (#1414)

garretfick:
Plan only — no implementation. Opening this for review of the approach before any code is written.

Adds `specs/plans/2026-08-25-whole-aggregate-copy.md` for #1414 (whole-array assignment aliases the source instead of copying it).

## What the plan proposes

**A `COPY_REGION` opcode that carries no length.** The issue floated `COPY_REGION dst_var, src_var, n_slots`. The plan argues against the length immediate: an unvalidated `n` relocates this exact bug from codegen into the opcode, since a size that is too large silently walks into the neighbouring variable and the data-region bounds check will not catch it. Instead the instruction names two array descriptors and the VM derives both sizes from them, trapping if they disagree — the same discipline `LOAD_ARRAY` already follows by taking `total_elements` from the descriptor rather than trusting an immediate.

```
COPY_REGION = encode_opcode(OP_CLASS_REGION_OP, 0)   // 0xF8
  [op][dst_var:u16][dst_desc:u16][src_desc:u16]      // 7 bytes
  pops src_offset
```

Op class `0x3E` becomes `OP_CLASS_REGION_OP` with the type tag selecting the region operation, following the `BOOL_OP`/`STACK_OP` precedent. Only two op-class slots remained, so this leaves three region ops for later and keeps `0x3F` free.

Destination by variable index so it is scope-checked; source offset from the operand stack so `s := t` and `s := f()` (struct-returning function) share one path.

**Two facts the plan establishes first**, since they are what make the design work:

- IEC 61131-3 has no array slice notation. `y[1..2]` is a syntax error today (`subscript ::= expression`), and ironplc does not implement Ed. 3 `ARRAY[*]`. So every whole-aggregate copy has a compile-time-constant length and no runtime length arithmetic is needed.
- `COPY_REGION` and `STORE_ARRAY` are not substitutes. Folding `STORE_ARRAY` into it would drop the index bounds check that is the only thing stopping `x[3] := 1` on an `ARRAY[1..2]` from writing into the next variable.

**Split of responsibility:**

| Layer | Owns |
|---|---|
| Analyzer (new P2037) | declared types identical — element type, dimensions, `STRING` max_len/width, struct type |
| VM (new V9018 + bounds) | descriptor byte sizes agree; both ranges inside the data region |

The analyzer rule is scoped to aggregates only. Extending it to scalars is the correct end state but would start rejecting scalar assignments that compile today and interacts with implicit widening (ADR-0029/0031) — its own change.

**Whole-struct assignment migrates onto the same opcode**, deleting the push-all-slots protocol at `compile_stmt.rs:222-253` and with it the operand-stack ceiling the issue flagged as latent for large structs.

## Notes

- No `FORMAT_VERSION` bump — additive opcode in a free slot, same as the `CMP_BR` addition.
- `ARRAY OF STRING` needs no special case: a byte copy is correct precisely because the analyzer has already required identical types, so each element's `[max_length][cur_length][data]` header is copied from a source with the same `max_length`.
- The plan flags that `specs/design/bytecode-instruction-set.md` is already stale (says `FORMAT_VERSION = 2`, and still lists `0x3D` as free though `CMP_BR` took it) and folds the correction into the same pass.
- It also flags `project/src/disassemble.rs` as a silent-failure risk: it hardcodes its own `pc +=` steps rather than calling `opcode::instruction_size`, so a missing arm misaligns every instruction after the new one.

Closes nothing yet; #1414 stays open until the implementation lands.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DdURgoQ8MJtYKuhVAQ6UjD)_


specs/plans/2026-08-25-whole-aggregate-copy.md
```diff
@@ -0,0 +1,344 @@
...
+(`compile_stmt.rs:222`) but has no equivalent for arrays, so an array target
+falls through to the scalar arm (`compile_stmt.rs:278`) and emits:
+
+```
+x := y   =>   LOAD_VAR_I32  y
+              STORE_VAR_I32 x
+```
+
+Since an array variable's slot holds its data-region byte offset, that copies
+the offset. Structures are unaffected — they already have a copy protocol.
```
garretfick: Can we document the copy protocol for structures in an appendix?

specs/plans/2026-08-25-whole-aggregate-copy.md
```diff
@@ -0,0 +1,344 @@
...
+
+IEC 61131-3 has no array slice or array-section notation. The grammar is
+`subscript_list ::= '[' subscript {',' subscript} ']'` with
+`subscript ::= expression` (`parser/src/parser.rs:936-937`) — commas separate
+*dimensions*, and `..` is not an operator, so `y[1..2]` is a syntax error today.
+Ranges appear only in declarations (`ARRAY[1..10] OF INT`), subrange types, and
+CASE labels. Array sections are a Fortran/Ada feature.
+
+The one other construct that could give an aggregate a runtime size is the
+Ed. 3 variable-length array (`ARRAY[*]`, VAR_IN_OUT only), which ironplc does
```
garretfick: We need to decide how we would implement this.

specs/plans/2026-08-25-whole-aggregate-copy.md
garretfick: Added as "Appendix: the existing struct copy protocol" in 9c785ba, with the emitted sequence written out, its cost (~`4n + 4` instructions, peak operand stack `n + 1`, vs 2 and 1 for `COPY_REGION`), and why the store loop runs in reverse.

Documenting it surfaced a third defect I had not called out: the protocol reads the source through the *destination's* descriptor and slot count, so `a := b` between different struct types over-reads past `b`'s region when `a` is the larger. Nothing checks that today — the new P2037 rule closes it. The appendix also notes the transient window where `dst_var` holds the source's offset, which is this issue's aliasing as an intended intermediate state.

---

specs/plans/2026-08-25-whole-aggregate-copy.md
garretfick: Fair — "we don't implement it" isn't a decision. Worked it out in 9c785ba as a new section, "If variable-length arrays arrive later".

Short version: `ARRAY[*]` is only ever a `VAR_IN_OUT` formal parameter bound to a caller's actual array, so its size is a property of the call, not the program text. The callee therefore cannot have a container-level descriptor for it — the parameter has to carry a **runtime** descriptor, which is the same thing `LOWER_BOUND`/`UPPER_BOUND` would read. That makes it a second region operation, not a change to this one:

```
COPY_REGION     = encode_opcode(OP_CLASS_REGION_OP, 0)   // container descriptors
COPY_REGION_DYN = encode_opcode(OP_CLASS_REGION_OP, 1)   // runtime descriptors
```

Zero op-class slots, and `0x3F` still stays free. This is the concrete payoff for spending a type tag rather than the op class.

One consequence lands on the current work, which is why it was worth deciding now: with VLAs the analyzer can compare element types but not extents, so `RegionSizeMismatch` stops being a backstop against codegen bugs and becomes a legitimate runtime error a correct compiler can produce from a correct program. So the V9018 documentation should be worded now so that reframing doesn't contradict it — define it as "source and destination regions differ in size" and put "indicates a compiler defect" in the cause section rather than the definition. I've folded that into the plan.

Nothing here is implemented; it's recorded so the opcode's shape is a deliberate choice rather than an artifact of what the language supports today.

---

compiler/analyzer/src/rule_assignment_aggregate_type_compat.rs
```diff
@@ -0,0 +1,390 @@
+//! Requires both sides of a whole-aggregate assignment to have identical
+//! declared types.
+//!
+//! IEC 61131-3 §7.3.3.1 makes assignment over a multi-element variable a value
+//! copy. Codegen implements that with `COPY_REGION`, whose length the VM
+//! derives from the two array descriptors — so the two ends must describe the
+//! same shape. The VM cross-checks the derived byte sizes and traps
+//! (`RegionSizeMismatch`, V9018), but that is a backstop against a compiler
```
garretfick: Code generation doesn't need a new problem code. If a rule such as this should catch earlier, then that is an internal error and we should reuse that problem code.

compiler/analyzer/src/rule_assignment_aggregate_type_compat.rs
```diff
@@ -0,0 +1,390 @@
...
+            ));
+        }
+    }
+}
+
+impl Visitor<Diagnostic> for RuleAggregateAssignment<'_> {
+    type Value = ();
+
+    fn visit_function_declaration(&mut self, node: &FunctionDeclaration) -> Result<(), Diagnostic> {
+        self.clear_variables();
```
garretfick: We just committed a new approach for entering and exiting scopes. We should use that mechanism.

compiler/analyzer/src/rule_assignment_aggregate_type_compat.rs
```diff
@@ -0,0 +1,390 @@
...
+        self.var_decls.clear();
+    }
+
+    /// Resolves a declared variable to its [`IntermediateType`].
+    ///
+    /// Handles both spellings a variable's type can take: a named type
+    /// (`p : Point`) resolved through the type environment, and an inline
+    /// specification (`a : ARRAY[1..2] OF DINT`) built from the declaration.
+    fn declared_type(&self, id: &Id) -> Option<IntermediateType> {
+        match self.var_decls.get(id)? {
```
garretfick: I don't see how this handles variables that have global scope. Wouldn't those show up as being unknown names?

compiler/codegen/src/compile_aggregate.rs
```diff
@@ -0,0 +1,135 @@
...
+    let Some(target_name) = resolve_variable_name(&assignment.target) else {
+        return Ok(false);
+    };
+
+    let Some(dst) = resolve_region(ctx, target_name)? else {
+        return Ok(false);
+    };
+
+    let src_desc_index = resolve_source_descriptor(ctx, &assignment.value).ok_or_else(|| {
+        Diagnostic::not_implemented(Label::span(
```
garretfick: This should be internal error rather than no implemented.

compiler/codegen/tests/it/end_to_end_aggregate_copy.rs
```diff
@@ -0,0 +1,294 @@
+//! End-to-end tests for whole-aggregate assignment (`x := y` where both
+//! sides are arrays or structures).
+//!
+//! IEC 61131-3 §7.3.3.1 makes assignment over a multi-element variable a
+//! value copy. Before COPY_REGION an array target copied the data-region
+//! offset instead, so the destination aliased the source (issue #1414).
```
garretfick: Don't reference issues in comments. Just say what's happens.

compiler/container/src/opcode.rs
```diff
@@ -502,6 +513,31 @@ pub const LOAD_ARRAY_DEREF: Opcode = encode_opcode(OP_CLASS_LOAD_ARRAY_DEREF, 0)
...
+/// the class of bug this opcode exists to prevent.
+///
+/// The destination is named by variable index so it is scope-checked; the
+/// source arrives as a stack offset so that a struct-returning function call,
+/// which leaves its `data_offset` on the stack and has no variable index in
+/// the caller's scope, uses the same instruction.
+///
+/// Overlapping regions are well defined (the VM uses `copy_within`), so
+/// `x := x` is a no-op rather than corruption.
+pub const COPY_REGION: Opcode = encode_opcode(OP_CLASS_REGION_OP, 0);
```
garretfick: Is there another "class" that we can put this into that is not full (and isn't really a class)? We had this idea of classes so that we can have a smaller dispatch table, but we don't want them to unnecessarily occupy spaces (and testing shows the dispatch table size isn't the performance bottleneck).

compiler/codegen/tests/it/end_to_end_aggregate_copy.rs
```diff
@@ -0,0 +1,294 @@
+//! End-to-end tests for whole-aggregate assignment (`x := y` where both
```
garretfick, line 1: Some additional tests:

* with global variables
* within function and function block (ideally I'd also like within a method).

compiler/container/src/opcode.rs
garretfick: Yes — moved it to `STORE_ARRAY` type tag 1, so `COPY_REGION` is now 0xAD and 0x3F is free again.

That class is the natural home: the type tag selects the *granularity* of a store into an array's storage — tag 0 writes one element at a runtime index, tag 1 writes the whole region. Tags 2–3 stay open, and a future `COPY_REGION_DYN` for variable-length arrays fits there instead of needing anything new.

I'd argued in the plan that the op class was worth spending because it preserved room for future region ops. That reasoning was backwards given what you say about the dispatch table not being the bottleneck: the type tags give the same room at no cost, and the class bought nothing. The spec now states the rule as "op classes exist to keep the dispatch table small, not to name operations".

Worth noting this is the second byte this instruction has had — it was 0xF8 until METHOD_CALL (#1449) landed on 0x3E and collided. Nothing outside the repo has consumed either value.

---

compiler/analyzer/src/rule_assignment_aggregate_type_compat.rs
garretfick: Switched to `ScopedTable`. Declarations now go in through `visit_var_decl` and the POU visitors just `enter()`/`exit()`, matching `rule_use_declared_symbolic_var`.

Two tests pin what the flat map could not express: a POU-local declaration shadowing an outer one of the same name, and a function block's locals *not* leaking into the next POU's scope.

---

compiler/analyzer/src/rule_assignment_aggregate_type_compat.rs
garretfick: Checked this empirically rather than reasoning about it, and the answer is not what I expected: the flat map already handled globals, but for a reason that made it fragile.

A program reaches a global through a `VAR_EXTERNAL` redeclaration, and that redeclaration is a `VarDecl` in the POU's own `variables` — so it carried the type, and the lookup succeeded. The `VAR_GLOBAL` block in the `CONFIGURATION` was never consulted at all.

So no bug today, but it only worked because IEC makes you restate the type locally. The `ScopedTable` change makes it structural instead of incidental: the outermost scope holds declarations made outside any POU, and a POU's lookups fall through to it. Added tests both directions (global→local and local→global) plus the shadowing case, and matching end-to-end tests that actually run the copy across the global/local boundary of the variable table.

---

compiler/codegen/src/compile_aggregate.rs
garretfick: Changed to `internal_error_at`.

It needed an analyzer change to actually be true, though. As written, `a := 5` reached this line: the rule stayed silent when the source type did not resolve, so a bad program — not a compiler defect — landed here. The rule now rejects *anything* that is not a same-typed aggregate on the right of an aggregate target, so the only way to arrive here is analysis and code generation disagreeing. There is a test for `a := 5` reporting P2037.

The one deliberate exception is a function call, which returns early: P4027 already compares a call's return type against its destination, so this rule would have produced a second diagnostic for one mistake. Codegen resolves those through `return_struct_desc_index` instead.

---

compiler/codegen/tests/it/end_to_end_aggregate_copy.rs
garretfick: Removed. The file header now describes the aliasing behaviour directly, and the two other references (to the array-of-struct issue, and "the issue's exact reproduction") are gone too — the latter now says what the test actually does.

---

compiler/codegen/tests/it/end_to_end_aggregate_copy.rs
garretfick: Globals are added, both directions — global→local and local→global — so the copy crosses the global/local boundary of the variable table. Matching analyzer tests too, plus shadowing and scope-leak cases.

**The function / function block / method ones I could not write: no aggregate can be declared as a POU local yet.** This is not about copying — both kinds fail on a plain member write, before a copy is involved:

```st
FUNCTION Probe : DINT
  VAR a : ARRAY[1..2] OF DINT; END_VAR
  a[1] := 5;        (* P9999 "Unsupported array element type", compile_array.rs *)
```
```st
FUNCTION Probe : DINT
  VAR b : Point; END_VAR
  b.x := 5;         (* P9999 "Variable 'b' is not a structure", compile_struct.rs *)
```

I confirmed both against `main` as well, so it predates this branch. Only `PROGRAM` variables and globals get registered in `array_vars` / `struct_vars`; `compile_user_function` registers scalars, strings and `REF_TO` arrays but never allocates a region for a local aggregate.

So the tests you want are currently unwritable rather than skipped. The test file says so at the point where they'd go, and names both error messages so whoever adds POU-local aggregates finds the spot. Happy to file that as its own issue if you'd like — it's a real gap and wider than this PR.

---

garretfick: Filed the blockers:

- **#1483** — `ARRAY` locals in a `FUNCTION`/`FUNCTION_BLOCK` are never registered, so any access fails
- **#1484** — `STRUCT` locals, same gap through a different path
- **#1485** — `LEN()` of a string array element is not implemented

#1483 and #1484 share a root cause — the local-variable pass in `compile_user_function` matches `Simple`, `String` and `Reference`, and `Array`/`Structure` fall into the trailing `_ => {}` — but they surface through different code and could be fixed independently, so they are separate.

Those two are what block the FUNCTION / FUNCTION_BLOCK / METHOD tests you asked for. #1485 is why the `ARRAY OF STRING` test stages elements through `STRING` variables.

Not filed: array-of-struct is already **#1383**.

Two things worth noting beyond the missing capability, both captured in the issues. The struct message is actively misleading — it says `Variable 'b' is not a structure` about a variable declared as one three lines above, which reads as a user type error rather than a compiler gap. And both report at the *use* rather than at the declaration that was never registered, so the location does not point at the thing to change.

---

PR #1456, opened 2026-08-29
Remove design and ADR references to implementation plans

garretfick:
First slice of #1464 — making implementation plans ephemeral.

Implementation plans are snapshots of intent that are not maintained after merge (228 of 237 were touched by exactly one commit; 1,074 checkboxes remain unticked). They are not stable reference targets, but 92 sites across the repository currently cite them. Those citations must go before any plan can be deleted.

This PR takes the 12 citations in `specs/design/` and `specs/adrs/`, plus the rule that prevents new ones.

## Triage result

Every one of the twelve was **unnecessary** rather than load-bearing. In each case the durable document already stated the fact in full and the plan link was decoration on top of it — ADR-0033 already explained why baseline measurement is required before the encoding change, and `debugger-support.md` already explained the `--features dap` reversal and the `shipped_binaries_guard` test that replaced it. No new ADR or design section was needed to absorb any of them.

## Changes

| File | Sites |
|---|---|
| `specs/design/debugger-support.md` | 6 |
| `specs/adrs/0033-opcode-encoding-by-class-and-type.md` | 2 |
| `specs/design/vm-performance.md` | 1 |
| `specs/design/library-interfaces/tc2-math.md` | 1 (kept the #1348 link, dropped the plan) |
| `specs/design/partial-access-bit-syntax.md` | 1 |
| `specs/design/ref-to.md` | 1 |

Two edits the removals forced, beyond deleting the links:

- ADR-0033 had a follow-on sentence beginning "The plan defers this work", which would have become a dangling referent once the citation above it was removed. It now states the deferral directly.
- `development-standards.md` instructed authors to split combined design and plan content "with cross-references between them", contradicting the rule added here. The reference is now one-way: the plan cites the design it implements, and the design never cites the plan.

`development-standards.md` also gains the rule prohibiting plan citations, so no new ones appear while the remaining 80 sites are migrated.

## Following the new process

This PR uses the process it is introducing. The plan was committed first and reviewed as a file diff, its undelivered content (Phases 2 and 3) was moved to #1464, and the plan file is deleted here — so nothing lands on `main`.

## Remaining

80 citation sites, tracked in #1464: 77 in `compiler/**/*.rs` across 11 crates, 2 in `.github/workflows/`, 1 in the `justfile`.

No REQ IDs were touched, so there is no spec-conformance impact.


specs/plans/2026-08-29-plans-as-ephemeral-artifacts.md
```diff
@@ -0,0 +1,221 @@
...
+
+The squashed commit contains the implementation and any ADR or design change,
+and no plan. The plan's add-commit remains viewable on the PR page — GitHub
+retains `refs/pull/*/head` permanently (1,354 such refs on this repository
+today), so the reviewed artifact survives without a manual archiving step.
+
+**Multi-PR work** (43 of 237 existing plans use phase/slice language, so
+roughly one change in five):
+
+- A GitHub **issue** holds the overall plan and slice breakdown. Its job is
```
garretfick: Using issues is intentional so that we cannot forget work.

garretfick: You're right, and I had this backwards. Fixed in e9409c8.

Issues are now a first-class step for every non-trivial change, not a multi-PR special case, and I removed the "not archival" framing that understated the point. The process now says:

- Every non-trivial change opens an issue first; the plan references it.
- Anything the plan described that the PR does not deliver goes back onto the issue **before** the plan file is removed.
- The issue closes only when nothing is outstanding.

This also closes a hole I'd left: under the previous text, single-PR work had no durable record at all once the plan was deleted, so undelivered work would have vanished silently.

Worth noting the current backlog is this exact failure mode in a quieter form — 1,074 unticked checkboxes across 108 plans, and 46 plans where no box was ever ticked. That work was written down, never done, and never surfaced again because nobody re-reads a merged plan. Deleting plans without issues would have made it silent rather than merely quiet.

---

PR #1475, opened 2026-08-29
Add scoping to rule_function_call_type_check

garretfick:
The sixth pass tracking scope by hand, recorded as out of scope by the
method-scoping plan. It hides two real defects: a method local leaks its
type to the next method, so an INT field assigned a REAL literal is
accepted; and a declaration's own name is never recorded, so assigning a
wrong-typed value to a result variable is unchecked for a FUNCTION as
much as for a METHOD.

Measured before planning: neither fix changes the outcome of any existing
test.

No prefactoring needed. The shape this needs was built by the three
earlier pull requests, so migrating a pass onto it is now a drop-in.

Claude-Session: https://claude.ai/code/session_01DdDr6NVSimqUjzZupHf3jE


compiler/analyzer/src/rule_function_call_type_check.rs
```diff
@@ -1602,4 +1620,180 @@ END_PROGRAM"
             &opts,
         ));
     }
+
+    // ---------------------------------------------------------------------
+    // METHOD scoping.
+    // See specs/plans/2026-08-29-call-type-check-scoping.md.
```
garretfick: Do not reference plans in docs. This is a new rule.

specs/plans/2026-08-29-call-type-check-scoping.md
```diff
@@ -0,0 +1,145 @@
+# Plan: Scope `rule_function_call_type_check`'s variable types
```
garretfick: Do not check in this file. Please remove it. If there are design or ADR decisions we need to keep, then let's discuss. Those don't belong in "plans" because we are getting rid of committed plan documents.

garretfick (commented):
see comments

compiler/analyzer/src/rule_function_call_type_check.rs
garretfick: Removed. The comment block now just names what the tests below cover.

---

specs/plans/2026-08-29-call-type-check-scoping.md
garretfick: Removed. The PR is now a single file, `rule_function_call_type_check.rs`.

Nothing that explains the change is lost with it: both defects are stated in the commit messages that fix them, and each reproduction is encoded as a test. No design or ADR decisions were in the file — it was a work plan, not a record of a decision.

Two follow-ups I'd rather raise than assume:

- `specs/steering/development-standards.md` on main still says non-trivial changes "must begin with an implementation plan committed to `specs/plans/`". That's what I was following. Worth updating alongside the citation-removal PRs so the standard matches the practice.
- The earlier merged PRs in this series left plan citations in code — `dsl/src/scope.rs`, `rule_use_declared_symbolic_var.rs`, `xform_fold_initializer_expressions.rs`, `xform_resolve_expr_types.rs`, `compile_method.rs` and `end_to_end_methods.rs`. Happy to strip those in a separate PR if you want them gone too.

---

garretfick:
The build failure was not this PR's — **`main` is red on its own.**

`Compiler Tests (ubuntu-latest)` and `Compiler Quality Checks` failed with 110 analyzer failures, all of them:

```
panicked at analyzer/src/scoped_table.rs:199:9:
scope stack unbalanced: ended at depth 2, expected 1
```

I reproduced it on `origin/main` with none of this branch's changes: 685 passed, 110 failed. The branch already contained current main, so rebasing would not have changed anything.

**Cause: two changes that landed independently and had not met until now.**

- #1443 added `rule_assignment_aggregate_type_compat`, which opens a scope at setup for declarations outside any POU and never exits it.
- #1463 added a debug assertion in `ScopedTable::drop` that the scope stack unwound, so a pass that opens a scope nothing closes fails the suite instead of silently mis-resolving later names.

The assertion is doing its job here, just not on the defect it was written for. `ScopedTable::new` has already opened the outermost scope — exactly the one that rule's comment describes — so the extra `enter()` served no purpose beyond leaving the table one frame deep at drop.

**Fix ported into this PR** (`7527ce0`): remove the redundant `enter()`. Behaviour-identical — globals land in the outermost scope either way and a lookup walks every frame. The rule's own 19 tests pass unchanged, the workspace is at 4354 passing / 0 failing, and `cd compiler && just` is green.

That one line is what `main` needs. It is included here so this PR can go green, and it no-ops once main carries it — but it is worth landing on main directly rather than waiting on this PR, since every other open PR is red for the same reason.

---

PR #1481, opened 2026-08-29
Render every assigned opcode in the container viewer

### LINKED ISSUE #1451: Container viewer renders 50 of 126 assigned opcodes as `UNKNOWN`, including `CALL` and all 64-bit and float arithmetic
## Summary

`decode_instructions` has a match arm per opcode and a catch-all `unknown` arm
that prints `UNKNOWN(0x..)`. 50 of the 126 opcodes that `opcode::is_assigned`
reports as assigned have no arm, so they fall to the catch-all. Any program
using a function call, `REAL`/`LREAL` arithmetic, or 64-bit integer arithmetic
disassembles with unreadable rows.

Byte offsets stay correct — #1446 replaced the per-arm `pc += <literal>` with a
single `pc += opcode::instruction_size(opcode_byte)`, and the catch-all no
longer advances by a hardcoded 1 — so this is a rendering gap, not a decode
desync. But it defeats the purpose of the viewer for the affected programs.

## Unhandled opcodes

**Call (5 bytes, `opcode + u16 function_id + u16 arg_base`)**

`CALL`

Only `FB_CALL`, `METHOD_CALL`, and `BUILTIN` have arms today, so a plain
function call is the most visible instance of this gap.

**64-bit integer arithmetic and comparison (1 byte)**

`ADD_I64`, `SUB_I64`, `MUL_I64`, `DIV_I64`, `MOD_I64`, `NEG_I64`,
`EQ_I64`, `NE_I64`, `LT_I64`, `LE_I64`, `GT_I64`, `GE_I64`

**Unsigned division and comparison (1 byte)**

`DIV_U32`, `MOD_U32`, `DIV_U64`, `MOD_U64`,
`LT_U32`, `LE_U32`, `GT_U32`, `GE_U32`, `LT_U64`, `LE_U64`, `GT_U64`, `GE_U64`

**Float arithmetic and comparison (1 byte)**

`ADD_F32`, `SUB_F32`, `MUL_F32`, `DIV_F32`, `NEG_F32`,
`ADD_F64`, `SUB_F64`, `MUL_F64`, `DIV_F64`, `NEG_F64`,
`EQ_F32`, `NE_F32`, `LT_F32`, `LE_F32`, `GT_F32`, `GE_F32`,
`EQ_F64`, `NE_F64`, `LT_F64`, `LE_F64`, `GT_F64`, `GE_F64`

**Typed stores (3 bytes, `opcode + u16 var_index`)**

`STORE_VAR_I64`, `STORE_VAR_F32`, `STORE_VAR_F64`

Note the asymmetry in the last group: `LOAD_VAR_I64`, `LOAD_VAR_F32`, and
`LOAD_VAR_F64` were added in #1446, but their matching stores were not — so a
`LREAL` assignment currently disassembles as a named load followed by an
`UNKNOWN`.

All 50 are emitted by codegen today (`grep -rn 'opcode::<NAME>' compiler/codegen/src`),
so none of this is dead wire-format surface.

## Expected

Every assigned opcode renders with its mnemonic and decoded operands.
`UNKNOWN(0x..)` should mean an unassigned byte — a corrupt container or a
container from a newer compiler — not a normal instruction.

## Notes

The bulk of the work is mechanical: 47 of the 50 are 1-byte, no-operand
instructions that need the same four-line arm as `ADD_I32`. That block of
near-identical arms is already the largest thing in the function; a table-driven
mnemonic lookup for the no-operand opcodes would collapse it and make the next
addition a one-line change rather than a five-line one.

The three that need real work are `CALL` (function id plus argument base, and
ideally the callee name resolved from the container the way `lookup_const_comment`
resolves pool entries) and the three typed stores (a `var[n]` operand each,
identical in shape to the existing `STORE_VAR_I32` arm).

## Preventing the next gap

The reason this drifted is that nothing fails when an opcode is added to
`instruction_size_opt` without a disassembler arm. There is an exact precedent
for the guard in `compiler/codegen/tests/it/wire_format.rs:726`:

```rust
#[test]
fn opcode_pins_when_compared_to_assigned_set_then_complete() {
```

which compares a pinned registry against `opcode::is_assigned` and fails with a
clear diff when the two diverge. An equivalent test for the disassembler —
decode a one-instruction buffer for every assigned opcode and assert the
mnemonic is not `UNKNOWN` — would make this class of gap impossible to
reintroduce, and would fail loudly the moment a new opcode lands without viewer
support.

## Relevant code

- `compiler/project/src/disassemble.rs:245` — `decode_instructions`
- `compiler/project/src/disassemble.rs:978` — the `unknown` catch-all arm
- `compiler/container/src/opcode.rs:871` — `instruction_size_opt`, the single
  enumeration of the opcode space
- `compiler/container/src/opcode.rs:865` — `is_assigned`, which a completeness
  test can drive
- `compiler/codegen/tests/it/wire_format.rs:726` — the completeness guard to model

Follow-up to #1446, which added many opcode arms under the title "add all op
codes to the container viewer" but did not reach the full set.

garretfick:
Fixes #1451.

## The gap, and why it kept reappearing

The opcode space was written down three times: a constant per opcode, a size
per opcode in `instruction_size_opt`, and a mnemonic-plus-operands per opcode
in `decode_instructions`. Nothing tied the third to the first two, so 50 of
the 126 assigned opcodes fell through the viewer's catch-all — `CALL`, all
64-bit and float arithmetic, unsigned division and comparison, and the
`STORE_VAR_I64`/`F32`/`F64` stores whose matching loads were already handled.
Worse, the size and the operand reads each encoded the layout independently,
so the two could disagree about where an instruction ends.

The issue proposes a completeness test. This goes further and removes the
duplicate enumerations, so the incomplete state cannot be written down.

## One declaration per opcode

A `declare_instruction_set!` row gives an opcode its name, its encoding, and
the operands that follow it:

```rust
/// Load a 32-bit integer constant from the constant pool.
/// Operand: u16 constant pool index (little-endian).
LOAD_CONST_I32 = (OP_CLASS_LOAD_CONST, T_I32) => [ConstIndex];
```

From that row come the opcode constant, `Instruction::decode` (mnemonic and
operand layout), and — because each operand's role fixes its width —
`instruction_size`, which is now derived rather than stated. `Operand` is the
vocabulary of operand roles; it and the macro live in a new `instruction`
module.

`decode_instructions` then has **no per-opcode arms at all** — it renders each
row from the instruction's declaration, which is why the diff deletes 738
lines and adds 340. Consequences:

- An assigned opcode cannot be missing from the viewer; there is no arm to
  forget. `UNKNOWN(0x..)` again means only "not an assigned opcode".
- The one remaining extension point is `format_operand`, an exhaustive match
  on `Operand` with no catch-all: a **new operand shape fails to compile**
  until it is rendered.
- Size and rendering read the same layout, so they cannot desync.

Operand reads are now bounds-checked: a container truncated mid-instruction
renders `<truncated>` instead of panicking.

## Behaviour changes

Three renderings change, all deliberate:

| | before | after |
|---|---|---|
| `CMP_BR_*` constant | `const[2]` | `pool[2]`, with the value as a comment |
| `METHOD_CALL` | `fields[7..+2]` | `fields[7], num_fields: 2` |
| `CALL` | `UNKNOWN(0x84)` | `func[2], params[3]`, plus the callee's name from the debug section |

Disassembly of a program with an `LREAL` function call and 64-bit arithmetic,
which the viewer previously filled with `UNKNOWN` rows:

```
    0  LOAD_VAR_I64       var[2]
    3  LOAD_CONST_I64     pool[2]      = 3
    6  ADD_I64
    7  STORE_VAR_I64      var[2]
   17  LOAD_CONST_F32     pool[4]      = 4
   20  BUILTIN            0x038E
   23  CALL               func[2], params[3]   = scale
   36  JMP_IF_NOT         offset: +0x0006      -> 0x002D
```

## Also here

The operand table pushed `opcode.rs` eight lines past the 1000-line module
limit, so `cmp_op` moves out beside `builtin` and `fb_type` (from #1478) —
same kind of module, a set of IDs addressed through an operand rather than
part of the opcode encoding. Re-exported at its existing path, so no call site
changes; `opcode.rs` ends at 963 lines.

## Not in scope

`BUILTIN 0x038E` above shows the same defect one level down: the viewer names
45 of the 107 built-in IDs by hand and prints the rest as bare hex. The same
macro treatment fixes it; it follows separately, and here the hand-written
list is confined to `format_builtin` so that change is a body swap.

## Verification

- No opcode byte moved: the `(op_class, type_tag)` pair of all 126 constants
  was extracted before and after and diffed — identical — and every derived
  size matches what the old size match gave it.
- New guard tests, both looping over every byte `opcode::is_assigned` accepts:
  each renders as its own mnemonic (never `UNKNOWN`), and each consumes
  exactly its own bytes so the following row starts at the right offset.
- Named cases for the previously-unhandled opcodes, the `CALL` callee-name
  comment, and the truncated-instruction path.
- `cd compiler && just` passes in full on the merged branch: compile, coverage
  (all tests), clippy, rustfmt, and the duplication gate.


specs/plans/2026-08-29-opcode-table-driven-disassembly.md
```diff
@@ -0,0 +1,132 @@
+# Table-Driven Disassembly of the Opcode Space
```
garretfick: Remove the plan document entirely.

garretfick: Removed in `dedc4b3a`. Its content stays reviewable in `c3225275`, the commit that added it, per the process in #1456.

I also resolved the merge conflict while here — the PR was `dirty` against main.

The conflict was with #1478, which moved the builtin and fb_type IDs out of `opcode.rs`. Both sides added imports at the same point in the file and neither touched the other's logic, so the resolution keeps both: this branch's `declare_instruction_set` import and `Instruction`/`Operand` re-export, and main's `builtin`/`fb_type` re-exports.

Keeping main's side was necessary, not just tidy — `disassemble.rs` still refers to `opcode::builtin::EXPT_I32` and the rest of that list in `format_builtin`, so dropping the re-export would have failed to compile. Worth flagging against the PR description's claim that the two are independent and "merge cleanly in either order (verified by test-merge)": they conflict textually, though the resolution is mechanical.

One thing I have left alone for you to decide. `opcode.rs` is now **1008 lines**, not the ~990 the description predicted once #1478 landed — the estimate didn't account for the doc comments on #1478's re-exports. That is over the 1000-line guideline in `CLAUDE.md`, but the guideline is not enforced anywhere (no `clippy.toml`, no module-size test), so CI is unaffected. Bringing it under would mean moving code, which seemed beyond resolving a conflict.

`cd compiler && just` passes on the merged tree: compile, coverage, clippy, rustfmt, and the duplication gate (7.4% exact / 2.8% near).

---

garretfick: Removed in dedc4b3 — the branch already carried that commit and the merge of main when I picked this up, so nothing further was needed here.

---

garretfick:
Merged `main` (4504835). The conflict was #1443's `COPY_REGION`, which is exactly the case this PR reshapes, so it is worth saying how it resolved:

- `COPY_REGION` becomes a table row — `(OP_CLASS_STORE_ARRAY, 1) => [VarIndex, ArrayDescIndex, ArrayDescIndex]` — carrying its doc comment unchanged. Its `instruction_size_opt` arm (`COPY_REGION => 7`) and its `decode_instructions` arm both go away: 1 + 2 + 2 + 2 = 7 falls out of the layout, and the row renders itself.
- Its row comment, *"size from descriptors; source offset from stack"*, says something the operands cannot. Rather than drop it, a row can now carry an optional `note`, which the viewer renders as the row's comment. That keeps it a property of the one declaration instead of a fact restated in the viewer.
- `decode_instructions_when_copy_region_then_decodes_and_stays_aligned` passes unchanged — same operands, same offsets. I did update its doc comment, which described the old per-arm design ("without an arm the instruction renders as `UNKNOWN`"), since there are no arms now.

**CI will be red, and it is not this PR.** `origin/main` at e879a64 (#1443) panics in `analyzer/src/scoped_table.rs:199` — `scope stack unbalanced: ended at depth 2, expected 1`. I ran the full suite on a clean checkout of that commit and on this branch: **the same 110 tests fail on both, and the sets are identical** — so this branch adds no failure and fixes none. There is no fix on `main` yet to port; the change is #1443's to make. This PR's own tests are green (container 212, project/disassemble 96, codegen 207), as are clippy and rustfmt.

---

PR #1488, opened 2026-08-30
Enforce plan deletion before merge with plan-citations check

garretfick:
## Summary

Implements enforcement that implementation plans in `specs/plans/` are never cited from code, workflows, design documents, or ADRs. Plans are branch-local artifacts committed for review and deleted before merge, so they must never be stable reference targets.

## Changes

- **Added `plan-citations` CI check** (`compiler/justfile`): New recipe that fails the build if any file outside `specs/plans/` references the plans directory. Integrated into the default CI pipeline.

- **Updated development standards** (`specs/steering/development-standards.md`): 
  - Clarified that plans are branch-local, committed as the first commit on a feature branch, and deleted before merge
  - Explained the squash-merge behavior that cancels the add/delete, keeping plans viewable on PRs but not in `main`
  - Documented the complete planning workflow: issue → plan → implement → land decisions → delete plan
  - Added historical rationale for why plans are not kept (237 files with 92 stale references)

- **Updated contributor guides** (`CONTRIBUTING.md`, `CLAUDE.md`, `CURSOR.md`):
  - Revised workflow to include opening an issue first
  - Added explicit step to `git rm` the plan file before merge
  - Clarified that decisions worth keeping must land as ADRs or design docs in the same PR
  - Added note about `just plan-citations` enforcement

- **Updated steering rules** (`.cursor/rules/ironplc-steering.mdc`):
  - Added rule that plans must be deleted before merge
  - Added rule that `specs/plans/` must never be cited from code, workflows, or docs

- **Updated guidance** (`specs/steering/steering-file-guidelines.md`):
  - Clarified that implementation plans are reviewed on the branch and deleted before merge

## Implementation Details

The `plan-citations` check uses `git grep` to search for references to `specs/plans/` while excluding:
- The `specs/plans/` directory itself
- Process documents (`CLAUDE.md`, `CURSOR.md`, `CONTRIBUTING.md`) that describe the rule
- Steering files (`.cursor/`, `specs/steering/`) that document the process
- The `justfile` itself (which documents the rule)

This ensures the rule is self-documenting while preventing accidental citations that would become dead links after merge.

https://claude.ai/code/session_016RWos6WEczkF7kDTLgLSRY

PR #1491, opened 2026-08-30
Fix method scope handling in aggregate type compatibility check

garretfick:
## Summary

Refactor the aggregate type compatibility rule to properly handle method scopes using the new `enter_scope`/`exit_scope` visitor pattern. This fixes a bug where method-local variables were incorrectly sharing the enclosing function block's declaration frame, causing type mismatches to be silently accepted.

## Changes

- **Consolidate scope handling**: Replace three separate visitor methods (`visit_function_declaration`, `visit_function_block_declaration`, `visit_program_declaration`) with a single `enter_scope` method that handles all declaration types (Function, FunctionBlock, Program, and Method) uniformly.

- **Add `exit_scope` method**: Implement the corresponding scope exit handler to properly close declaration frames.

- **Add test coverage**: Include a new test case (`apply_when_method_local_shadows_field_then_sibling_method_uses_field_type`) that verifies method-local variables no longer shadow function block fields across method boundaries, ensuring aggregate assignment type mismatches are properly detected.

- **Update imports**: Add `ScopeNode` import to support the new scope-aware visitor pattern.

## Implementation Details

The refactoring leverages the exhaustive `ScopeNode` enum to ensure all scope-creating declaration types are handled consistently. The match statement on `ScopeNode` variants guarantees that any new scope type added in the future will require explicit handling here, preventing silent bugs from scope handling omissions.

The test demonstrates the fix by defining a function block with two methods: the first method declares a local variable `v` of type `Other` (shadowing the field `v` of type `Pt`), and the second method uses `v` expecting the field type. Before this fix, the second method's assignment would incorrectly use the first method's local type; now it properly detects the type mismatch.

https://claude.ai/code/session_01DdDr6NVSimqUjzZupHf3jE

### DIFF
--- a/compiler/analyzer/src/rule_assignment_aggregate_type_compat.rs
+++ b/compiler/analyzer/src/rule_assignment_aggregate_type_compat.rs
@@ -46,6 +46,7 @@ use ironplc_dsl::{
     common::*,
     core::{Id, Located},
     diagnostic::{Diagnostic, Label},
+    scope::ScopeNode,
     textual::*,
     visitor::Visitor,
 };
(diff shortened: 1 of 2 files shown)


garretfick:
Merged main in and validated locally before CI ran. Two notes.

**`main` is red on `just plan-citations`, independently of this PR.**

```
error: specs/plans/ is cited outside specs/plans/ itself.
compiler/codegen/tests/it/end_to_end_array_of_struct.rs:4
```

Reproduced on `origin/main` with none of this branch's changes. #1415 added that doc comment citing its plan; #1488 added the check that forbids plan citations. They landed independently and hadn't met until now — the same shape as the unbalanced scope stack in #1487.

Fix ported here (`b4855ef`): the sentence already cites issue #1383, a stable reference target, so only the plan reference is dropped. That one line is what main needs; it no-ops once main carries it.

Worth noting separately: `specs/plans/2026-08-22-top-level-array-of-struct.md` is still checked in. The check only looks for *citations*, so its presence doesn't fail anything, but under the policy #1488 encodes it should have been deleted before its PR merged. I left it alone rather than delete another PR's document.

**On the description above** — it was generated from the diff and slightly undersells the change. This isn't only a refactor: the migration fixes a real defect. On main, this program exits 0:

```st
FUNCTION_BLOCK FB_Motor
VAR
    v : Pt;
    src : Other;
END_VAR
METHOD A
VAR
    v : Other;
END_VAR
    v := src;
END_METHOD
METHOD B
    v := src;
END_METHOD
END_FUNCTION_BLOCK
```

In `B`, `v` is the function block's `Pt` field, so `v := src` assigns an `Other` to it. Method `A`'s local left `v` recorded as `Other`, hiding the mismatch. It now reports P2037 — verified with the real compiler before and after, and encoded as the new test.

This is the seventh and last pass to move off hand-rolled scope tracking.

`cd compiler && just` is green on the merged tree: 4401 tests, all steps including the new `plan-citations` check.

---

PR #1503, opened 2026-08-30
Add plan for folding TRUNC that follows a constant load

garretfick:
Codegen emits an explicit TRUNC_* before every store to a sub-32-bit
variable (ADR-0001). When the value on the stack came from a constant
load, that instruction is doing work the compiler could have done: the
constant either already fits the narrow type, or its truncated value is
known at code generation time.

The plan proposes recognizing LOAD_CONST_I32 + TRUNC_* in the existing
peephole optimizer, deleting the TRUNC when the constant is in range and
rewriting the load's pool operand to the truncated value when it is not.

Claude-Session: https://claude.ai/code/session_016Q6DoHpAJEiEniXdnnMjpo

### DIFF
--- a/specs/plans/2026-08-30-constant-truncation-folding.md
+++ b/specs/plans/2026-08-30-constant-truncation-folding.md
@@ -46,16 +46,22 @@ The result is visible in the existing tests, which document the exact sequence:
 
 Compiling a representative program (structure with `INT`/`BYTE`/`SINT` fields, a
 function block with narrow locals, a `PROGRAM` with narrow scalar, struct-field
-and array-element assignments, and a `FOR` loop) and scanning the resulting
-`.iplc` for the four `TRUNC_*` opcodes: **18 of 19 occurrences are immediately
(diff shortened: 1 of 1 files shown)


garretfick:
Closing in favour of #1697, which carries this content as a design document rather than a plan.

The work this plan proposed has since shipped — `optimize/` is split into the pass pipeline, `pass_const_trunc` is in place with its tests, and the `vm-performance.md` §13 note was updated. What was left was the durable half, so #1697 adds `specs/design/bytecode-peephole-optimizer.md` with the offset-map contract, the rewrite-engine contract and each pass's soundness argument, backed by fourteen `REQ-PEEP-codegen-*` requirements.

One item from the task list here had never landed — the driver test pinning that the passes may run in any order. #1697 adds it rather than leaving it behind.

Both follow-ups listed at the bottom of this plan already have issues, so nothing needed filing: #1527 (closed by #1611 — `P2026` now rejects an out-of-range constant on the numeric narrow types) and #1529 (open; its reproduction is stale as a result, and now has a comment with the current one).

---

Issue #1509, opened 2026-08-30
Codegen emits a runtime TRUNC after loading a constant that is already narrow


garretfick:
Per ADR-0001, sub-32-bit integer types are promoted to 32 bits on load and codegen emits an explicit `TRUNC_*` before any store back to a narrow slot. `emit_truncation` (`codegen/src/compile_expr.rs`) always emits the opcode, because at that point it knows the target's type but not whether the value on the stack is a compile-time constant.

`compile_constant` cannot compensate: it receives an `OpType` of `(OpWidth, Signedness)`, which collapses `SINT`/`INT`/`DINT` into `(W32, Signed)`. The `storage_bits` that selects the truncation width lives one level up in `VarTypeInfo`. So the constant is emitted at 32 bits and narrowed by a runtime instruction that could have been resolved at compile time.

### Measurement

Compiling a representative program (structure with `INT`/`BYTE`/`SINT` fields, a function block with narrow locals, a `PROGRAM` with narrow scalar, struct-field and array-element assignments, and a `FOR` loop) and disassembling the result: **16 of 17 `TRUNC_*` instructions are immediately preceded by `LOAD_CONST_I32`**, and all 16 constants are already in range, so all 16 are pure deletions. The single survivor is `arr[i] := i`, where the value genuinely is computed at run time.

Struct-field initialization alone emits `LOAD_CONST; TRUNC; LOAD_CONST idx; STORE_ARRAY` for every narrow field of every structure.

### Proposed change

Recognize `LOAD_CONST_I32 p; TRUNC_*` in the peephole optimizer (`codegen/src/optimize.rs`), which already runs per function with access to the constant pool:

- if the constant already fits the narrow type, delete the `TRUNC_*`;
- otherwise apply the truncation at compile time, intern the result, and rewrite the load's pool operand.

The VM is unconditionally wrapping, so both cases preserve behaviour exactly.

### Slices

- [ ] **Slice 1 — prefactor.** Restructure the optimizer into an ordered pipeline of named passes over shared rewrite machinery, modelled on `analyzer/src/stages.rs`. Behaviour-preserving. The current `is_removable_pair -> bool` plus remove-both driver cannot express "remove one instruction and rewrite the other's operand", and `optimize.rs` is at 833 lines against the 1000-line module limit.
- [ ] **Slice 2 — the fold.** Add the truncation pass, widen the pool to `&mut` so an out-of-range constant can be interned, and update the affected bytecode-structure tests.

### Related

[...]

PR #1512, opened 2026-08-30
Refactor optimizer into pass pipeline architecture

garretfick:
## Summary

Restructure `compiler/codegen/src/optimize.rs` from a single 833-line module with three interleaved rewrites into an ordered pipeline of named passes over shared rewrite machinery. This is a behaviour-preserving refactor that prepares the optimizer for a fourth rewrite pattern (constant truncation folding) planned in a follow-up PR.

## Changes

- **Split `optimize.rs` into a module hierarchy:**
  - `optimize/mod.rs` — driver, pass sequencing, offset-map composition, public API
  - `optimize/rewrite.rs` — shared machinery (`Instruction`, `decode`, jump-target collection, `apply_peephole`, branch-offset rewriting)
  - `optimize/pass_self_assign.rs` — `LOAD_VAR x; STORE_VAR x` removal
  - `optimize/pass_arith_identity.rs` — `LOAD_CONST 0; ADD|SUB` and `LOAD_CONST 1; MUL|DIV` removal
  - `optimize/tests.rs` — existing test suite (unchanged)

- **Introduced `apply_peephole` helper** that encapsulates the decode-mark-rebuild-rewrite cycle, taking a matcher closure over two-instruction windows. Each pass declares its own parameters (e.g., `pass_self_assign::apply(bytecode)` vs `pass_arith_identity::apply(bytecode, constants)`), following the pattern used by analyzer transforms.

- **Pipeline threads bytecode through passes** in order, accumulating a single old→new offset map that covers the entire sequence. No pass currently exposes work for another, so any order produces identical bytecode today; the pipeline structure makes this a stated property with one place to record ordering constraints.

- **Verified no cascading optimizations needed:** identity removals cannot expose work for truncation folding because the analyzer folds literal-op-literal expressions before codegen, and named constants compile to variable loads rather than constant loads.

## Implementation Details

- Jump-target protection and branch-offset rewriting logic moved to `rewrite.rs` and reused by all passes via `apply_peephole`.
- Offset-map composition in the pipeline driver ensures callers see a single map covering the entire optimization sequence.
- Test suite moved to `optimize/tests.rs` with no assertion changes; all 30 tests pass unchanged.
- Module size reduced from 833 lines to ~140 (mod.rs) + ~160 (rewrite.rs) + ~35 (pass_self_assign.rs) + ~85 (pass_arith_identity.rs) + ~600 (tests.rs), staying well under the 1000-line limit and leaving room for the fourth pass.

## Rationale

This refactor addresses three steering signals at once:
1. The new truncation-folding rewrite removes one instruction and rewrites an operand (not a pair removal), which the current `is_removable_pair(a, b, constants) -> bool` signature cannot express.
2. The new rewrite needs mutable access to the constant pool (to intern truncated values); one shared predicate signature would require widening it for every pattern.
3. The module is at 833 lines; a fourth pattern plus tests would exceed the 1000-line limit.

The pipeline architecture, modelled on `analyzer/src/stages.rs`, makes the pass sequence explicit and extensible without further refactoring.

https://claude.ai/code/session_016Q6DoHpAJEiEniXdnnMjpo


specs/plans/2026-08-30-optimizer-pass-pipeline.md
```diff
@@ -0,0 +1,140 @@
+# Optimizer pass pipeline (prefactor)
```
garretfick: This should not be committed.

compiler/codegen/src/optimize/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+/// of the function can still be remapped.
+pub(crate) type OffsetMap = HashMap<usize, usize>;
+
+/// Remaps an emitter line map through the optimizer's old→new offset table.
+///
+/// For each entry the old `bytecode_offset` is looked up in `offset_map`,
+/// which by construction maps every instruction boundary — including
+/// removed instructions — to the position the next surviving instruction
+/// occupies in the optimized stream ("snap forward"). Entries whose
+/// remapped offset lands past the new end-of-function (`new_bytecode_len`)
```
garretfick: Why would we drop? Isn't that a critical error?

compiler/codegen/src/optimize/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+/// are dropped, and entries that map to the same remapped offset as the
+/// previous kept entry are deduplicated (the optimizer can collapse two
+/// pre-optimization positions onto one post-optimization position).
+///
+/// Entries whose old offset is not in the map are dropped silently; the
```
garretfick: Why silently drop? If this should never happen, can we have this return a diagnostic because this means we hit a compiler defect. This would be an internal error.

specs/plans/2026-08-30-optimizer-pass-pipeline.md
garretfick: Removed in 6d5e0e5.

For context on why it was there: CLAUDE.md asks for the plan as the first commit on the branch so it is reviewable as a file diff, then `git rm`'d before merge. I was holding it for you to read and would have removed it before merge — but there is no reason to make you ask, so it is gone now. The content it carried that outlives the branch is on #1509.

---

compiler/codegen/src/optimize/mod.rs
garretfick: Agreed — fixed in 6d5e0e5.

The offset map covers every instruction boundary plus one-past-the-end, and the emitter records an entry only immediately before pushing an opcode. So a miss means an entry sits at an offset that is not an instruction start, which no input program can cause — it is a compiler defect. `remap_line_map` now returns `Result` and reports `P9998` via `Diagnostic::internal_error_at`, matching how `stack_balance::verify_container` and `call_graph::compute_max_call_depth` report their own invariant violations.

`finalize_function` returns `Result<FinalizedFunction, Diagnostic>` and the five call sites take `?`; all were already inside functions returning `Result<_, Diagnostic>`, so nothing else moved. Test added at `optimize/tests.rs` asserting the code and message.

Worth noting these two threads are on code I moved verbatim from the old `optimize.rs` — it reads as new because the file is new. The defect is pre-existing, not introduced here.

---

compiler/codegen/src/optimize/mod.rs
garretfick: These two are different from the one below, and I don't think either is an error — but the comment did a poor job saying so, which is why they read as suspicious. Reworded in 6d5e0e5.

**Past the new end of the function.** Every instruction the entry could point at was removed, so there is no surviving instruction left to attribute the source position to. There is nothing to record, and keeping the entry would put a debugger on an offset outside the function. This needs the whole tail of a function to be removed, so in practice it is a guard rather than something that fires — a function ends in `RET`/`RET_VOID`, which no pass removes.

**Same offset as the previous kept entry.** This one is routine, not exceptional. The optimizer collapses several source positions onto one surviving instruction, and the emitter's store-load peephole does it on ordinary code — there is an existing `emit.rs` test noting "line 2 was elided by the peephole so no entry exists for it". A line map holds one position per offset, so the duplicate has nowhere to go.

The genuinely suspicious case was the third one, which is now a `P9998`.

---

PR #1533, opened 2026-08-30
Prefactor the analyzer and codegen for constant range checking

### LINKED ISSUE #1527: Out-of-range constant assigned to a narrow integer type wraps silently
A constant that does not fit its narrow target type is accepted without any diagnostic and silently wraps.

## Reproduction

```iecst
PROGRAM main
  VAR
    a : USINT := 300;   (* stores 44 *)
    b : USINT;
    c : SINT;
  END_VAR
  b := 300;             (* stores 44 *)
  c := 200;             (* stores -56 *)
END_PROGRAM
```

Compiles with exit 0 and no output. Both forms are affected — the variable initializer and the assignment statement.

`USINT` is 0..255 and `SINT` is -128..127, so none of these three values is representable. The compiler wraps them per the VM's two's-complement `TRUNC_*` semantics.

## Why this is worth a diagnostic

`c : SINT; c := 200;` is far more likely a typo or a misunderstanding of the type's range than a deliberate request for `-56`. The wrap is invisible: nothing in the source says the value changes, and IEC 61131-3 gives no reason to expect a literal to mean something other than what it says.

Contrast the existing behaviour for a literal that does not fit its *arithmetic* width, which is already diagnosed:

```
error[P0022]: Constant is out of range
```

`Problem::ConstantOverflow` is raised in `compile_constant` (`codegen/src/compile_expr.rs`) when a literal exceeds `i32`/`u32`/`i64` range. So the compiler already has the concept and a problem code for it; the narrow types simply are not checked, because `compile_constant` receives an `OpType` — a width and a signedness — and `SINT`/`INT`/`DINT` all collapse to `(W32, Signed)`. The `storage_bits` that bound the real range live one level up in `VarTypeInfo`.

## Scope note

Deliberate wrapping is a legitimate thing to want on the bit-string types. `BYTE`/`WORD` are bit patterns rather than numbers, and an existing test asserts the wrap:

```rust
// codegen/tests/it/end_to_end_bitstring.rs
e2e_i32!(
    end_to_end_when_byte_overflow_then_wraps,
    "PROGRAM main VAR x : BYTE; END_VAR x := 255 + 1; END_PROGRAM",
    &[(0, 0)],
);
```

So a check that fires on `SINT`/`USINT`/`INT`/`UINT` but not on `BYTE`/`WORD` may be the right split, and would leave that test standing. That is a design decision, not an obvious one — worth settling against IEC 61131-3 before implementing.

## Relationship to #1509

#1509 (folding `LOAD_CONST_I32; TRUNC_*` at compile time) moves *when* the wrap happens from run time to compile time, but deliberately preserves the result exactly: `x : USINT := 300` stores 44 before and after. This issue is about whether the wrap should be diagnosed at all, which is a user-visible language decision and needs its own problem code and documentation.

garretfick:
Prefactoring for #1527, which this does not fix. Each commit deletes a copy of something rather than adding a layer, and the two behaviour changes are separated from the two extractions.

## Commits

**Extract variable type resolution into a shared module.** Walking a variable reference to its `IntermediateType` — through field accesses, subscripts and dereferences — lived inside `rule_bit_access_range` where only that rule could reach it. It moves to `analyzer/src/variable_type.rs`. No behaviour change; the rule's tests pass untouched.

**Derive codegen's type info from the analyzer's type table.** `ELEMENTARY_TYPES_LOWER_CASE` already states that an `INT` is a signed 16-bit integer, and codegen restated all of it in a 24-arm match producing `VarTypeInfo`. Codegen now reads that table and projects it, with the operation width *computed* from the value width rather than written out per type — so a value that fits a type cannot fail to fit the width it is operated on, by construction. `BOOL` stays explicit: it holds one bit of value in a byte of footprint, which the analyzer's byte-granular size cannot state. The projection moves to `codegen/src/type_info.rs`, which keeps `compile_setup.rs` under the 1000-line guideline. No behaviour change; every elementary type's projection is pinned by a test.

**Name the bit access rule for everything it checks.** It checks byte, word, dword and lword slices as well as bits, and has since partial access landed, but its name said only bit access and its header examples showed only `myByte.8`. Renamed to `rule_bit_and_partial_access_range`, with a passing and a failing example of each form. The slices had no tests of their own, so their two failure modes — an index past the last slice, and a slice wider than the variable — now have cases.

**Resolve declarations through the analyzer's scoped table.** The extracted declaration map was the flat per-POU `HashMap` the rule already carried, which is a third copy of something the analyzer has: six passes drive `ScopedTable` through the `enter_scope`/`exit_scope` hooks, and `rule_assignment_aggregate_type_compat` already stored exactly this value in one. It now uses that, sharing the `Declared` value with the rule that had its own.

## Two bugs this fixes

A name now resolves against every open scope rather than only the enclosing POU's own variables, so two cases the flat map dropped are checked:

- a bit access on a `VAR_GLOBAL` (`g.8` where `g : BYTE`)
- a bit access on a method local

Both were silently accepted before and report P4025 now. I confirmed the first was genuinely missed rather than assuming it — stashed the change, rebuilt, ran the program through `ironplcc check` for exit 0, then restored for P4025. Both have tests.

## Not here

The rule itself (`rule_constant_range`) and its `P2026` documentation, which is what closes #1527. The plan for it is deliberately not committed on this branch; #1527 is the record across the series.

Two adjacent gaps found and left alone, both user-facing:

- `docs/reference/compiler/problems/P4025.rst` documents bit access only, though the rule raises P4025 for slice failures too.
- P4025's message reads "Bit access index is out of range for the variable type", which is also what a `%W0`-on-a-`BYTE` failure prints.

`cd compiler && just` passes.

https://claude.ai/code/session_01UtwxiCCLtHgKgAurqYKvjD


specs/plans/2026-08-30-constant-out-of-range-analysis.md
```diff
@@ -0,0 +1,147 @@
...
+
+Once the analyzer checks a constant against its declared type, codegen cannot
+legitimately meet one that does not fit: every integer type is compiled at an
+`OpWidth` at least as wide as the type's own storage, so a value that fits the
+declared type fits the width. Reaching `compile_constant` with a value that
+does not fit therefore means the analyzer had a gap — a defect in the
+compiler, not in the program. Those sites change from `P2026` to
+`Problem::InternalError` (`P9998`).
+
+The risk this carries is that a gap in the rule turns a bad program into
```
garretfick: I suggested making this internal error, but I'm concerned we'll eventually have a defect because they are different code paths. We need some way so that this in not possible. That's likely the prefactor.

garretfick (commented):
see my comments

specs/plans/2026-08-30-constant-out-of-range-analysis.md
garretfick: Agreed, and the plan leaned on an argument rather than a mechanism. Revised in edeb8be to join the two paths at both seams.

**The type facts become one table.** `ELEMENTARY_TYPES_LOWER_CASE` in the analyzer already says `INT` is `Int { size: B16 }`. Codegen restates all of it in `resolve_type_name` as a hand-written match producing `VarTypeInfo { op_width, signedness, storage_bits }` — that is the copy that can drift. It goes: codegen derives `VarTypeInfo` from the analyzer's `IntermediateType`, with `op_width` computed from the storage size instead of written out per type. "The operation width is never narrower than the storage the range was computed from" then holds by construction, with a property test over the table. This is prefactor 2 and lands before any new behaviour — it deletes ~100 lines from codegen.

**The type each literal was checked against is recorded on the expression.** One pass decides what type a constant is stored or compared as and writes it onto the node — the same channel `xform_resolve_expr_types` already uses so that codegen reads types rather than re-inferring them. The rule checks against the recorded type; codegen compiles the recorded type. One enumeration of contexts, two readers.

That second seam is also what makes the internal error honest, which I think was the real worry. A constant carrying a recorded type has been checked, so a value that does not fit is the compiler contradicting itself — `InternalError`. A constant with no recorded type is one the analyzer could not type, so codegen keeps reporting `P2026` to the user. A gap in the pass degrades to today's behaviour, never to "internal error" on a program that is merely wrong.

Cost is a field on `Expr` and a new pass. The alternative — a span-keyed side table on `SemanticContext` — avoids the DSL change but keys types by span instead of by node, and nothing else in the compiler passes types that way; it is noted at the bottom of the plan in case you prefer it.

`INT#40000` is now https://github.com/ironplc/ironplc/issues/1545 and out of this plan.

---

specs/plans/2026-08-30-constant-out-of-range-analysis.md
```diff
@@ -0,0 +1,188 @@
+# Diagnose out-of-range constants during analysis
+
+Implements the fix for
```
garretfick: Do not commit the plan.

PR #1548, opened 2026-08-30
Validate scanLimit before VM starts, reject zero/negative values

garretfick:
## Summary

Reject invalid `scanLimit` values (zero, negative, fractional) in the `launch` request before the VM starts, rather than silently reinterpreting them or failing with confusing error messages. This prevents the session from starting when the bound is unsatisfiable, and clarifies that "unlimited" is spelled by omitting `scanLimit` entirely, not by using sentinel values like `0` or `-1`.

## Changes

- **Validation moved earlier**: `launch::check_scan_limit()` now validates the `scanLimit` argument before the VM is built, so invalid values fail the `launch` request itself (error V6011) rather than being silently accepted or causing confusing downstream errors.

- **Type change for type safety**: `LaunchRequestArguments::scan_limit` now holds `Option<Number>` (the raw JSON number) instead of `Option<u64>`, allowing validation to report exactly what the client sent. This also prevents the entire argument parse from failing on a negative value, which would incorrectly report a missing `program`.

- **NonZeroU64 downstream**: The validated bound is now `Option<NonZeroU64>`, making zero unrepresentable in the session logic and eliminating the need for runtime checks.

- **New error code V6011**: Documents the validation rule: `scanLimit` must be a whole number of at least 1 scan cycle, or omitted for an unbounded run. Zero is explicitly rejected (not reinterpreted as unlimited) because the bound is tested after a scan completes, so the shortest run the debugger can produce is one cycle.

- **Documentation updates**: 
  - `docs/reference/runtime/problems/V6011.rst` — new problem code documentation explaining why zero is rejected and how to request an unbounded run
  - `docs/reference/editor/debugging.rst` and `docs/reference/runtime/ironplcvmd.rst` — clarified that `scanLimit` must be ≥1 or omitted
  - `specs/design/debugger-support.md` and `integrations/vscode/package.json` — updated schema to reflect `integer` type with `minimum: 1`

- **Tests**: Added three new tests covering zero, negative, and omitted `scanLimit` cases to verify the session behavior matches the new validation rules.

## Implementation Details

The validation happens in `load_and_check()` before `launched_session()` is called, so an invalid bound never reaches the VM startup code. The `scan_limit_reached()` function now works with `NonZeroU64`, eliminating the possibility of a zero bound slipping through.

https://claude.ai/code/session_01FTQEg8q4RUHvJ9W6W5yfdv


docs/reference/runtime/problems/V6008.rst
```diff
@@ -4,13 +4,91 @@ V6008
...
+   * - ``stopOnEntry``
+     - ``true`` or ``false``. Defaults to ``false``.
+   * - ``scanLimit``
+     - A whole number of at least ``1``. Omit the argument to run without a
+       bound.
+   * - ``freewheelingIntervalMs``
+     - A cycle time in milliseconds, for a program whose task declares no
+       ``INTERVAL``. Defaults to ``100``.
+
+Why ``scanLimit`` Has No "Unlimited" Value
```
garretfick: Don't explain whee in the reference.

garretfick: Removed. The "Why `scanLimit` Has No Unlimited Value" section is gone, along with the sentences above the argument table explaining why one code covers every argument, and the aside in `debugging.rst`. That rationale already lives in `specs/design/debugger-support.md`, so nothing was lost.

The page keeps only what a reader needs at the point of failure: what each argument accepts, that `scanLimit` has no unlimited value, and what to write instead.

`specs/steering/doc-standards.md` arrived on this branch with the last `main` merge and says the same thing — reference describes the machinery, design decisions go to explanation — so I also moved the two remaining statements to active voice to match its writing style.

---

PR #1555, opened 2026-08-30
Clarify dialect presets and fix Dialect Display formatting

garretfick:
## Summary

Updates the Syntax Support Guide to clarify how dialect presets are configured and fixes a formatting bug in the `Dialect` Display implementation.

## Key Changes

- **Documentation clarification**: Updated `syntax-support-guide.md` to explain that dialect presets are now configured via tags in the `define_compiler_options!` macro rather than manual `from_dialect()` arms. This reflects the current code generation approach and reduces the risk of missing dialects when adding new flags.

- **RuSTy dialect documentation**: Corrected the table entry for the RuSTy dialect to document that it does not support `REFERENCE TO`, `POINTER TO`, or `ADR()` syntax (these are TwinCAT-specific).

- **Common mistakes section**: Updated the guidance to reference the new `define_compiler_options!` tagging approach instead of the old manual `from_dialect()` pattern.

- **Dialect Display fix**: Changed `Dialect::fmt()` to use `f.pad()` instead of `f.write_str()` so that format specifiers like `{:<20}` (width padding) work correctly. The previous implementation bypassed the formatter's width settings, causing alignment directives to be silently ignored.

## Implementation Details

The Display fix ensures that when a `Dialect` is formatted with width specifiers in format strings, the padding is properly applied. This is important for CLI output alignment and any other formatted display of dialect names.

https://claude.ai/code/session_016RWos6WEczkF7kDTLgLSRY

### DIFF
--- a/specs/steering/syntax-support-guide.md
+++ b/specs/steering/syntax-support-guide.md
@@ -243,7 +243,7 @@ pub fn insert_keyword_statement_terminators(
 
 ## Non-Standard Syntax Gating (`--allow-x` Flags)
 
-**Rule**: Anything not in the IEC 61131-3 standard **must** be gated behind an `--allow-x` flag. `--dialect=codesys` currently enables the broadest set; `ironplcc dialects` prints what each one turns on.
+**Rule**: Anything not in IEC 61131-3 Edition 2 **must** be gated behind an `--allow-x` flag. That includes Edition 3 syntax, which is gated the same way.
 
 ### Before Creating a New Flag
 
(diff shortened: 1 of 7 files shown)


specs/steering/syntax-support-guide.md
```diff
@@ -243,7 +243,7 @@ pub fn insert_keyword_statement_terminators(
 
 ## Non-Standard Syntax Gating (`--allow-x` Flags)
 
-**Rule**: Anything not in the IEC 61131-3 standard **must** be gated behind an `--allow-x` flag. Using `--dialect=rusty` enables the broadest set of extensions.
+**Rule**: Anything not in the IEC 61131-3 standard **must** be gated behind an `--allow-x` flag. `--dialect=codesys` currently enables the broadest set; `ironplcc dialects` prints what each one turns on.
```
garretfick: Let's not say which one has the broadest set. That's likely to change. And when we say "anything no in the standard" we mean IEC 61131-3 ed2.

specs/steering/syntax-support-guide.md
```diff
@@ -272,7 +272,7 @@ Dialects (`--dialect`) set the base configuration. Individual `--allow-*` flags
 |---------|-------------------|----------------|--------|-------------------|
 | IEC 61131-3 Ed 2 (default) | `iec61131-3-ed2` | OFF | OFF | all OFF |
 | IEC 61131-3 Ed 3 | `iec61131-3-ed3` | ON | ON | all OFF except `allow_partial_access_syntax` and `allow_fb_inheritance` — those two gate Edition 3 *standard* syntax (partial access, object orientation), not vendor extensions |
-| RuSTy | `rusty` | OFF | ON | all ON |
```
garretfick, line 275: This is repeating information that's elsewhere. I don't think we should list the features here. We might just list the dialects we support but no more than that.

garretfick (commented):
See my comments

specs/steering/syntax-support-guide.md
garretfick: Both done in `d76f39f1`.

The rule now reads: *"Anything not in IEC 61131-3 Edition 2 **must** be gated behind an `--allow-x` flag. That includes Edition 3 syntax, which is gated the same way."* No dialect is named as broadest.

Worth saying that my original edit here was the wrong instinct. I found the sentence was factually wrong (it said RuSTy was broadest; CODESYS is, 26 flags to 23) and corrected the fact — which would have left the same sentence to go stale again the next time a flag moves. Removing the claim is the fix.

---

specs/steering/syntax-support-guide.md
garretfick: Done in `d76f39f1`. The whole feature matrix is gone; the section now names the five dialects and stops:

> The dialects are `iec61131-3-ed2` (the default), `iec61131-3-ed3`, `rusty`, `codesys` and `twincat`.
>
> Which flags each one enables is **not** listed here, for the same reason the flags themselves are not: a mirrored table drifts. Run `ironplcc dialects`, or read the per-dialect `**Enables:**` lists in `enabling-dialects-and-features.rst`, which a build-time check keeps in step with `options.rs`.

That mirrors what the same file already says a few lines above about the flag list — *"The full list of existing flags is intentionally **not** duplicated here — a mirrored table drifts out of date."* The dialect table was the one place that ignored its own advice, and it is the one that drifted.

The evidence is fairly direct: `enabling-dialects-and-features.rst` has a build-time guard (`docs/extensions/ironplc_flags.py`) that diffs each dialect's `**Enables:**` list against the macro, and I verified it currently passes for all five dialects with no extra or missing flags. The steering table had no guard, and had drifted on two counts. Deleting it removes the class of problem rather than one instance.

---

PR #1597, opened 2026-09-01
Enforce unique ADR numbers and renumber the three collisions

garretfick:


specs/steering/development-standards.md
```diff
@@ -35,6 +35,21 @@ Trade-off analyses that capture **why** a particular approach was chosen over al
 
 That immutability covers the *decision*. It does not cover the *record* of it: a title that names the losing option, a status that never advanced past `proposed`, or a section describing a rejected option as though it had been built are all defects in the record, and leaving them in place is how an ADR comes to mislead the reader it exists to inform.
 
+#### Numbering
```
garretfick: Can we omit this? We have this already in common tasks. This add context with little value since there is a test for this anyway with a good error message.

garretfick: Removed the whole `#### Numbering` section in 30b52af.

The only edit left in this file is the invocation in the `plan-citations` paragraph — it said `just plan-citations` runs "as part of `just`", which is no longer true now that the recipe lives in `specs/justfile`, so it now reads `cd specs && just` ... "runs in CI".

---

PR #1600, opened 2026-09-01
Clarify temp buffer allocation and hard-link safety

garretfick:
This PR adds clarifying documentation about two subtle behaviors that can trap users:

## Summary

Expanded documentation in three areas to explain non-obvious runtime behavior and a potential safety gap:

1. **Temp buffer exhaustion in loops** — Clarified that the VM's temp buffer allocator is rewound only on function return, not on loop iterations. A string operation inside a loop therefore consumes one buffer per iteration despite being counted once at compile time, causing `V9009 TempBufferExhausted` if the loop runs more than a couple of times.

2. **Hard-link aliasing gap** — Documented that the output-conflicts-with-source check does not detect hard links. Two names for the same inode are distinct paths that both canonicalize to themselves, so writing output through one name can still overwrite a source read through the other.

3. **VS Code CodeLens layout** — Clarified that multiple CodeLens objects at the same range are intentional, not a layout accident; VS Code renders them as `|`-separated links.

## Key Changes

- **`compiler/codegen/src/compile.rs`** — Expanded `num_temp_bufs` documentation to explain that it counts static call sites, but the VM allocator behavior means loops cause per-iteration allocation, leading to pool exhaustion.

- **`specs/design/bytecode-instruction-set.md`** — Added section explaining temp buffer allocation semantics and why bundled utilities like `LREAL_TO_FMTSTR` use unrolled loops instead of actual loops.

- **`compiler/ironplc-cli/src/cli.rs`** — Expanded `output_conflicts_with_source` documentation to note that hard-link aliasing is not caught and suggest a potential fix (comparing device and inode).

- **`docs/reference/compiler/problems/P6009.rst`** — Added note that the hard-link gap exists and how it can occur.

- **`integrations/vscode/src/runCodeLensProvider.ts`** — Added comment explaining that multiple CodeLens at the same range is intentional for side-by-side rendering.

## Implementation Details

All changes are documentation-only; no functional behavior is modified. The documentation clarifies existing behavior that was previously implicit or undocumented.

https://claude.ai/code/session_016RWos6WEczkF7kDTLgLSRY


integrations/vscode/src/runCodeLensProvider.ts
```diff
@@ -29,6 +29,9 @@ const PROGRAM_RE = /^\s*PROGRAM\s+(\w+)/i;
  * stop lenses take no arguments; the run lens carries the program name so
  * it can be surfaced to the run command.
  */
+// Two lenses at the same range are not a layout accident: VS Code renders
```
garretfick: This would break JSDoc

garretfick: Right — a `//` comment between the JSDoc block and the declaration severs the two, so editors and API extractors drop the doc entirely. Fixed in `c25b2a50` by folding the note into the existing JSDoc as a trailing paragraph instead of adding a separate comment.

---

PR #1607, opened 2026-09-01
Implement string encoding verification (R0304)

garretfick:
## Summary

Implements compile-time verification of string encoding agreement across bytecode operations, preventing runtime `V9014` traps caused by mismatched `STRING` (Latin-1) and `WSTRING` (UTF-16LE) encodings. This adds the third layer of string encoding enforcement: the analyzer rejects declared encoding mismatches (P4034), codegen resolves operand encodings and produces values at the agreed width, and this change verifies the bytecode actually emitted matches those declarations.

## Key Changes

- **New verifier rule R0304** (`compiler/container/src/verify_string.rs`): Reads back declared slot widths from `STR_INIT` instructions and validates that operations addressing multiple slots agree on encoding. Checks:
  - Two-slot operations (`FIND_STR`, `CONCAT_STR`, `REPLACE_STR`, `INSERT_STR`)
  - `CMP_STR` operands pushed by preceding `LOAD_CONST_I32` instructions
  - String constants stored into slots (constant pool tag vs. slot declared width)

- **String width resolution** (`compiler/codegen/src/string_width.rs`): New module that determines the encoding each string operation's operands share:
  - `operand_width()` queries declared widths from variables, struct fields, array elements, and function returns
  - `resolve_operand_char_width()` settles the single encoding all operands of one operation must use
  - `compile_string_value()` produces expressions at the resolved encoding (literals adapt, named variables must match)
  - Literals take the encoding of their use context, enabling `w := 'abc'` to store UTF-16LE into a `WSTRING`

- **Bytecode verification integration** (`compiler/codegen/src/string_encoding.rs`): Entry point that runs R0304 verification after container construction and reports violations as internal errors (P9998).

- **Updated string compilation** (`compiler/codegen/src/compile_string.rs`):
  - `compile_string_compare()` resolves operand encoding before compiling both sides
  - `resolve_string_arg()` now takes `char_width` parameter and validates named variables match
  - `alloc_string_scratch()` allocates scratch slots at the resolved encoding
  - Removed `encode_string_literal()` in favor of `compile_string_value()`

- **Assignment compilation** (`compiler/codegen/src/compile_stmt.rs`): Uses `compile_string_value()` to produce string values at their destination's declared encoding.

- **Constant pool tracking** (`compiler/codegen/src/compile.rs`): Added `note_char_width()` call when loading string constants, enabling the verifier to correlate pool entries with their declared widths.

- **Documentation updates**:
  - `V9014.rst`: Clarified three-layer encoding enforcement
  - `wstring.rst`: Documented that literals adapt to their use context
  - `P4034.rst`: Extended to cover operations (not just assignments)
  - `bytecode-verifier-rules.md`: Added R0304 specification

- **End-to-end tests** (`compiler/codegen/tests/it/end_to_end_wstring.rs`): 11 new tests validating:
  - Literal adaptation to `WSTRING` context in comparisons, `FIND`, `CONCAT`
  - Nested calls with only literals adapting to target encoding
  - Literals passed to `WSTRING` function parameters
  - Struct field and array element assignments
  - Compile-time rejection of declared encoding mismatches (P4034)

## Implementation Details

- Verification runs after container construction, same point as stack balance verification
- Violations are reported as internal errors (P9998) because encoding mismatches indicate codegen bugs, not user program errors
- The verifier is conservative: it only checks what can be proven statically (immediate offsets, constant pool indices); computed offsets and values through temp buffers are not checked
- Literals have no inherent encoding and adapt to their use context, enabling natural code like `w = "abc"` where the literal is interned at the peer operand's width


docs/reference/runtime/problems/V9014.rst
```diff
@@ -7,18 +7,30 @@ V9014
 The VM detected a string operation whose operand encoding (1 for ``STRING``,
 2 for ``WSTRING``) does not match the destination's declared encoding.
 
-The analyzer rejects cross-encoding assignments and operations statically,
-so this trap is reached only via:
+Three compile-time layers stand between a program and this trap:
+
+1. The analyzer rejects assignments and comparisons that mix the declared
```
garretfick: This is explaining the internals of the compiler. We shouldn't be doing that here.

docs/reference/language/data-types/elementary/wstring.rst
```diff
@@ -46,9 +46,23 @@ count code units, not bytes and not user-perceived characters.
 
 The maximum declared length is 65,535 code units.
 
-``STRING`` and ``WSTRING`` cannot be mixed. Assigning one to the other is a
-compile error (:doc:`/reference/compiler/problems/P4034`); the runtime also
-traps a mixed operation that reaches it.
+``STRING`` and ``WSTRING`` cannot be mixed. Assigning one to the other, or
+comparing or combining them in one operation, is a compile error
+(:doc:`/reference/compiler/problems/P4034`); the runtime also traps a mixed
```
garretfick: I think we don't need the statement here about the runtime behavior. We just need to say the compiler gives an error in this case.

docs/reference/compiler/problems/P4034.rst
```diff
@@ -23,6 +23,14 @@ The following code will generate error P4034:
 
 The assignment ``narrow := wide`` copies a ``WSTRING`` into a ``STRING``, which would reinterpret UTF-16LE bytes as Latin-1 characters.
 
+The same applies to the operands of one operation. ``CONCAT(narrow, wide)``,
+``FIND(narrow, wide)`` and ``narrow = wide`` each need a single encoding to
+work in, and there is none both operands share.
+
+A string literal is an exception: it has no encoding of its own and takes the
+encoding of whatever it is used with, so ``wide := 'abc'`` and
+``wide = "abc"`` are both accepted and both store or compare UTF-16LE.
```
garretfick: Why are they stored at UTF-16LE?

compiler/container/src/verify_string.rs
```diff
@@ -0,0 +1,615 @@
...
+
+use crate::char_width::CharWidth;
+use crate::code_section::CodeSection;
+use crate::constant_pool::ConstantPool;
+use crate::container::Container;
+use crate::id_types::{ConstantIndex, FunctionId};
+use crate::instruction::Instruction;
+use crate::opcode::{self, Opcode};
+
+/// R0304: an operation whose operands do not agree on a string encoding.
```
garretfick: What is `R0304`?

compiler/container/src/verify_string.rs
```diff
@@ -0,0 +1,615 @@
...
+
+/// Decodes a function body linearly into its instructions, stopping at the
+/// first byte that is not an assigned opcode or whose operands run past the
+/// end. Both of those are what `verify_stack_balance` reports; this pass just
+/// stops looking.
+fn decode_body(bytecode: &[u8]) -> Vec<Decoded<'_>> {
+    let mut decoded = Vec::new();
+    let mut pc = 0usize;
+
+    while pc < bytecode.len() {
```
garretfick: Is this function truly unique? It's hard for me to image this the only time we've needed to decode the body.

docs/reference/runtime/problems/V9014.rst
garretfick: Agreed — removed. The page now says only that mixing STRING and WSTRING is a compile error (P4034), so reaching the trap means a compiler bug or a corrupt/hand-crafted `.iplc` file.

I did keep the correction that motivated the edit: the page previously claimed "the analyzer rejects cross-encoding assignments and operations statically", which is what made a six-line program's trap look impossible. That claim is gone rather than re-explained.

---

docs/reference/language/data-types/elementary/wstring.rst
garretfick: Done — the sentence now ends at the compile error, with the runtime-trap clause dropped.

---

docs/reference/compiler/problems/P4034.rst
garretfick: Because the destination is a `WSTRING`, and a literal has no encoding until it has a destination — my sentence asserted the outcome without saying that. Rewritten:

> A string literal is an exception, because it is not yet stored in either encoding: it is a sequence of characters that the compiler encodes when it sees where the value is going. Assigned to a `WSTRING` it is encoded as UTF-16LE, and assigned to a `STRING` as Latin-1, whichever quotes it is written with. So `wide := 'abc'` and `wide = "abc"` are both accepted, and neither reinterprets any bytes.

The last clause ties it back to the paragraph above it, which is what makes the exception an exception: the error exists because copying between the two encodings reinterprets bytes, and encoding a literal at its destination never does.

---

compiler/container/src/verify_string.rs
garretfick: A rule in `specs/design/bytecode-verifier-rules.md` — this PR adds it there (R0304: String Encoding Agreement, in the R0300 stack-type-correctness range). The code cited the bare number without saying that, which is the gap you're pointing at.

The module header now names the spec the way `verify.rs` does for its own rules:

> This module implements rule R0304 of the bytecode verifier specified in `specs/design/bytecode-verifier-rules.md`, which requires every string operation to find its operands in agreement about their encoding.

and the enum and its `rule()` accessor refer back to that instead of restating the code cold.

---

compiler/container/src/verify_string.rs
garretfick: You're right, it wasn't. There were three others before mine:

- `verify.rs::instruction_boundaries` — walks a body to mark instruction starts
- `verify.rs::method_return_depth` — walks a callee looking for `RET`
- `disassemble.rs::decode_instructions` — walks a body to render it

Each re-derived instruction lengths from the opcode tables, and a walk that gets one length wrong reads every later instruction at the wrong offset — so this is exactly the kind of thing that should exist once.

Extracted `instruction::decode_body` (in the module that already owns `Instruction` and `Operand`), yielding a `DecodedInstruction` — offset, opcode, declared layout, operand bytes — or a `DecodeStop` for a byte it cannot read. The four walks differed only in what they made of a stop, so the shared walk takes no position and each caller states its own: the stack-balance verifier rejects the container, the disassembler renders an `UNKNOWN`/`<truncated>` row, the string-encoding verifier stops looking. An unassigned byte is reported and the walk resumes after it, which is what preserves the disassembler's ability to render every row of a corrupt body.

Two accessors came out of it too: `operand(n)` reads the nth operand at its declared width (replacing the hand-rolled version you commented on), and `operands_with_kinds()` pairs each operand with its own bytes, which replaced the disassembler's absolute-offset arithmetic in `format_operand`.

One decoder stays separate: codegen's peephole optimizer (`optimize/rewrite.rs`). It holds owned bytes so it can re-emit them, and deliberately keeps the clamped tail of a truncated instruction — this walk yields an error and no bytes there, so adopting it would change what that pass emits. Happy to look at that separately if you'd rather it converge too.

---

garretfick:
Super-seeded by individual PRs.

PR #1610, opened 2026-09-02
Accept a literal and a nested call as LEN's argument

### LINKED ISSUE #1591: LEN rejects a literal argument with P9999 — including the exact example published in len.rst
Found triaging plans for #1464. **Ran; reproduces with two controls.**

`docs/reference/standard-library/functions/len.rst:66` publishes this as a runnable example:

```
result := LEN('Hello');    (* result = 5 *)
```

It does not compile.

```iecst
PROGRAM main
VAR result : INT; END_VAR
  result := LEN('Hello');
END_PROGRAM
```

```
$ ironplcc compile -o l.iplc lendoc.st
error[P9999]: Capability is not implemented (yet!)
   Not implemented at codegen/src/compile_string.rs#L50
```

**Control 1** — the same shape with a different string function: `result := LEFT('Hello', 3);` compiles clean. Every other string function's docs example also compiles and runs — `CONCAT`, `LEFT`, `MID`, `FIND`, `INSERT`, `DELETE`, `REPLACE`. `LEN` is the sole outlier.

**Control 2** — `LEN` with the literal hoisted into a variable works:

```iecst
s : STRING[32] := 'hello world';
t := MID(s, 3, 1);
n := LEN(t);        (* n: 3 — fine *)
```

`LEN(MID(s, 3, 1))` inline gives the same P9999.

## Cause

`compile_len` (`compiler/codegen/src/compile_string.rs:45-50`) matches only `ExprKind::Variable` and falls through to `Diagnostic::todo_with_span` for anything else. It never calls `resolve_string_arg`, which is the helper that gives every other string function its literal and nested-call support.

The nested-call fix (`2026-04-04-fix-nested-stdlib-calls`) added the catch-all arm to `resolve_string_arg` — but because `compile_len` bypasses that helper entirely, the fix never reached it.

## Why the docs drifted

`docs/extensions/ironplc_playground.py` renders `playground-with-program` directives but does not compile them. So a published example that is advertised as runnable in the browser is not checked against the compiler at build time. That is worth considering separately — this bug is one instance of a class those directives cannot currently catch.

## Fix

Route `compile_len` through `resolve_string_arg` like its siblings. The temp-buffer accounting will need the same treatment as the other functions (see #1590 for a separate hazard in that area).

garretfick:
Fixes #1591. Rebased onto `main` now that the prefactor (#1609) has merged.

`compile_len` matched only `ExprKind::Variable` and reported P9999 for anything else, so `LEN('Hello')` — the example published in the LEN reference page — did not compile, and a nested call had to be hoisted into a variable first. This routes the argument through `resolve_string_arg`, the helper that already gives every other string function its literal and nested-call support.

That helper allocated every intermediate slot narrow, which the wide case cannot use: a `WSTRING` literal came out encoded as Latin-1, and a wide nested call trapped because the VM refuses a store whose source and destination headers disagree (ADR-0034). The encoding is known at compile time in every case — a literal spells it, a declaration states it, and each string function returns the encoding of its first string argument — so infer it and allocate the slot at that width. A wide intermediate also marks the program as holding a wide string, so temp buffers are sized in wide bytes even when no `WSTRING` variable is declared.

This repairs the wide case for `CONCAT`/`LEFT`/`MID`/etc. too, which shared the same narrow-only helper.

## Tests

Ten new end-to-end tests, each verified to fail without the change:

- `end_to_end_len.rs` — `LEN` of a string literal, an empty literal, a `WSTRING` literal, a non-ASCII BMP `WSTRING` literal, a nested `MID` on both `STRING` and `WSTRING`, and a nested `CONCAT` of narrow and of wide literals.
- `end_to_end_wstring.rs` — a `WSTRING` literal inside `CONCAT` is genuinely UTF-16LE, both alongside a variable and in a literals-only expression.

Verified through the CLI as well: the issue's program compiles and prints `result: 5` / `wresult: 5`.

`cd compiler && just` and `cd specs && just` both pass on the rebased branch.

## Not included

`len.rst`'s "accepts `STRING`" line matches the phrasing of all nine string-function reference pages, so mentioning `WSTRING` in just that one would leave the set inconsistent. The published example itself now works as written.

https://claude.ai/code/session_01ASTm4PY78yMHvmiGnqBCJA


compiler/codegen/src/compile_string.rs
```diff
@@ -98,10 +85,88 @@ pub(crate) fn compile_string_compare(
...
+/// Anything whose width cannot be determined here keeps the narrow default.
+fn string_expr_char_width(ctx: &CompileContext, expr: &Expr) -> CharWidth {
+    match &expr.kind {
+        ExprKind::Const(ConstantKind::CharacterString(lit)) => {
+            char_width_for_string_type(&lit.width)
+        }
+        ExprKind::Expression(inner) => string_expr_char_width(ctx, inner),
+        ExprKind::Variable(variable) => variable_char_width(ctx, variable),
+        ExprKind::Function(func) => function_char_width(ctx, func),
+        _ => NARROW_CHAR_WIDTH,
```
garretfick: Defaulting to narrow is wrong - that make this a runtime error as opposed to compile time. If we cannot figure it out, make this a compiler internal error

compiler/codegen/src/compile_string.rs
```diff
@@ -98,10 +85,88 @@ pub(crate) fn compile_string_compare(
...
+    match variable {
+        Variable::Symbolic(SymbolicVariableKind::Structured(structured)) => {
+            match crate::compile_struct::walk_struct_chain(
+                ctx,
+                &structured.record,
+                &structured.field,
+                0,
+            ) {
+                Ok((_, _, IntermediateType::String { char_width, .. })) => char_width,
+                _ => NARROW_CHAR_WIDTH,
```
garretfick: Similar - don't default to narrow.

compiler/codegen/tests/it/end_to_end_len.rs
```diff
@@ -63,6 +63,119 @@ END_PROGRAM
...
+// n is at variable slot 0. LEN accepts a literal argument directly; this is
+// the example published in the LEN reference documentation.
+e2e_i32!(
+    end_to_end_when_len_of_string_literal_then_returns_length,
+    "
+PROGRAM main
+  VAR
+    n : INT;
+  END_VAR
+  n := LEN('Hello');
```
garretfick, line 75: Create an issue for this as an enhancement. This returns a constant so there is no reason to execute this at runtime. This would also apply if we can otherwise determine that the string is constant (if it has the CONST qualifier or if it is never written to - we might do this later part as an xform where we apply const so that code gen is simple).

garretfick (commented):
see comments

compiler/codegen/src/compile_string.rs
garretfick: Agreed — fixed in 97ee833. `string_expr_char_width` now returns `Result<CharWidth, Diagnostic>` and reports P9998 via `Diagnostic::internal_error_at` rather than guessing narrow. Same for the two helpers it delegates to.

Confirmed it isn't reachable from ill-typed source: `LEN(5)` is still rejected by the analyzer with P4026 before codegen sees it, which is the right split — a program error stays a program error, and P9998 means only what it says.

---

compiler/codegen/src/compile_string.rs
garretfick: Done in 97ee833.

Worth flagging what removing this default exposed: two existing tests started failing on P9998 — `end_to_end_when_find_with_struct_array_field_then_returns_position` and `end_to_end_when_struct_2d_string_array_field_compared_then_matches`. A STRING array inside a structure field (`setup.NAMES[2]`) resolves through the struct chain, not `array_vars`, so it was silently taking the narrow default and only working because those tests are narrow.

`variable_char_width` now walks past subscripts and dereferences to the access root and reads the width from the declaration, so that shape is resolved properly rather than guessed. Exactly the kind of thing the default was hiding.

---

compiler/codegen/tests/it/end_to_end_len.rs
garretfick: Filed as #1612, covering all three cases you named: literals (and expressions built only from literals, like `LEN(CONCAT('ab', 'cde'))`), `CONSTANT`-qualified strings, and never-written strings via a transform that applies `CONSTANT` before codegen so the code generator only has to understand the first two.

Noted there that folding must count UTF-16 code units for a `WSTRING` literal, to match what `LEN_STR` reads at run time.

---

Issue #1545, opened 2026-08-30
A prefixed literal outside its own type's range is not diagnosed


garretfick:
A typed integer literal states its own type, and the value can contradict it:

```iecst
PROGRAM main
  VAR
    x : DINT := INT#40000;
  END_VAR
END_PROGRAM
```

`INT` is -32768..32767, so `INT#40000` is not an `INT`. The value fits the `DINT` it is stored into, so a check against the *destination* type accepts it. Nothing checks the literal against the type it claims to be.

## How this differs from #1527

#1527 is a constant that does not fit its destination (`a : USINT := 300`). This one fits its destination and contradicts itself, so it needs the literal's own `data_type` rather than the destination's. The two checks answer different questions and were separated deliberately: the fix for #1527 covers only the destination.

## Notes

`IntegerLiteral::data_type` holds the prefix (`Option<IntegerTypeName>`), so the value is available where the range check for #1527 runs. `IntegerTypeName` maps to `SINT`/`INT`/… directly, so the range comes from the same elementary type facts that check the destination.

Worth settling before implementing: whether a radix-prefixed form such as `INT#16#FFFF` counts. That is a bit pattern spelled with an integer type prefix, and the fix for #1527 deliberately leaves radix literals unchecked on the grounds that a pattern is not a number.

PR #1611, opened 2026-09-02
Report a constant that does not fit its type

### LINKED ISSUE #1527: Out-of-range constant assigned to a narrow integer type wraps silently

garretfick:
Fixes #1527. Builds on the prefactoring in #1533.

A constant outside its type's range was accepted and silently wrapped. `a : USINT := 300` stored 44, `c : SINT := 200` stored -56, and `ironplcc check` exited 0. Codegen checked a literal against the width it operates at, so `SINT`, `INT` and `DINT` all collapsed to 32 bits and only a value outside `i32` was caught — and `check` never generates code, so an editor showed nothing at all.

## Commits

**Record where an integer literal is.** The parser built every integer literal with an empty span, so a diagnostic about one landed on line 1, column 1 — including today's P2026 from codegen and P4025 from the bit access rule. The token's span is threaded through, and a folded literal takes the span of the expression it replaces so that `255 + 1` points at the sum. `SourceSpan` comparison always answers true, so the ASTs the parser tests build with default spans still match.

**State a type's value range once.** `value_range` answers what a type can hold, from the same `IntermediateType` record codegen projects onto `VarTypeInfo`. `narrow_type_range` in codegen takes its numbers from it rather than restating the bounds of an 8- and 16-bit integer.

**Report a constant that does not fit its type.** `rule_constant_range` checks a constant against the type it is stored or compared as, pushing that type through operators the way codegen pushes one operation type into both operands — so `x := 255 + 1` is caught exactly as `x := 256` is.

## What is checked

| Where | Example |
|---|---|
| Variable initializer | `a : USINT := 300` |
| Assignment, including struct fields, array elements and globals | `b := 300`, `s.count := 300` |
| Operand of an operator | `b := 255 + 1` |
| Comparison | `IF c = 200 THEN` |
| `CASE` label | `200:` on a `SINT` selector |

Subrange types are checked against their declared bounds (`Ratio : INT(0..10)` with `r : Ratio := 20`), which nothing checked before.

## What is not

`BYTE`, `WORD`, `DWORD` and `LWORD` keep wrapping — they are patterns rather than magnitudes, and `end_to_end_when_byte_overflow_then_wraps` still asserts it. `REAL` and the rest are not integer storage.

**One decision changed while implementing.** The plan excluded radix-prefixed literals (`16#FF`) as patterns in the same sense as a bit string. That turned out to be neither implementable nor right: a radix does not survive parsing, so `16#1FF` and `511` are the same literal by the time a rule sees them, and IEC 61131-3 bounds a literal by the type it is used as. A test asserting the old behaviour is what caught it. The *type* decides, not the spelling.

Codegen keeps its own check as a user-facing diagnostic. It cannot tell a constant this rule checked from one the analyzer could never type, so a gap here degrades to today's behaviour rather than to an internal error about a program that is merely wrong.

A literal that contradicts its own prefix (`x : DINT := INT#40000`) is a different question, tracked as #1545.

`cd compiler && just` passes, as does `cd specs && just`. The plan has been removed per the repository's process; it stays readable in the commit that added it (`Plan the out-of-range constant rule`).

https://claude.ai/code/session_01UtwxiCCLtHgKgAurqYKvjD


docs/reference/compiler/problems/P2026.rst
```diff
@@ -4,3 +4,123 @@ P2026
 
 .. problem-summary:: P2026
 
+This error occurs when a constant does not fit the type it is stored into.
+
+Every integer type holds a fixed range of values:
+
+- ``SINT``: -128 to 127
```
garretfick: Would this be better represented as a table?

garretfick: Agreed — it is now a `list-table` with Type / Minimum / Maximum columns, matching the tables in the explanation pages, so the bounds line up for comparison. Pushed as f7e17f5; the docs build and the duplication check pass.

---

PR #1616, opened 2026-09-03
Fix for the float additive-identity peephole

garretfick:
Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_013qrymdaJ5ZuzqYBQK8Kbeq


compiler/codegen/tests/it/end_to_end_arith_identity.rs
```diff
@@ -0,0 +1,111 @@
...
+    );
+    let (_c, bufs) = parse_and_run(&source, &CompilerOptions::default());
+
+    assert_eq!(float.read(&bufs, 1), y, "y mismatch");
+    assert_eq!(float.read(&bufs, 2), z, "z mismatch");
+}
+
+#[rstest]
+#[case::value_plus_zero("x + 0", -7)]
+#[case::zero_plus_value("0 + x", -7)]
```
garretfick, line 85: Should we not have the `-0` test which shows that it has the value `-0`? After all, that's the point of this whole work.

garretfick: Agreed. In 3ef7d9a the `y` and `z` assertions compare `to_bits()`, so `-0.0` and `0.0` are distinct, and the `neg_zero_minus_zero` case now states `y = -0.0` explicitly (`x - 0.0` on `-0.0` keeps the sign). The `x + 0.0` cases state `y = 0.0`, which the bit comparison now actually checks.

---

Issue #1619, opened 2026-09-03
MOD is declared ANY_NUM, so MOD on REAL passes check and fails compile with P9998


garretfick:
Filed from the plan for #1567, which scoped this out.

IEC 61131-3 defines `MOD` over `ANY_INT`. IronPLC declares its function form `ANY_NUM` (`compiler/analyzer/src/intermediates/operator_function_form.rs`), and codegen's `emit_mod` has no float arm (its comment says the analyzer should reject float `MOD` first). Nothing does, so a `REAL` `MOD` gets past `check` and then fails in `compile` as an internal error:

```iecst
PROGRAM main
VAR
  r1 : REAL := 7.5;
  r2 : REAL := 2.0;
  r3 : REAL;
END_VAR
  r3 := MOD(r1, r2);
END_PROGRAM
```

```
$ ironplcc check mod.st      (* clean *)
$ ironplcc compile mod.st
error[P9998]: Internal error indicating a bug in the compiler
```

The operator form `r1 MOD r2` is not type-checked at all today, so it takes the same path.

Suggested fix: change the `MOD` row's operand category to `ANY_INT`, which makes the function form P4026 on a real argument; the operator form needs an operand check of its own. The `__MOD` intrinsic remains the floating-point remainder for compatibility libraries.

Issue #1618, opened 2026-09-03
Extensible standard functions (ADD, MUL, AND, OR, XOR) accept only two inputs


garretfick:
Filed from the plan for #1567, which scoped this out.

IEC 61131-3 defines `ADD`, `MUL`, `AND`, `OR` and `XOR` as extensible: any number of inputs `IN1..INn`. IronPLC declares each with exactly two inputs, so a three-input call is rejected:

```iecst
PROGRAM main
VAR
  b1 : BOOL;
  b2 : BOOL;
  b3 : BOOL;
  b4 : BOOL;
END_VAR
  b4 := AND(b1, b2, b3);
END_PROGRAM
```

```
error[P4018]: Function call has wrong number of arguments (function=AND, expected=2, actual=3)
```

`MUX` already uses the extensible signature mechanism (`FunctionSignature::stdlib_extensible`), so the analyzer side is a matter of marking these rows extensible in `compiler/analyzer/src/intermediates/operator_function_form.rs`. Codegen compiles a function form as its operator (`compile_operator_form` in `compiler/codegen/src/compile_call.rs`), so an n-input call would fold left: `AND(a, b, c)` as `(a AND b) AND c`.

PR #1620, opened 2026-09-03
Accept ANY_BIT operands in the function forms of AND, OR, XOR and NOT

### LINKED ISSUE #1567: Functional AND/OR/XOR reject ANY_BIT operands: AND(w1, w2) on WORD is a false positive
Found triaging `assignment-and-stdlib-arg-type-check.md` for #1464. **Ran; reproduces with a control.**

IEC 61131-3 defines `AND`, `OR`, `XOR` and `NOT` over `ANY_BIT`. IronPLC declares them `BOOL`-only, so the functional form is rejected for its standard operand type.

```iecst
PROGRAM main
VAR
  w1 : WORD := 16#F0;
  w2 : WORD := 16#0F;
  w3 : WORD;
END_VAR
  w3 := AND(w1, w2);
END_PROGRAM
```

```
$ ironplcc check andword.st
error[P4026]: … (function=AND, parameter=IN1, expected=BOOL, actual=word)
error[P4026]: … (function=AND, parameter=IN2, expected=BOOL, actual=word)
```

**Control** — the same program with the operator form `w3 := w1 AND w2;` checks clean (0 errors). So this is specific to the functional spelling, and bitwise AND on `WORD` is otherwise fully supported.

## Why it appeared now

This is a **regression introduced by closing a different gap**, and the plan that introduced it foresaw it and scoped it out:

> Bit-string family for functional `AND`/`OR`/`XOR` (modeled as `BOOL` in stdlib, a pre-existing limitation) beyond what the corpus needs.

Before stdlib argument type-checking existed, arguments were unchecked, so the `BOOL`-only declaration was latent and harmless. Adding the check turned it into a live false positive. The scope note was never converted into an issue, which is why this is being filed now — the plan is being deleted.

## Cause

`compiler/analyzer/src/intermediates/stdlib_function.rs:602-625` declares the family `BOOL`-only, with the comment "All take BOOL inputs and return BOOL". They should be `ANY_BIT` (`BOOL`, `BYTE`, `WORD`, `DWORD`, `LWORD`), with the result type following the operand type.

## No docs claim is broken

`docs/reference/language/data-types/elementary/word.rst:38` shows the operator form, which works. So nothing in the docs promises the functional form — but a standard function rejecting its standard operand type is still wrong, and a user coming from another IEC toolchain will write `AND(a, b)`.

garretfick:
Fixes #1567. Stacked on the prefactor #1617; this PR's base is that branch, so its diff shows only the fix. GitHub retargets it to `main` when #1617 merges and its branch is deleted.

The first commit is the plan (deleted in the last commit, per the development standards); it is the proposal for the prefactor and this fix, and is the place to read for the reasoning.

## The defect

`AND(w1, w2)` on `WORD` was rejected with P4026 on each argument while `w1 AND w2` checked clean. `NOT(IN := w1)` (the spelling that reaches the function form; positional `NOT(x)` parses as the operator) was rejected the same way. IEC 61131-3 defines all four over `ANY_BIT`; IronPLC declared them `BOOL`.

## The fix

Four operand cells in the operator-form table become `ANY_BIT`. After #1617 the result type follows the operand type by construction, so `AND(w1, w2)` resolves to `WORD`, and codegen compiles the call as the operator, so `NOT(IN := w1)` takes the bitwise complement rather than the `BOOL_NOT` the old function-form path always emitted.

## Keeping it fixed

- `specs/design/keyword-function-forms.md` gains a *Signatures and code generation* section with requirement IDs, one per family, registered in both crates' `build.rs`.
- Analyzer conformance tests: each family accepts every type in its category, and an argument outside it is P4026.
- Codegen conformance test: for nineteen (function, operator, type) cases, the function form and the operator expression compile to identical bytecode.
- Rule, resolver and end-to-end tests for the `WORD` cases in the issue, plus `NOT(IN := w)` on `WORD` and `BOOL`.

## Docs

`AND`, `OR`, `XOR` and `NOT` had no reference pages, unlike `ADD` or `SHL`. Added under *Bit String Functions*, following the `ADD` page.

## Out of scope, filed as issues

- #1618 — `ADD`, `MUL`, `AND`, `OR`, `XOR` are extensible in the standard; IronPLC declares two inputs.
- #1619 — `MOD` is declared `ANY_NUM`; `MOD` on `REAL` passes `check` and fails `compile` with P9998.

## Verification

`cd compiler && just`, `cd docs && just ci` and `cd specs && just` pass. The issue's program and the `NOT(IN := w1)` variant check clean with the built CLI.

https://claude.ai/code/session_01GErPKvxsfPnYcBkz4cHq3v

---
_Generated by [Claude Code](https://claude.ai/code/session_01GErPKvxsfPnYcBkz4cHq3v)_


compiler/analyzer/src/spec_conformance_keyword_function_forms.rs
```diff
@@ -0,0 +1,123 @@
...
+    );
+}
+
+/// REQ-KF-analyzer-001: the arithmetic forms accept ANY_NUM and return the
+/// operand type, so the result assigns to a variable of that type.
+#[spec_test(REQ_KF_analyzer_001)]
+fn analyzer_spec_req_kf_001_arithmetic_forms_accept_any_num() {
+    for function in ["ADD", "SUB", "MUL", "DIV", "MOD"] {
+        for operand_type in [
+            "SINT", "INT", "DINT", "LINT", "USINT", "UINT", "UDINT", "ULINT",
```
garretfick: Why not have REAL and LREAL here? Can these use rstest instead of loops?

compiler/analyzer/src/spec_conformance_keyword_function_forms.rs
```diff
@@ -0,0 +1,123 @@
...
+        for operand_type in ["REAL", "LREAL"] {
+            assert_clean(operand_type, operand_type, &format!("{function}(a, b)"));
+        }
+    }
+}
+
+/// REQ-KF-analyzer-002: the comparison forms accept ANY_ELEMENTARY and return
+/// BOOL.
+#[spec_test(REQ_KF_analyzer_002)]
+fn analyzer_spec_req_kf_002_comparison_forms_accept_any_elementary() {
```
garretfick: Can these use rstest? Or is the problem the spec_test?

compiler/analyzer/src/spec_conformance_keyword_function_forms.rs
```diff
@@ -0,0 +1,123 @@
...
+    for function in ["GT", "GE", "EQ", "LE", "LT", "NE"] {
+        for operand_type in ["INT", "DINT", "REAL", "WORD", "TIME"] {
```
garretfick: This doesn't include all of the types

compiler/analyzer/src/spec_conformance_keyword_function_forms.rs
```diff
@@ -0,0 +1,123 @@
...
+            assert_clean(operand_type, "BOOL", &format!("{function}(a, b)"));
+        }
+    }
+}
+
+/// REQ-KF-analyzer-003: AND, OR and XOR accept every ANY_BIT type and return
+/// the operand type.
+#[spec_test(REQ_KF_analyzer_003)]
+fn analyzer_spec_req_kf_003_bitwise_forms_accept_any_bit() {
+    for function in ["AND", "OR", "XOR"] {
```
garretfick: This isn't the complete list.

compiler/analyzer/src/spec_conformance_keyword_function_forms.rs
garretfick: Done in 7399893. REAL and LREAL are now in the list for all five functions, MOD included: the row says ANY_NUM, and the analyzer accepts MOD on REAL today. #1619 is the ticket to change that row to ANY_INT, and this case will change with it.

The loops are now `#[rstest]` `#[values]` matrices. What blocked that was `spec_test` unconditionally adding `#[test]`; it now defers to `#[rstest]`, which generates the test functions from the cases itself.

---

compiler/analyzer/src/spec_conformance_keyword_function_forms.rs
garretfick: It was the `spec_test` macro: it always added `#[test]`, which does not combine with `#[rstest]` (a function with parameters cannot be a plain test). In 7399893 the macro skips `#[test]` when `#[rstest]` is present, so `#[spec_test]` goes first and `rstest` expands the cases as usual. All five analyzer tests and the codegen one are rstest matrices now.

---

compiler/analyzer/src/spec_conformance_keyword_function_forms.rs
garretfick: Now all 25 elementary types the compiler knows, for each of the six comparison forms (7399893). Strings, both string encodings, and all eight time and date types included; every case analyzes clean.

---

compiler/analyzer/src/spec_conformance_keyword_function_forms.rs
garretfick: Agreed that a hand-written list is the weak point. In 7399893 the positive tests list every type of the category explicitly, and the P4026 test no longer has a list to get wrong: it is a 15-function × 25-type matrix, and which side of the line a type falls on comes from the form's row (`operator_function_form(name).signature()`) and `GenericTypeName::is_compatible_with`, so an admitted type must analyze clean and every other type must be P4026. 375 cases, all passing.

---

Issue #1621, opened 2026-09-04
Operator spellings of +, -, *, / are not operand-type-checked, and ADD(t1, t2) on TIME is a false positive


garretfick:
Filed from the plan for #1619, which scoped this out.

#1619 adds `rule_operator_operand_type_check`, which holds the operands of `a MOD b` to the operand category the operator-form table declares for `MOD` (`ANY_INT`), so the operator and the function form `MOD(a, b)` agree by construction. Only `MOD` is checked. The other arithmetic operators were deliberately left out, because applying the table to them today would turn working programs into errors:

- The table declares `ADD`, `SUB`, `MUL`, `DIV` over `ANY_NUM`, and the function-call rule enforces that for the function forms: `ADD(t1, t2)` on `TIME` is P4026 today.
- The operator spellings compile fine for operands outside `ANY_NUM`: `t1 + t2` on `TIME` and `b1 + 1` on `BYTE` both check and compile cleanly.
- IEC 61131-3 Table 30 defines `ADD` and `SUB` on `TIME` (and `MUL`/`DIV` of `TIME` by `ANY_NUM`), so the function-form rejection of `ADD(t1, t2)` is a false positive in the same family as #1567, and bit-string operands in arithmetic are a dialect question rather than a standard one.

So two things remain to decide, together:

1. What the operator spellings of `+`, `-`, `*`, `/` should accept (the standard's `ANY_NUM`, `ANY_MAGNITUDE` for the `TIME` overloads, and whether bit strings are allowed under a dialect or `--allow-*` flag).
2. Whether the table's `ANY_NUM` cell for `ADD`/`SUB`/`MUL`/`DIV` is the right declaration for the function forms, given Table 30.

Once decided, `checked_form` in `compiler/analyzer/src/rule_operator_operand_type_check.rs` is the one place to widen the operator check, and the table cell is the one place to change what the function form accepts.

PR #1623, opened 2026-09-04
MOD accepts ANY_INT in both spellings, and says so at check time

### LINKED ISSUE #1619: MOD is declared ANY_NUM, so MOD on REAL passes check and fails compile with P9998

garretfick:
Fixes #1619.

Built on the prefactor #1622 (now merged); the branch carries a merge of `main` after that.

## Problem

IEC 61131-3 defines `MOD` over `ANY_INT`. IronPLC declared the function form `ANY_NUM`, so `MOD(r1, r2)` on `REAL` passed analysis, and the operator spelling `r1 MOD r2` was not type-checked at all. Codegen has no floating-point `MOD` opcode, so both reached the bytecode verifier with an unbalanced operand stack and surfaced as an internal error (P9998) from `compile` instead of a type error from `check`.

## Change

- **Function form.** The `MOD` row of the operator-form table now says `ANY_INT`. The signature is derived from the row, so the existing check reports P4026 on a real argument with no further change.
- **Operator form.** A new rule, `rule_operator_operand_type_check`, looks up the row by operator and checks each operand with the same `type_compat::are_types_compatible` predicate the function-call rule uses, so `MOD(a, b)` and `a MOD b` agree by construction. A mismatch is a new problem code, **P4049**, reported once per offending operand at the operand's span.
- **Constant folder.** A real `MOD` on literals (`7.5 MOD 2.0`) is no longer folded into a remainder the language does not define; it is left for the rule to reject.
- **Codegen.** `emit_mod` keeps its no-op float arms as the last line of defence; the comment now names the rules that make them unreachable.
- **Spec.** `specs/design/keyword-function-forms.md` (which gained requirement IDs in #1620 after this branch was cut): REQ-KF-analyzer-001 now names the four `ANY_NUM` operators, REQ-KF-analyzer-006 states `MOD` over `ANY_INT`, and REQ-KF-analyzer-007 states that the operator spelling is held to the same row. Conformance tests for both new requirements; the 007 test walks every elementary type and checks both spellings agree (P4026 for the form, P4049 for the operator, clean for both otherwise).
- **Docs.** `docs/reference/compiler/problems/P4049.rst`.

## Scope

Only `MOD` is checked. The other arithmetic operators are declared `ANY_NUM` in the table, but their operator spellings compile today for `TIME` and bit-string operands (`t1 + t2`, `b1 + 1`), and IEC 61131-3 Table 30 defines `ADD`/`SUB` on `TIME`, so holding them to the table would break working programs. That decision is tracked in #1621.

## Behaviour to be aware of

- A bit-string operand to the `MOD` operator (`d MOD b` with `b : BYTE`) is now P4049. `MOD(d, b)` was already P4026 before this change, so the two spellings now agree; the operator spelling previously compiled.
- An operand whose type the predicate cannot judge (a subrange, an enumeration) is skipped, not reported. `p MOD 2` on a subrange of `INT` compiles before and after.
- A real literal carries no source span, so P4049 on `7.5 MOD 2.0` points at the start of the file. Same limitation as other diagnostics on real literals.

## Verification

`cd compiler && just` (compile, coverage, lint, dupes), `cd docs && just`, and `cd specs && just` all pass on the current head. Through the CLI, the program from the issue now reports P4026 for `MOD(r1, r2)` and P4049 for `r1 MOD r2` from `check`; `t1 + t2` on `TIME` still compiles.

The plan was committed as the first commit on the branch and removed before the merge of `main`, per the workflow.

https://claude.ai/code/session_01Ur9De1VUvwEbpEHTw8uhBk


docs/reference/compiler/problems/P4049.rst
```diff
@@ -0,0 +1,48 @@
...
+   PROGRAM main
+   VAR
+       angle : REAL := 400.5;
+       wrapped : DINT;
+   END_VAR
+       wrapped := REAL_TO_DINT(angle) MOD 360;
+   END_PROGRAM
+
+For a floating-point remainder, use a library function written for it. The
+Beckhoff ``Tc2_Math`` compatibility library provides ``LMOD`` for ``LREAL``
```
garretfick: Let's not mention Beckhoff here. That won't extend to all. What we should do is document what function you can use. That should also be an example of what to do.

garretfick: Done in 204cd6f. The Beckhoff reference is gone. The page now shows the floating-point remainder written with `TRUNC` and `DINT_TO_REAL`, which every program has, as a second worked example:

```
wrapped := angle - DINT_TO_REAL(TRUNC(angle / 360.0)) * 360.0;  (* 40.5 *)
```

I compiled and ran it to confirm the pinned result. The integer example (`REAL_TO_DINT(angle) MOD 360`) stays as the first fix.

---

PR #1629, opened 2026-09-04
Document byte/word/dword/lword partial access and fix 32/64-bit slice writes

### LINKED ISSUE #1595: Docs claim byte/word/dword/lword partial access is unsupported and predict a specific error; it is fully implemented
Found triaging plans for #1464. **Ran; reproduces with a control.**

`.%Bn`, `.%Wn`, `.%Dn` and `.%Ln` work. Five user-facing doc locations say they do not, and two of them predict a specific diagnostic that never fires.

## Reproduction

```iecst
PROGRAM main
VAR
    d : DWORD := 16#12345678;
    b : BYTE;
    w : WORD;
END_VAR
    b := d.%B2;
    w := d.%W1;
END_PROGRAM
```

```
$ ironplcc check --dialect rusty partial.st
exit 0, no diagnostics          # docs predict P0003

$ ironplcc check partial.st     # flag off
error[P4033]                    # correct
```

**Control**: the same program using `.%X2` also exits 0 under `--dialect rusty`. So the flag-on path is not a lexer error for either form.

Implementation: `compiler/parser/src/parser.rs:882-885` (four token alternatives), `compiler/dsl/src/textual.rs:221-272` (`PartialAccessSize`, `PartialAccessVariable`), `compiler/plc2plc/src/renderer.rs:1829`, plus a 17-test e2e suite at `compiler/codegen/tests/it/end_to_end_partial_access.rs` covering reads, writes, array elements and struct fields.

## The five locations

- `docs/explanation/enabling-dialects-and-features.rst:296-297` — "Byte/word/dword/lword partial access (``.%Bn``, ``.%Wn``, ``.%Dn``, ``.%Ln``) is not yet supported."
- `docs/reference/compiler/ironplcc.rst:199-200` — same sentence.
- `docs/reference/compiler/problems/P4033.rst:42-47` — "The wider forms … **are not yet accepted when the flag is on and still produce a lexer error there.**"
- `docs/reference/language/structured-text/bit-access.rst:130-152` — a four-row table with Status "Not yet supported" for each, plus "``P0003`` (``Unmatched character sequence``) when the flag is on."
- `docs/reference/language/edition-support.rst:48-49` — "byte/word/dword/lword forms are not yet supported".

## Two companions to the same drift

**The CLI help text names the wrong dialect set.** `compiler/ironplc-cli/bin/main.rs:168-170` says the flag is "Enabled by `--dialect=iec61131-3-ed3` and `--dialect=rusty`". Verified that `--dialect codesys` and `--dialect twincat` also enable it (both exit 0 on the control program). This string is visible in `ironplcc check --help`.

`specs/design/partial-access-bit-syntax.md:29-30` has the same omission.

## Rescue needed before the plan is deleted

`specs/design/partial-access-bit-syntax.md:23-27` explicitly defers multi-byte access — "those are out of scope here and **will be covered by a separate design**". That separate design was never written. So the only statement of multi-byte *semantics* outside test fixtures is the table in the retired `2026-04-17-complete-bit-access-support` plan:

| form | width | result type | example |
|---|---|---|---|
| `.%Bn` | 8-bit | BYTE | `DWORD_VAR.%B2` → bits 16-23 |
| `.%Wn` | 16-bit | WORD | `LWORD_VAR.%W1` → bits 16-31 |
| `.%Dn` | 32-bit | DWORD | `LWORD_VAR.%D1` → bits 32-63 |
| `.%Ln` | 64-bit | LWORD | `LWORD_VAR.%L0` → all 64 bits |

That should land as a `## Multi-Byte Partial Access` section in `partial-access-bit-syntax.md` with `REQ-PAB-1xx` IDs — the e2e tests are already there to cite — and the same edit should retire the deferral paragraph and fix the dialect list.

Partial mitigation already in code: `PartialAccessSize::bit_width` (`textual.rs:228-235`) and per-test comments like `// Byte 3 = most significant byte = 0xAA`.

## Related, found alongside

`P4025`'s diagnostic has a degenerate span and a truncated message. On `r := b.9;` with `b : BYTE`, the label points at `oob2.st:1:1` (`PROGRAM main`) and the text ends `"Valid range is 0..7 for type"` — the type name is missing. The `.%X9` spelling produces the identical span and identical truncation, so it is a general P4025 defect, not partial-access-specific. P4033 and P4028 both point at the offending token correctly.

garretfick:
## Summary

Fixes #1595.

The docs said `.%Bn`, `.%Wn`, `.%Dn` and `.%Ln` partial access was not yet supported and predicted a P0003 lexer error when the flag is on. The feature was already implemented on `main` (lexer, parser, AST, analyzer, codegen and plc2plc). While verifying that, two codegen bugs turned up in slice *writes*, which this PR also fixes.

## Documentation

- `docs/reference/language/structured-text/bit-access.rst` — describes the four wider forms as supported: result types, valid index range by base width, an example, and how plc2plc renders them.
- `docs/explanation/enabling-dialects-and-features.rst` and `docs/reference/compiler/ironplcc.rst` — the `--allow-partial-access-syntax` descriptions no longer say the wider forms are unsupported.
- `docs/reference/compiler/problems/P4033.rst` — the flag enables all five forms; the P0003 prediction is gone.
- `docs/reference/compiler/problems/P4025.rst` — now covers out-of-range slices, which raise the same code.
- `docs/reference/language/edition-support.rst` — lists all five forms.
- `ironplcc --help` text and the lexer rule's module comment named only the `iec61131-3-ed3` and `rusty` dialects; `codesys` and `twincat` enable the flag too.

## Codegen fix

`x.%Dn := v` and `x.%Ln := v` panicked with `attempt to shift left with overflow`, and a 64-bit slice could not receive a 64-bit value. Both defects are in `emit_partial_access_read_modify_write` (`compiler/codegen/src/compile_expr.rs`), shared by all four write paths:

- The slice mask was built with `1i32 << access_bits` (and `1i64 <<` in the 64-bit branch), which overflows when the slice is exactly as wide as the integer. Masks are now built at 128 bits and narrowed to the operand width.
- The right-hand side was always compiled at 32 bits, so an `LWORD` variable kept only its low half and an `LWORD` literal was rejected as a 32-bit constant with P2026. The value is now compiled at the slice's own width, unsigned, since a slice is a bit string.

Reads of every width, and 8- and 16-bit writes, were already correct.

## Tests and specs

- `compiler/codegen/tests/it/end_to_end_partial_access.rs` — prefactored from one `e2e_i32_with!` per case into two `rstest` tables over a shared program template (coverage unchanged), then extended with rows for `.%D` and `.%L` writes on `DWORD` and `LWORD` bases, a `DINT` base, a literal with its top bit set, and variable vs literal right-hand sides. Six of the new rows fail on `main`.
- `compiler/parser/src/tests/partial_access.rs` — one lexer test for the `%B`/`%W`/`%D`/`%L` selectors, which had none.
- `specs/design/partial-access-bit-syntax.md` — retires the "separate design" deferral, fixes the dialect list, and records multi-byte semantics as REQ-PAB-100 through 150 plus REQ-PAB-133 (value width), each tied to an existing test.

Not addressed here: the P4025 diagnostic for an out-of-range slice index points at the start of the file rather than the index token (noted in #1595).

## Verification

`cd compiler && just` (compile, coverage, lint) passes. The docs build with `-W -n` and the jscpd duplication check pass. `cd specs && just` passes.

https://claude.ai/code/session_01NDAmAdTNTbX3TV6cE99Qvb


docs/reference/language/structured-text/bit-access.rst
```diff
@@ -123,32 +127,96 @@ chosen bit index is preserved.
...
+An index outside the valid range, or a slice wider than the base type,
+raises :doc:`P4025 </reference/compiler/problems/P4025>`.
+
+Writing to a slice replaces only the bits of that slice; the other bits of
+the base value keep their values. Like ``.%Xn``, these forms are gated
+behind ``--allow-partial-access-syntax`` and raise
+:doc:`P4033 </reference/compiler/problems/P4033>` without it. Unlike
+``.%Xn``, they have no Edition 2 short form.
+
+.. code-block::
```
garretfick: This should be a playground example, not just a code block

docs/reference/language/structured-text/bit-access.rst
```diff
@@ -123,32 +127,96 @@ chosen bit index is preserved.
...
+       lo := packet.%B0;         (* 16#DD — least significant byte *)
+       hi := packet.%W1;         (* 16#AABB — upper 16 bits *)
+       packet.%B1 := 16#FF;      (* packet becomes 16#AABBFFDD *)
+   END_PROGRAM
 
-These forms return a non-``BOOL`` view of the underlying value and require
-additional codegen work. Until they land, a source that uses them raises
-:doc:`P4033 </reference/compiler/problems/P4033>` when the flag is off and
-``P0003`` (``Unmatched character sequence``) when the flag is on.
+The plc2plc renderer preserves these forms on output, since they have no
```
garretfick: Let's not mention plc2plc. That's primary for testing.

docs/reference/compiler/problems/P4025.rst
```diff
@@ -43,3 +46,25 @@ To fix this error, use an index within the valid range or use a wider type:
...
    END_FUNCTION_BLOCK
+
+Partial Access
+--------------
+
+The same check applies to the Edition 3 byte, word, double word, and long
+word forms. The slice must fit within the variable, and the index must not
+run past its last slice:
+
+.. code-block::
```
garretfick: this should be a playground example.

compiler/parser/src/rule_token_no_partial_access_syntax.rs
```diff
@@ -1,8 +1,8 @@
 //! Validation rule: reject partial-access syntax (`.%Xn`, `.%Bn`, `.%Wn`,
 //! `.%Dn`, `.%Ln`) unless the `allow_partial_access_syntax` flag is set.
 //! IEC 61131-3:2013 standardizes this form; IronPLC accepts it under
-//! `--allow-partial-access-syntax` (implied by the `rusty` and
-//! `iec61131-3-ed3` dialects).
+//! `--allow-partial-access-syntax` (implied by the `iec61131-3-ed3`, `rusty`,
```
garretfick: We should not mention specific dialects here. That's just gets out of date.

compiler/parser/src/tests/partial_access.rs
```diff
@@ -50,6 +50,37 @@ fn lexer_spec_req_pab_002_direct_address_still_takes_precedence() {
         .any(|t| t.token_type == TokenType::PartialAccessBit));
 }
 
+/// REQ-PAB-100: The lexer tokenizes `%B`, `%W`, `%D` and `%L` followed by
+/// digits as the byte, word, dword and lword partial-access selectors,
+/// case-insensitive.
+#[test]
```
garretfick: This looks like a spec test but it doesn't have the spec test annotation. Why not?

compiler/ironplc-cli/bin/main.rs
```diff
@@ -165,9 +165,10 @@ struct FileArgs {
     #[arg(long)]
     allow_cross_family_widening: bool,
 
-    /// Allow IEC 61131-3:2013 partial-access bit syntax (`.%Xn`) as an alias
-    /// for the short form `.n`. Enabled by `--dialect=iec61131-3-ed3` and
-    /// `--dialect=rusty`.
+    /// Allow IEC 61131-3:2013 partial-access syntax: the bit form `.%Xn` (an
```
garretfick, line 168: Don't mention the dialect here. I see that on others but that's an anti-pattern.

compiler/parser/src/tests/partial_access.rs
garretfick: The design doc predated cross-crate spec conformance and used unslugged `REQ-PAB-NNN` IDs with a naming convention (its Enforcement section said so), and the generator rejects any doc containing an unslugged ID, so the annotation had to be all-or-nothing for the doc.

Done in 34241a4: every requirement now carries its owning crate's slug (`REQ-PAB-parser-*`, `-analyzer-*`, `-codegen-*`, `-plc2plc-*`), the four build scripts list the doc, and every test including this one is annotated `#[spec_test]`. The codegen e2e binary includes the generated constants so the annotation resolves there. REQ-PAB-parser-101, -110 and -141 gained dedicated parser tests instead of leaning on e2e cases. All four `all_spec_requirements_have_tests` meta-tests and the workspace orphan guard pass.

---

compiler/codegen/src/compile_expr.rs
```diff
@@ -1519,13 +1539,13 @@ fn emit_partial_access_read_modify_write(
             emitter.emit_bit_or_64();
         }
         _ => {
```
garretfick: Can we list the specific OpWidth here. For example, I don't see why we would support F32 or F64 but those would hit this branch.

garretfick: Good catch, and it was live: `f : REAL; f.%B0 := 16#FF;` compiled today, because the analyzer's range rule only asks the base for its size and a float has one. Both the read and the write path now match `W32` explicitly and return an internal error at the access for `F32 | F64`, with a negative e2e table covering REAL and LREAL reads and writes (9d22bec).

The proper home for the rejection is the analyzer (a slice needs an integer or bit-string base), which would be a new problem code plus docs. I left that out of this PR; bit access `f.3` on a REAL has the same gap.

---

PR #1640, opened 2026-09-05
Fix P4043's documented remedy, its gate, and its span

### LINKED ISSUE #1543: P4043's documented fix does not work, and its gate misses a bare variable reference
Found while triaging `2026-07-26-twincat-struct-init-expression-value` (#1464). Reproduced.

## The documented remedy fails

`docs/reference/compiler/problems/P4043.rst` tells the reader:

> To fix this error, either use a constant value:
> ```
> FUNCTION_BLOCK FB_Example
> VAR
>     tonDelta : TON := (PT := T#100MS);
> END_VAR
> END_FUNCTION_BLOCK
> ```

Running exactly that declaration:

```
$ ironplcc check ton.st
error[P4012]: Function block invocation is not a variable in scope (invocation=tonDelta)
  ┌─ ton.st:6:1
6 │ tonDelta(IN := TRUE);
```

A `:=`-parenthesized initializer on a standard-library FB type makes the variable a **struct** variable rather than an FB instance, so it can no longer be invoked. Declaring it without invoking still fails at `compile` with `P9999 "Unknown structure type"` (`compile_struct.rs:557`).

So a user who hits P4043 and follows the page gets a different error and no working program. That is worse than no advice.

The underlying quirk — that a parenthesized initializer on an intrinsic FB type is parsed as a struct initializer and never reaches FB-instance codegen — is recorded nowhere in `specs/` or in any code comment. It belongs on `allocate_struct_variable`.

## The gate misses a bare variable reference

```iecst
TYPE MyStruct : STRUCT x : INT; END_STRUCT; END_TYPE
PROGRAM main
VAR
    g : INT;
    s : MyStruct := (x := g);
    r : INT;
END_VAR
r := 0;
END_PROGRAM
```

`ironplcc check` under the **default strict Edition 2 dialect** accepts this with no diagnostic. `g` is a runtime value by any reading, so P4043 should fire.

Cause: `enumerated_value()` matches the whole identifier and its trailing lookahead succeeds, so the PEG ordered choice locks in `EnumeratedValue` and never reaches the `expression()` alternative. Codegen later reports `P9999 "Unknown enum value"` (`compile_enum.rs:124`) — naming the wrong construct entirely.

This is the exact hazard the plan's own implementation note warned about. The lookahead fixed it for `pDevice^.Delta` but by construction cannot fix it for a bare identifier, and no comment records that residual hole.

## Two smaller documentation gaps

**The message does not match the rule.** P4043 refuses `(x := 1 + 1)` under the strict dialect while accepting `(x := 1)` and `(x := -1)`. The gate does not distinguish runtime values from foldable constants, but the diagnostic talks about *non-constant* expressions — so a user is told their constant expression is not constant. Worth a sentence in `P4043.rst` explaining the distinction from `--allow-constant-initializer-expressions`/P4037, which *does* fold.

**The motivating example needs a flag the docs do not mention.** `--allow-struct-initializer-expressions` is enabled for `[Rusty, Codesys, TwinCat]`, but `--dialect=twincat` deliberately does not enable `REF_TO`. So the `pDevice^.Delta` example in `enabling-dialects-and-features.rst:372` and `ironplcc.rst:245` is a `P0002` under the very dialect those pages associate with it; it additionally needs `--allow-ref-to` or `--dialect=codesys`.

## Span defect

Both P4043 and the codegen P9999 render their label starting at file offset 0 when the value is a compound expression — `expr.span()` on a `BinaryOp` or member-access chain yields a start of 0. The `EnumeratedValue` path is correct. Same family as #1537, #1539 and #1540; likely a general `ExprKind` span issue rather than anything this feature introduced.

garretfick:
Fixes [#1543](https://github.com/ironplc/ironplc/issues/1543).

Three defects, all reachable from the `:= (name := value)` initializer form. Rebased onto [#1654](https://github.com/ironplc/ironplc/pull/1654), which landed the declaration-level half of what this branch used to carry; the field-store extraction landed as [#1643](https://github.com/ironplc/ironplc/pull/1643).

## Relationship to #1654

#1654 established the principle (ADR-0050): the parser records a user-typed initializer as written and the type resolver classifies it. This branch had reached the same conclusion independently for `x : T := (a := 1)`, with a second placeholder variant beside the existing one. #1654's single `LateResolvedType(LateResolvedInitializer)` — covering `x : T`, `x : T := (a := 1)` and `x : T := Red` together — is the better shape, so this branch drops its own and builds on that one.

What remains here is the part #1654 does not cover, plus the codegen that makes its placeholder actually run.

## 1. The documented remedy did not work

`docs/reference/compiler/problems/P4043.rst` tells the reader to fix P4043 by using a constant value:

```iecst
PROGRAM main
VAR
    tonDelta : TON := (PT := T#100MS);
END_VAR
tonDelta(IN := TRUE);
END_PROGRAM
```

Before #1654 that failed with `P9999 "Unknown structure type"`. #1654 makes the declaration resolve to a function block instance, then rejects it at codegen with a not-implemented diagnostic rather than allocating the instance and dropping the values — a clearer failure, but still a failure: following the page does not get you a working program.

`FunctionBlockInitialValueAssignment::init` has been in the AST since the initializer was introduced but was never non-empty, so codegen ignored it. `emit_initial_values` now emits one store per member using the sequence extracted in #1643 — the same one `timer.PT := T#100MS;` emits — and the guard in `assign_variables` goes. An invocation argument still wins, because arguments are stored every scan and the initializer runs once.

## 2. The gate missed a bare variable reference

Under the default strict Edition 2 dialect this is accepted with no diagnostic at all:

```iecst
TYPE MyStruct : STRUCT x : INT; END_STRUCT; END_TYPE
PROGRAM main
VAR
    g : INT;
    s : MyStruct := (x := g);
END_VAR
END_PROGRAM
```

`g` is a runtime value by any reading, so P4043 should fire. Instead codegen later reported `P9999 "Unknown enum value"` — naming a construct the program does not contain.

This is the same guess #1654 removed, one level down: `(x := g)` is a value *inside* a member list, where `LateResolvedInitialValue::Value` is the whole initializer (`x : T := Red`). `structure_element_initialization()` still offered `enumerated_value()` before `expression()`, and a bare identifier is a complete match for the former, so the ordered choice locked in `EnumeratedValue` whether or not `g` named one.

So the same treatment: the parser emits `StructInitialValueAssignmentKind::LateBound`, and `xform_resolve_late_bound_expr_kind` resolves it with the enum-or-variable call it already makes for every other bare identifier — a variable becomes `Expression`, anything else an `EnumeratedValue`. A qualified `Type#VALUE` is unambiguous and gets its own node straight from the parser, via an `enumerated_value__qualified()` rule that cannot match a bare identifier.

P4043 now fires on it, and an undeclared name reports `P4007 "Variable not defined before used"` instead of a codegen failure.

## 3. Span defect

Both P4043 and the codegen P9999 rendered their label at file offset 0, with an empty file id, whenever the value was a compound expression:

```
error[P4043]: General expression as a struct/FB-instance initializer value ...
  ┌─ :1:1
1 │
  │ ^ Expression-valued struct/FB-instance initializer
```

`ConstantKind::span()` returned `SourceSpan::default()` for every literal kind except integers and bit strings, and the integer path was handed a default span by the parser. `Located for ExprKind` joins its operands' spans, so one span-less literal made the whole expression report `0..0`.

All nine literal kinds now carry a span, recorded once around the choice in `constant()`: the grammar takes the token slice as an argument so a rule can map the token indices `position!()` yields back to byte offsets. Keeping it at the choice rather than in each literal's own rule is what makes it impossible to add a tenth kind that is silently span-less. That also fixes `duration()`, which was building its span out of token indices as though they were byte offsets. Constant folding carries the folded expression's span onto the literal it produces, so a diagnostic on a folded value still points at the source.

```
   ┌─ span1.st:4:27
 4 │     s : MyStruct := (x := 1 + 1);
   │                           ^^^^^ Expression-valued struct/FB-instance initializer
```

## Documentation

**The message did not match the rule.** P4043 keys on the *shape* of the value, not on whether it could be computed at compile time: `(x := 1)` and `(x := -1)` are accepted, `(x := 1 + 1)` is not, even though its result is fixed. `P4043.rst` gains a section saying so, and distinguishing this gate from `--allow-constant-initializer-expressions`/P4037, which *does* fold. No behaviour change.

**The motivating example needed a flag the page did not mention.** The `pDevice^.Delta` example spells its pointer with `REF_TO`, which is a separate extension needing `--allow-ref-to`; not every dialect that enables one enables the other, and some spell it `POINTER TO`. The page says that now and links to the dialects page rather than naming dialects itself.

## Testing

- Parser: `g` and `RED` both parse to `LateBound` — the point being that the parser cannot tell them apart — and `Color#GREEN` to `EnumeratedValue`.
- Parser: all thirteen literal spellings record a span covering the literal as written.
- Analyzer: `(x := g)` produces P4043 under the default dialect and is accepted with the flag; `(c := RED)` is never flagged; an FB instance's member initializer reaches the same gate.
- Analyzer: the P4043 label for `(x := 1 + 1)` spans `1 + 1`, asserted against the source text.
- End to end on the VM: a `TON` declared `:= (PT := T#5s)` and invoked without `PT` keeps `Q` false at 2s and true at 6s; an invocation's own `PT` overrides it; a user-defined FB reads its initialized member.
- plc2plc round-trips for the bare-identifier and FB-instance-initializer spellings.

`#1654`'s `compile_fb_member_initializer.rs`, which asserted the not-implemented rejection, is removed: `end_to_end_user_fb` and `end_to_end_fb_ton` run the same shapes on the VM and check the values land.

`cd compiler && just`, `cd specs && just` and `cd docs && just ci` all pass.

## Not in scope

- Folding in the P4043 gate — `(x := 1 + 1)` stays a P4043; the distinction from P4037 is documented rather than removed.
- Expression-valued *struct field* initializers at codegen, which remain `not_implemented` (pre-existing, separately scoped).
- `ExprKind::UnaryOp` reports its operand's span, so `-42` outside a declaration initializer underlines `42`. Noted in a test comment; a separate truncation from the defect fixed here.

https://claude.ai/code/session_016bPBbrc1x2uLWY7S5vTVfn


compiler/analyzer/src/xform_resolve_late_bound_expr_kind.rs
```diff
@@ -166,6 +166,39 @@ impl DeclarationResolver<'_> {
 }
 
 impl Fold<Diagnostic> for DeclarationResolver<'_> {
+    /// Reclassifies a bare identifier used as a structure or function-block
+    /// member initializer value.
+    ///
+    /// `structure_element_initialization()` offers `enumerated_value()`
```
garretfick: Hold on. If we don't know if something is actually an enumerated value during parsing, then we should return a late bound item rather than re-interpret this as a struct element initialization. If we don't, then errors we report may suggest we have an enumerated value but that's because we haven't yet figure out what this is. I think we need a fix in the parser.

compiler/analyzer/src/xform_resolve_late_bound_type_initializer.rs
```diff
@@ -137,12 +137,55 @@ struct TypeResolver<'a> {
     diagnostics: Vec<Diagnostic>,
 }
 
+impl TypeResolver<'_> {
+    /// Returns whether `name` names a function block type.
+    ///
+    /// Covers all three places a function block type can come from: the
+    /// type environment (standard library blocks such as `TON`), the set of
+    /// standard types recognized but not yet supported, and the `FUNCTION_BLOCK`
```
garretfick: I've removed this idea of unsupported std lib types. We don't have that anymore.

I still don't understand why we need to look in both self.type_environment and self.types. The type environment should have all of the types. We shouldn't need more than one place.

compiler/analyzer/src/xform_resolve_late_bound_type_initializer.rs
```diff
@@ -137,12 +137,55 @@ struct TypeResolver<'a> {
...
+            // the parser cannot know whether `T` names a STRUCT or a
+            // function block, because no type declaration is in scope yet.
+            // A function-block type makes it an instance with member
+            // initial values -- `StructureInitializationDeclaration` and
+            // `FunctionBlockInitialValueAssignment` carry the same
+            // `Vec<StructureElementInit>` precisely because the two are
+            // the same construct. Without this the variable stays a
+            // structure, so it can be neither invoked (P4012) nor laid out
+            // (there is no such structure type).
+            InitialValueAssignmentKind::Structure(decl)
```
garretfick: Similar to before - should we instead create a "not yet known" type rather than parse this and re-interpret it? What's the consequence of doing that?

compiler/parser/src/parser.rs
```diff
@@ -337,7 +351,15 @@ parser! {
...
-    rule constant() -> ConstantKind =
+    // Every literal kind records the span of the tokens it matched. Doing it
+    // once here, rather than in each literal's own rule, is what guarantees
+    // no kind is left span-less: `Located for ExprKind` joins the spans of
+    // an expression's operands, so a single span-less literal makes the
+    // whole expression report position 0.
+    rule constant() -> ConstantKind = start:position!() c:constant__unspanned() end:position!() {
+        c.with_span(span_of_tokens(tokens, start, end))
+    }
+    rule constant__unspanned() -> ConstantKind =
```
garretfick: Why do we want an un-spanned rule? This seems unnecessarily complex. Can we just have the constant() rule create the item with the span?

docs/reference/compiler/problems/P4043.rst
```diff
@@ -36,7 +36,12 @@ The following code will generate error P4043:
    END_VAR
    END_FUNCTION_BLOCK
 
-To fix this error, either use a constant value:
+``REF_TO`` is itself a dialect extension, so this example additionally needs
```
garretfick: It isn't a dialect extension - it is a syntax extension. Double check the right way to refer to this from the glossary.

docs/reference/compiler/problems/P4043.rst
```diff
@@ -36,7 +36,12 @@ The following code will generate error P4043:
...
+``--allow-ref-to`` (or a dialect that enables it, such as
+``--dialect=codesys``). Under ``--dialect=twincat``, which does not enable
```
garretfick: We don't try to mention dialects here because that gets out of date. Give a pointer to the appropriate page.

garretfick (commented):
See comments

compiler/analyzer/src/xform_resolve_late_bound_expr_kind.rs
garretfick: Agreed, and fixed in the parser.

`structure_element_initialization()` now has a `LateBound` alternative for a bare identifier, and `StructInitialValueAssignmentKind` has a matching variant. A qualified `Type#VALUE` keeps `EnumeratedValue` via a new `enumerated_value__qualified()` rule, which cannot match a bare identifier — so the ambiguous case and the unambiguous one are different nodes rather than the same one with a post-hoc correction.

The resolution in `xform_resolve_late_bound_expr_kind` now consumes `LateBound` and produces either `EnumeratedValue` or `Expression`, using the same `resolve_late_bound` call it already makes for every other bare identifier. Codegen treats a surviving `LateBound` as an internal error rather than as a diagnostic about the user's program, since it means the pass did not run.

Three parser tests pin the shapes: `g` and `RED` both parse to `LateBound` (the point being that the parser genuinely cannot tell them apart), and `Color#GREEN` parses to `EnumeratedValue`.

---

compiler/analyzer/src/xform_resolve_late_bound_type_initializer.rs
garretfick: You're right that one place is enough. `is_function_block_type` is now just:

```rust
self.type_environment
    .get(name)
    .is_some_and(|ty| ty.representation.is_function_block())
```

`xform_resolve_type_decl_environment::fold_function_block_declaration` registers every `FUNCTION_BLOCK` declaration before this transform runs, so the environment answers for user-declared blocks as well as standard library ones. The `self.types` lookup answered nothing the first one did not.

What made it look otherwise was my own tests: they called `apply` against a `TypeEnvironmentBuilder` assembled by hand, which of course had no user function block in it, so dropping the second lookup failed them. That was testing the assembly rather than the resolution. They now go through `parse_and_resolve_types`, and the end-to-end codegen test that runs a user FB with a member initializer passes either way.

On unsupported standard library types — I've dropped the `is_unsupported_standard_type` call from this function, so nothing I added depends on it. Flagging one thing though, in case it is news: the concept is still on `main` as of `f6c1559` — `analyzer/src/stdlib.rs:55`, used by `rule_unsupported_stdlib_type.rs:61` and by the pre-existing `LateResolvedType` arm in this same file (`xform_resolve_late_bound_type_initializer.rs:175`). If removing it is in flight somewhere I haven't seen, that last call site is the one that will need attention; I haven't touched it.

---

compiler/analyzer/src/xform_resolve_late_bound_type_initializer.rs
garretfick: Done the same way, and the consequences turned out small enough that I implemented it rather than just answering.

`x : T := (a := 1)` now parses to `InitialValueAssignmentKind::LateResolvedTypeInit(LateResolvedTypeInitializer { type_name, elements_init })` instead of `Structure`, and this transform turns it into `Structure` or `FunctionBlock`. The parser no longer claims it is a structure and get corrected afterwards — which matters here more than anywhere, because that claim is the bug this PR started from: a `TON` written this way was reported as `P9999 "Unknown structure type"`.

**Consequences, measured rather than guessed.** Adding the variant broke exactly seven exhaustive matches:

| Site | Sees it? | Handling |
|---|---|---|
| `dsl::VarDecl::type_name` | yes | the type name, same as `Structure` |
| `xform_toposort_declarations` | **yes**, runs before resolution | dependency edge on the named type — without it a forward reference would surface as a spurious P2011 |
| `xform_resolve_late_bound_expr_kind::insert` | **yes**, runs before resolution | `VariableType::LateResolvedType` — the type is no more known here than for a bare `x : T` |
| `intermediates::structure::field_has_default` | no | explicit member values are a default by definition |
| `xform_resolve_expr_types` | no | the type name |
| `rule_var_decl_const_initialized` | no | internal error, beside the existing `LateResolvedType` arm |
| `xform_resolve_type_decl_environment` | no | aliases like the `Structure` arm it mirrors; only a VAR declaration produces this shape, `initialized_structure()` still yields `Structure` for a TYPE declaration |

The two that genuinely see it are the two passes that run before resolution, and both need the type name rather than the kind, so neither had to guess.

One thing that did need care: plc2plc renders the parse tree, so `LateResolvedTypeInit` is what it sees for *every* such declaration. It has its own `visit_late_resolved_type_initializer` spelling it identically to the structure form; the round-trip tests caught the omission.

---

compiler/parser/src/parser.rs
garretfick: No good reason — the extra rule was scaffolding I left in. `constant()` now takes the span itself, with the alternatives inlined in a parenthesized group:

```rust
rule constant() -> ConstantKind = start:position!() c:(
    real:real_literal() { ConstantKind::RealLiteral(real) }
    / integer:integer_literal() { ConstantKind::IntegerLiteral(integer) }
    / ...
) end:position!() {
    c.with_span(span_of_tokens(tokens, start, end))
}
```

`constant__unspanned()` is gone.

Keeping the span assignment at the choice rather than pushing it into each literal's own rule is deliberate: those rules are reached from other positions too (`signed_integer()` from subranges and case labels, `duration()` directly), and one assignment around the choice is what makes it impossible to add a tenth literal kind that is silently span-less.

---

docs/reference/compiler/problems/P4043.rst
garretfick: Checked the glossary — it is explicit that "dialect extension" is wrong: *"Do not attach the 'dialect' qualifier to an extension: an extension is a syntax feature; a dialect is a bundle of them."* The default word is just **extension**, with **language extension** only when the bare word would collide with the extension library or the editor extension. Neither is ambiguous here, so it now reads "``REF_TO`` is itself an extension".

Also grepped my other additions on this branch for `dialect extension` / `vendor extension` — no other occurrences.

---

docs/reference/compiler/problems/P4043.rst
garretfick: Removed. The paragraph now says only what stays true, and points at the dialects page:

> ``REF_TO`` is itself an extension, so this example needs ``--allow-ref-to`` as well as the flag below. Not every dialect that enables one enables the other, and some spell the pointer ``POINTER TO FB_Device`` instead — see :doc:`/explanation/enabling-dialects-and-features` for which dialect enables which flag.

I applied the same reasoning to `ironplcc.rst`, where I had also named a dialect — that page names one in only one other place, so the mapping clearly isn't meant to live there either. On `enabling-dialects-and-features.rst` I dropped the specific names too and pointed at the per-dialect lists at the top of that same page, since those already carry the mapping and my sentence was a second copy of it.

---

compiler/analyzer/src/xform_resolve_late_bound_type_initializer.rs
garretfick: Update after rebasing onto #1654: `is_function_block_type` is gone entirely — the resolver now uses #1654's `classify`, which reads the type environment first and falls back to this pass's own table in the same order the bare-declaration arm already used. Your point stands and is now satisfied by that one function rather than by mine.

---

compiler/analyzer/src/xform_resolve_late_bound_type_initializer.rs
garretfick: Superseded — ignore the variant names in my previous reply.

#1654 landed the same answer to your question at this level, and did it better: one `LateResolvedType(LateResolvedInitializer)` covering `x : T`, `x : T := (a := 1)` and `x : T := Red` together, where I had added a second variant beside the existing one. This branch drops `LateResolvedTypeInit`, its payload struct, the seven consumer arms and `visit_late_resolved_type_initializer`, and builds on #1654's placeholder instead. The consequence table above was measured against my variant and no longer describes the PR.

What survives here is the level #1654 doesn't reach: `StructInitialValueAssignmentKind::LateBound` for `(x := g)`, the value *inside* a member list, where `LateResolvedInitialValue::Value` is the whole initializer. That's thread [#discussion_r3942664584](https://github.com/ironplc/ironplc/pull/1640#discussion_r3942664584), and it's still needed — without it the P4043 gate misses a bare variable reference.

---

PR #1641, opened 2026-09-05
Resolve a conversion's string encoding from its return type

garretfick:
Fixes a regression on `main`: a conversion feeding a string operation does not compile.

```
$ ironplcc compile conv.st -o conv.iplc     # out := CONCAT(INT_TO_STRING(i), 'x');
error[P9998]: Internal error indicating a bug in the compiler
  ┌─ conv.st:6:17
6 │   out := CONCAT(INT_TO_STRING(i), 'x');
  │                 ^^^^^^^^^^^^^ Cannot determine the string encoding of a function call that does not return a string
```

## Cause

`string_expr_char_width` (added in #1610) works out what encoding a string expression yields, so `resolve_string_arg` can allocate its temporary slot at that width rather than always narrow. Its function arm answers for two kinds of call — a standard function that passes its argument's encoding through, and a user-defined function that declares its return type — and reports P9998 for everything else. A conversion is everything else.

## Fix

A conversion builds a Latin-1 string and says so in its return type, so fall back to the type the analyzer resolved for the call. A call whose resolved type is not a string type at all still reports P9998, which is the case that message actually describes.

Ten lines in `function_char_width`, plus a small helper and an end-to-end test.

## Notes for review

- **No plan file** — this is the "single-line bug fix" exemption in the workflow: one fallback arm in one function.
- The test lives in `end_to_end_conv_string.rs` next to the other conversion tests, and runs the program rather than inspecting bytecode: `out` reads back as `'42x'`.
- This is the first of five PRs splitting #1607 apart for review. It is independent of the other four and can land on its own.

`cd compiler && just` passes.

---


compiler/codegen/src/compile_string.rs
```diff
@@ -220,6 +228,20 @@ fn function_char_width(ctx: &CompileContext, func: &Function) -> Result<CharWidt
...
+/// string type at all.
+fn resolved_string_char_width(expr: &Expr) -> Option<CharWidth> {
+    match expr
+        .resolved_type
+        .as_ref()
+        .and_then(|t| ElementaryTypeName::try_from(&t.name).ok())?
+    {
+        ElementaryTypeName::STRING => Some(CharWidth::Narrow),
+        ElementaryTypeName::WSTRING => Some(CharWidth::Wide),
+        _ => None,
```
garretfick: Can we at least log this an an error? We should never hit this.

garretfick: It already is an error — the `None` doesn't get swallowed. It flows into the `ok_or_else` right above and becomes **P9998**, which stops the compile and points at the call:

```
error[P9998]: Internal error indicating a bug in the compiler
  ┌─ prog.st:6:17
6 │   out := CONCAT(INT_TO_STRING(i), 'x');
  │                 ^^^^^^^^^^^^^ Cannot determine the string encoding of a function call that does not return a string
```

(That's the exact output this PR fixes, from before the change.) codegen has no logging crate — `log::`/`tracing::` appear nowhere in it — so P9998 is the mechanism here, and it's stronger than a log line: a log scrolls past, this fails the build.

What you're pointing at was weak for a different reason, though, and I've fixed that. `None` had **three** causes, and all three produced the same message, which is only true for one of them:

| Cause | Old message | New message |
|---|---|---|
| The analyzer left the call untyped | "does not return a string" | "a function call the analyzer left untyped" |
| Typed as something `ElementaryTypeName` doesn't name | "does not return a string" | "a function call the analyzer typed as `<name>`" |
| Typed as an elementary type that isn't a string | "does not return a string" ✓ | "a function call the analyzer typed as INT" |

The first two asserted knowledge we didn't have. Since the whole point of an internal error is to be debugged from, it now names what it found. `resolved_string_char_width` returns `Result` instead of `Option` so the reason is recorded where it's known, rather than reconstructed by a caller that can't tell the three apart.

And I added tests pinning the never-happen path — that it stops the compile with P9998, and that the message carries the offending type name — so the guarantee is enforced rather than hoped for.

Pushed as `e054088`. Say the word if you'd still like a log line on top of the hard error.

---

garretfick: On the WSTRING question (answering here since it's the same code): the *mapping* is tested — `resolved_string_char_width_when_wstring_then_wide`, added with the other unit tests. What is missing is an end-to-end WSTRING test of this arm, and the reason is that **no program reaches it**. I checked all three routes:

| Route to the fallback with a wide type | Result |
|---|---|
| A `*_TO_WSTRING` conversion | Does not exist — `parse_string_conversion` only builds `*_TO_STRING` (Latin-1) and `STRING_TO_*` |
| `SEL` / `MUX` returning a string | Rejected before codegen: `P4026 … expected=ANY_NUM, actual=wstring` |
| A user function returning `WSTRING` | Answered one branch earlier, by its declared return type — never reaches the fallback |

So `WSTRING => Wide` is there because that is what the name means, not because anything currently produces it. I left it in — a `*_TO_WSTRING` would otherwise land on a confusing internal error — and the arm now says this outright, so the next reader doesn't go looking for the program that exercises it.

The wide path a program *does* reach is the user-function branch, and you're right that it was untested end to end. Added:

```
FUNCTION mk : WSTRING …

out := CONCAT(mk(w), "cd");   (* reads back "abcd", char_width 2 *)
```

Pushed as `7b4865e`, rebased onto your merge of main. Full `just` passes, now including the no_std gate from #1644.

---

Issue #1612, opened 2026-09-02
Fold LEN of a constant string at compile time


garretfick:
Raised from review feedback on #1610.

`LEN` of a string literal compiles to the same sequence as `LEN` of a variable: allocate a data region slot, `STR_INIT` it, load the literal into a temp buffer, `STR_STORE_VAR` it into the slot, then `LEN_STR` to read `cur_length` back out.

```iecst
result := LEN('Hello');
```

The answer is a constant. Nothing about the program can change it, so none of that needs to run — the whole expression should fold to a single integer constant, saving the data region slot, the temp buffer and five instructions per call site.

## Scope

The same reasoning covers any string the compiler knows cannot change:

1. **String literals** — `LEN('Hello')`, and by extension `LEN(CONCAT('ab', 'cde'))`, where the argument is built entirely from literals.
2. **`CONSTANT`-qualified strings** — a `VAR CONSTANT` string's length is fixed at its declaration.
3. **Strings that are never written** — a variable with an initial value and no assignment anywhere in the program is constant in fact, even without the qualifier.

Case 3 is best handled not as a special case in `compile_len` but as a transform that marks such variables `CONSTANT` before codegen runs, so codegen only ever has to understand cases 1 and 2. That keeps the constant-detection reasoning in one place and out of the code generator, and it would benefit every other constant-folding opportunity, not just `LEN`.

## Notes

- `compiler/codegen/src/compile_string.rs` — `compile_len`, and `resolve_string_arg`, which materializes the argument into a data region slot.
- There is existing constant-folding machinery for binary expressions to model this on.
- Worth confirming the folded result matches the runtime one for a `WSTRING` literal: `LEN` counts UTF-16 code units, not characters or bytes, so folding must count code units too.

PR #1648, opened 2026-09-05
Mark never-written variables as CONSTANT in the analyzer

garretfick:
## Summary

Case 3 of #1612. A variable with an initializer that the program never writes holds its initial value for the whole run. A new analyzer transform, `xform_mark_unwritten_constants`, gives such declarations the `CONSTANT` qualifier at the end of type resolution, so the semantic rules and code generation see one notion of a constant variable. Code generation can then fold reads of it, starting with `LEN` of a constant string, without reasoning about writes itself. The `LEN` fold is a follow-up; this PR only establishes the qualifier.

## Stack

Targets `main`. The prerequisites have merged: #1654 (the resolver classifies user-typed initializers) and #1653 (P4009 collects globals only). The branch is rebased onto `main` after #1653, so the diff shows only the transform.

## Changes since the last review

- **Writes resolve to one declaration.** A write is resolved to the declaration the name reaches where it occurs: the symbol environment for the enclosing units (a method's declaration over its block's), `callee_resolution` for the block in the `EXTENDS` chain that declares an inherited field and for an instance's type, and the global through `VAR_EXTERNAL`. Only that declaration is blocked. A same-named declaration in another unit is unaffected. Only a target the transform cannot resolve to one declaration (a member through an array element or nested field, a configuration path, a constructor argument) blocks every declaration of the name.
- **The type environment is the one authority on whether a type is a function block.** The transform holds no table of its own.
- **`has_initial_value` and `input_expr` are methods on the DSL types**, not bare functions in the transform.
- **The design doc** states that the transform is an optimization a non-optimized or debug build may leave out, and that gating it is one call in `stages::resolve_types`.
- The by-name workaround for the P4009 rule is gone; #1653 fixes the rule instead.

## Design

`specs/design/constant-variable-inference.md` records the decisions and carries `REQ-CVI-analyzer-*` requirements with conformance tests.

- **Every construct that can change a variable counts**: assignment to any part of it, `FOR` control variables, `=>` outputs, `VAR_IN_OUT` arguments, `REF`/`ADR`/`REF=`, instance and `VAR_CONFIG` member initializers, program connection sinks, writable `VAR_ACCESS` paths, and SFC action associations. A callee whose parameter directions cannot be resolved is treated as writing every variable argument.
- **`VAR_INPUT` arguments stay reads**, so `LEN(msg)` and `timer(PT := delay)` leave `msg` and `delay` constant.
- **Only declarations the existing `CONSTANT` rules accept are marked**, so P4008, P4009 and P4010 are never introduced. Globals and their externals are marked together. Structures, references, located, retained and parameter declarations are left alone.

## Contract for later stages

After `resolve_types`, a `CONSTANT` qualifier means "holds its initial value for the whole run" whether it came from source or from this transform. A build that must let the operator change variables should run without the transform.

## Verification

`cd compiler && just` passes on the rebased branch (build, coverage, clippy, fmt, duplication check). Unit tests cover the corner cases; the conformance tests pin each requirement.

https://claude.ai/code/session_01PyMa6n6EDcNK6178uDdPUt


specs/design/constant-variable-inference.md
```diff
@@ -0,0 +1,168 @@
...
+
+After `stages::resolve_types` returns, a `VarDecl` whose qualifier is
+`DeclarationQualifier::Constant` holds its initial value for the whole
+run, whether the qualifier came from the source or from this transform.
+Code generation may read the initializer instead of the variable's storage
+and may skip the storage altogether when nothing else needs it.
+
+Two consequences follow, and both are deliberate:
+
+- A tool that lets the operator change a variable while the program runs
```
garretfick: This is true, but we could also run the code generation without this step. That's what a debug/no-optimized build would do. This isn't an all or nothing. We could configure this behavior.

compiler/analyzer/src/xform_mark_unwritten_constants.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+/// Returns whether `init` is an initializer that `rule_var_decl_const_initialized`
+/// accepts on a `CONSTANT` declaration without any further check.
+///
+/// Exhaustive on purpose: a new initializer kind must decide here whether a
+/// never-written variable of that kind may be marked.
+fn has_constant_initializer(init: &InitialValueAssignmentKind) -> bool {
+    match init {
+        InitialValueAssignmentKind::Simple(si) => si.initial_value.is_some(),
```
garretfick: This is a bare function. Should this be part of InitialValueAssignmentKind's impl block? Same for the next function.

compiler/analyzer/src/xform_mark_unwritten_constants.rs
```diff
@@ -0,0 +1,539 @@
...
+
+/// Gathers the name of every variable the library can write, and the global
+/// and external declarations the marking has to keep consistent.
+struct WriteCollector<'a> {
+    function_blocks: &'a FunctionBlocks<'a>,
+    type_environment: &'a TypeEnvironment,
+    function_environment: &'a FunctionEnvironment,
+    /// The function-block instances of the unit being walked.
+    instances: InstanceTypes,
+    written: HashSet<Id>,
```
garretfick: Something isn't right with this. `Id` has scope. I think this is saying we don't understand scopes for determining what's written.

compiler/analyzer/src/xform_mark_unwritten_constants.rs
```diff
@@ -0,0 +1,539 @@
...
+        match record {
+            SymbolicVariableKind::Named(named) => self.instances.type_of(&named.name).is_some(),
+            _ => true,
+        }
+    }
+
+    /// Whether `type_name` names a function block, declared in the library
+    /// or supplied by the standard library.
+    fn is_function_block(&self, type_name: &TypeName) -> bool {
+        self.function_blocks.contains(type_name)
```
garretfick: Why would we need both here? This sounds like we don't trust the type environment.

specs/design/constant-variable-inference.md
garretfick: Agreed. Reworded this section: the transform is an optimization a non-optimized or debug build leaves out, so a build that lets the operator change variables runs without it and never reads a folded value that disagrees with storage. Gating it is one call in `stages::resolve_types`; the doc says that is not wired to a build option yet.

---

compiler/analyzer/src/xform_mark_unwritten_constants.rs
garretfick: Moved: `InitialValueAssignmentKind::has_initial_value()` in the DSL says whether a declaration states a value; the transform layers only its own policy on top (structures and references stay unmarked, for the reasons in the design doc). The argument helper is `ParamAssignmentKind::input_expr()`.

---

compiler/analyzer/src/xform_mark_unwritten_constants.rs
garretfick: You're right, it was by name. Writes now resolve to the declaration they reach: the symbol environment answers which scope a name reaches from the current one (method over block over global, `VAR_EXTERNAL` to its global), `callee_resolution` answers which block in the `EXTENDS` chain declares an inherited field and what type an instance has for `inst.member` writes. The written set is keyed by `(ScopeKind, Id)` and only that declaration is blocked; a same-named variable in another unit is unaffected. Only a target that does not resolve to one declaration (a member through an array element or nested field, a `VAR_CONFIG` path, a constructor argument) blocks every declaration of the name. REQ-021 and its test now say this.

---

compiler/analyzer/src/xform_mark_unwritten_constants.rs
garretfick: It was distrust, and it was covering for a declaration that was mis-shaped at the source: the type resolver left `inst : FB := (x := 1)` as a structure initializer, so consumers had to guess. #1654 fixes that in the resolver, which also removes the predicate #1649 added to `InstanceTypes::declare`. Here the type environment alone answers whether a type is a function block; the library lookup is only for the declaration's parameters, which the type environment does not hold.

---

compiler/analyzer/src/xform_mark_unwritten_constants.rs
```diff
@@ -0,0 +1,599 @@
...
+                    self.mark_member(&fb_type, field);
+                }
+            }
+            SymbolicVariableKind::SelfRef(_) => self.mark(field),
+            _ => self.mark_any_scope(field),
+        }
+    }
+
+    /// Marks every member an instance initializer sets, at any depth.
+    fn mark_member_inits(&mut self, fb_type: &TypeName, inits: &[StructureElementInit]) {
```
garretfick, line 241: I'm unclear why we need to mark member inits. If we initialized and never changed, wouldn't that be a constant?

Issue #1550, opened 2026-08-30
Comparing a WSTRING to a literal traps at runtime, which V9014.rst says cannot happen


garretfick:
Found while triaging `2026-06-11-wstring-end-to-end` (#1464). Reproduced with a control.

```iecst
PROGRAM main
VAR
  w : WSTRING[10] := "abc";
  eq : BOOL;
END_VAR
  eq := w = "abc";
END_PROGRAM
```

```
$ ironplcc compile wcmp.st -o wcmp.iplc      # compiles clean
$ ironplcvm run --scans 1 --dump-vars=- wcmp.iplc
w: 0
eq: FALSE
Error: V9014 - runtime error: string encoding mismatch: expected char_width 2, got 1
```

**Control** — the identical program with the literal moved into a second WSTRING variable:

```
eq: TRUE
```

So variable-to-variable works; variable-to-literal traps.

## The documentation says this is impossible

`docs/reference/runtime/problems/V9014.rst`:

> The analyzer rejects cross-encoding assignments and operations statically, so this trap is reached only via: 1. **A compiler bug** … 2. **A tampered or hand-crafted `.iplc` file** … This error should not occur during normal operation.

A six-line program written the obvious way reaches it. Meanwhile `wstring.rst` presents WSTRING as supported and shows `"…"` literals, so nothing warns the user off.

## Cause

The analyzer rule is explicit that it does not cover this — `compiler/analyzer/src/rule_string_encoding_compat.rs:10`:

> It does not flag string literals … those narrower cases rely on the runtime trap.

And codegen hard-codes narrow for every scratch slot a string operand passes through: `compile_string.rs:122,132,142,171` all pass `NARROW_CHAR_WIDTH`, including the `ExprKind::Const(CharacterString)` arm. `CMP_STR` then verifies both header widths and traps (`vm.rs:1465`).

## Why no test caught it

`end_to_end_wstring.rs` compares only variable-to-variable (`:134`), and `FIND`/`CONCAT` likewise use variables. The literal-operand case is untested — consistent with the plan's own "out of scope: WSTRING literals as function-call arguments / comparison operands", which is a scope note that never became a diagnostic.

## Fix shape

Either widen `resolve_string_arg` to take the peer operand's width, or extend `rule_string_encoding_compat` to flag literal-vs-WSTRING operands with P4034 at compile time. Either way `V9014.rst`'s "should not occur during normal operation" needs correcting.

## Same root cause, cosmetic — RESOLVED in #1575

[...]

PR #1655, opened 2026-09-06
Resolve one encoding for a string operation's operands (#1550)

garretfick:
Closes the rest of #1550.

The literal case the issue reported already works — #1610 taught the width inference to read a literal's own spelling, so `w = "abc"` compiles and compares wide. What still failed is every operation whose operands disagree about their encoding. Measured on `3198fa0` before this change:

| Program | Before | After |
|---|---|---|
| `eq := w = 'abc';` | `V9014` trap, scan 1 | `P4034` at compile time |
| `out := CONCAT(s, w);` | `V9014` trap | `P4034` |
| `pos := FIND(w, 'cd');` | `V9014` trap | `P4034` |
| `eq := s = "abc";` | `V9014` trap | `P4034` |
| `eq := w = "abc";` | works | works |

Each of the first four is a type error the compiler should have caught. A literal's delimiter is what types it — `'abc'` is a `STRING` and `"abc"` a `WSTRING` (IEC 61131-3 Table 5) — and P4034 exists to reject exactly this mixing.

## Why they compiled

Each operand was resolved on its own. `resolve_string_arg` asked `string_expr_char_width` what *that* operand yields and allocated its temporary at that width, so the two operands of one `CMP_STR` could be given different widths with nothing ever comparing them. The VM found out one scan later.

## Change

An operation resolves **one** encoding for all of its operands:

- `resolve_operand_char_width` asks each operand and requires agreement. Operands that disagree have no encoding they can share — P4034, with the span of the operand that differs.
- `resolve_string_arg` takes that resolved width from its caller instead of asking per operand.
- `compile_string_value` is the other half: a value written into a *declared destination* — assignment target, array element, structure field, parameter copy-in — is encoded **for** that destination rather than checked against it. That is what lets a literal adapt where there is something to adapt to.

Call sites that resolve a group: `LEN`, comparison, `FIND`, `CONCAT`, `REPLACE`, `INSERT`, and the one-string helpers behind `LEFT`/`RIGHT`/`MID`/`DELETE`. The parameter copy-in passes the parameter's declared width; `STRING_TO_*` passes narrow, so a `WSTRING` argument to it is P4034 rather than a trap.

## Notes for review

- **Documentation.** `V9014.rst` claimed the analyzer rejects cross-encoding operations statically — the claim that made the reported trap look impossible. That is true for operands now, so the page says what actually reaches the trap. `P4034.rst` and `wstring.rst` gain the rule that a literal's quotes are its type.
- **Not in scope:** `w := "abc"` as a statement is rejected by the analyzer with P4035, because it types every character-string literal as `STRING` whatever its quotes. That predates this work. It does mean the destination-encodes-a-literal path is only reachable for `STRING` targets today, which `compile_string_value` documents.
- Fourth of the pieces splitting #1607, after #1641, #1642 and #1650. One remains: the R0304 verifier.

`cd compiler && just`, `cd specs && just`, and the docs build all pass.

---


compiler/codegen/tests/it/end_to_end_wstring.rs
```diff
@@ -429,3 +428,153 @@ END_PROGRAM
     assert_eq!(read_char_width(&bufs.data_region, out_offset), 2);
     assert_eq!(read_wstring(&bufs.data_region, out_offset), "abcd");
 }
+
+// =========================================================================
```
garretfick, line 432: We need a custom test that takes in two arguments - one that is string and one that is wstring. Return boolean if the lengths are the same (or anything really as return). We need to be careful that these are not const otherwise the optimizer might just remove them entirely so somehow we need to manipulate the strings.

garretfick: Added, as `function_when_parameters_are_string_and_wstring_then_each_copied_at_its_own_width`:

```iecst
FUNCTION same_len : BOOL
  VAR_INPUT
    s : STRING[10];
    w : WSTRING[10];
  END_VAR
  same_len := LEN(s) = LEN(w);
END_FUNCTION

PROGRAM main
  VAR
    s : STRING[10] := 'abc';
    w : WSTRING[10] := "wx";
    other : WSTRING[10] := "z";
    same : BOOL;
    differ : BOOL;
  END_VAR
  s := CONCAT(s, 'd');          (* built at run time, not a constant *)
  w := CONCAT(w, "yz");
  same := same_len(s, w);       (* 4 = 4  -> TRUE  *)
  differ := same_len(s, other); (* 4 = 1  -> FALSE *)
END_PROGRAM
```

Both arguments are `CONCAT`'d before the call, so what reaches the parameters is computed rather than the declared initial value — that's your not-const point. And `differ` gives a second call with a different answer, so a constant `TRUE` wouldn't satisfy it either.

**I checked it actually catches the bug**, because a passing test proves nothing on its own. Forcing the copy-in back to `NARROW_CHAR_WIDTH` (the pre-change behaviour):

```
called `Result::unwrap()` on an `Err` value: Diagnostic { code: "P4034",
  described: ["expected=STRING", "found=WSTRING"] }
```

Worth recording that my first attempt at that check was wrong: `cargo fmt` had collapsed the call to one line, my search string still had the multi-line form, and the substitution silently did nothing — so the test appeared to pass with and without the fix. It was a no-op edit, not a vacuous test. Second attempt asserted the pattern matched first.

One thing I couldn't do: pass a literal for the wide parameter. `same_len(s, "y")` is rejected with `P4026 … expected=wstring, actual=STRING`, because the analyzer types every character-string literal as `STRING` whatever its quotes — the same pre-existing typing that produces the `P4035` on `w := "abc"`. So the test uses a variable, and that limitation is noted in the commit.

Pushed as `6809031`.

---

PR #1663, opened 2026-09-07
Enforce full semantic analysis in codegen tests

garretfick:
## Summary

Codegen tests reassembled the compiler pipeline themselves, and got it wrong: the harness ran `resolve_types` and skipped the semantic rules, while `ironplcc` reaches codegen only when `semantic()` reports nothing. A codegen test could therefore assert bytecode for a program the shipped compiler refuses to compile. That is how the WSTRING literal typing bug (#1550) survived a full end-to-end suite.

The harness now calls `ironplc_project::compile::compile` — the same function the CLI calls — over a `MemoryBackedProject`, the way `project`'s own tests do. Asserting the current rule instead would have repeated the original mistake one level up: a harness that checks `analyze` has no diagnostics drifts again the next time the pipeline grows a step. Calling the production entry point means the tests cannot diverge from the compiler, because they *are* the compiler.

## Changes

- `parse()` and the compile helpers go through `ironplc_project::compile::compile`, with `ironplc-project` added as a **dev**-dependency of `ironplc-codegen` (a cycle, which Cargo permits for dev-dependencies).
- The error stays a single `Diagnostic` — the first the pipeline reports, as a CLI user sees it — so tests asserting a rejection by `.code` keep working, and now catch analyzer rejections rather than only codegen ones. That alone fixed 7 tests that asserted an error the old harness never ran the rule for.
- 67 fixtures corrected: bit-string tests needing `--allow-cross-family-widening` (ADR-0031), boolean/comparison tests declaring a `DINT` target and assigning a `BOOL`, and `SEL` tests passing `0`/`1` where `G` is `BOOL`.

## Status: 27 failing, not mergeable yet

Down from 94. Every one is a program `ironplcc` rejects — the point of the change, not a regression it introduced. Both red checks are the same failure; `Compiler Quality Checks` runs `just ci`, which includes coverage.

The 27 are now tracked as four independent issues:

| Issue | Kind | Tests |
| --- | --- | --- |
| #1672 — `VAR_IN_OUT` unusable (call rejected; also passed by value) | bug ×2 | 3 |
| #1673 — `ARRAY[0..0]` rejected with P2002 | bug | 5 |
| #1674 — constant-overflow semantics: fold-and-truncate or reject? | **decision** | 10 |
| #1675 — fixtures asserting programs `ironplcc` rejects | test-side | 9 |

### Correction

An earlier version of this description credited #1665 (bit-string operators held to `ANY_BIT`) with 3 of the 27. **That was wrong.** Measured after #1665 merged and this branch was rebased on it, the count is unchanged at 27 and the failing set is byte-identical.

The three `NOT`-over-`DINT` fixtures were never *in* the 27 — they had been reverted to `DINT`, which the old analyzer accepted, so they passed. What #1665 actually did for this PR is prevent three *new* failures: it made the analyzer reject `NOT` over `DINT`, and updated those same fixtures to `BOOL` in the same commit, so the two cancel. Valuable, but not progress against this counter.

#1672 and #1673 are the fixes that will actually move it, to 19. The remaining 19 need the #1674 decision and the #1675 fixture work.

## Merge order

This PR should land after those, rebased on them.


garretfick:
## Status: Compiler Tests stays red, and cannot go green from this PR alone

`Compiler Quality Checks` is fixed as of 33ca227 (import ordering, an unused import, a needless clone).

`Compiler Tests` is still red: **27 of 1301 codegen tests**, down from 94 when the harness change landed. Every one is a program `ironplcc` refuses to compile — which is the point of the change, not a regression it introduced.

### What this PR does

The harness reassembled the compiler pipeline itself and got it wrong: it ran `resolve_types` and skipped the semantic rules, while `ironplcc` reaches codegen only when `semantic()` reports nothing. A codegen test could therefore assert bytecode for a program the shipped compiler rejects. That is how the WSTRING literal typing bug (#1550) survived a full end-to-end suite.

The harness now calls `ironplc_project::compile::compile` — the same function the CLI calls — rather than asserting today's version of its rules, so it cannot drift again.

### The 67 fixed

- **~26** bit-string tests assigned an untyped integer literal to `BYTE`/`WORD`/`DWORD`/`LWORD`. ADR-0031 gates that behind `--allow-cross-family-widening`; they now enable it, as a user would.
- **~21** boolean/comparison tests declared a `DINT` target and assigned a `BOOL`, because the assertion only read the opcode.
- **10** `SEL` tests passed `0`/`1` as the selector; `G` is `BOOL` (Table 27).
- The rest were smaller fixture errors of the same kind.

### The 27 remaining — three compiler bugs and two decisions

**Bugs, each confirmed against the shipped binary:**

1. **Functions with `VAR_IN_OUT` cannot be called** (3 tests). The arity check does not count them:
   ```
   error[P4018]: Function call has wrong number of arguments
                 (function=ADD_TEN, expected=0, actual=1)
   ```
   Codegen supports the feature; three end-to-end tests prove it executes correctly. No user can reach it.

2. **Single-element arrays are rejected** (5 tests). The subrange strict-less-than rule is applied to array index ranges:
   ```
   error[P2002]: Subrange declaration minimum value is not less than the maximum
                 (minimum=1, maximum=1)
      arr : ARRAY[1..1] OF INT;
   ```
   `ARRAY[1..1]` is legal IEC 61131-3.

3. **`NOT` over `ANY_INT` silently returns the wrong value** (found while fixing fixtures; those 3 tests still pass and are left alone):
   ```
   x : DINT := 10;  y : DINT;  y := NOT x;   →  y = 0
   ```
   The analyzer types the result `dint` while codegen emits `BOOL_NOT`. Should be `-11`, or a type error — `NOT` is defined over `ANY_BIT`. No diagnostic is produced, so this one corrupts data rather than refusing to build.

**Decisions that are not mine to make:**

4. **Constant overflow** (10 tests). `end_to_end_const_trunc` asserts fold-then-truncate wrapping (`x : SINT := 200` → wraps); the analyzer now rejects it outright as P2026. Either the tests are stale or the rule is too strict.
5. **`SHR(ABS(a), 1)`** (2 tests). `SHR` takes `ANY_BIT`, `ABS` returns `ANY_NUM`, so this is not valid IEC as written. `--allow-cross-family-widening` correctly does not rescue it.

Plus 7 still to triage, including a `VAR_GLOBAL` that is invisible inside a function (`P4007 variable=devices`) which looks like a fourth bug.

### Why it is not fixed here

Fixing 1–3 inside this PR turns a test-infrastructure change into a compiler-behaviour change across three unrelated subsystems. The suggestion is three separate PRs for those bugs, this one rebased on them, and an issue for the P2026 decision. Awaiting a call on the split before proceeding — no fixture edit can close the remaining 27.

---

garretfick:
Correction to my comment above: I said "`Compiler Quality Checks` is fixed as of 33ca227". That is wrong.

That job runs `just ci`, which is `compile coverage lint dupes` — and `coverage` runs the whole test suite under `cargo llvm-cov`. So it is red for the **same** reason as `Compiler Tests`: the 27 failing codegen tests. It was never a lint failure.

The fmt/clippy fixes in 33ca227 were still needed (my scripted fixture edits left imports unsorted), but they were not what made that job red. Verified on the current head:

- `cargo clippy --all-targets` — 0 errors, 0 warnings
- `cargo fmt --all -- --check` — clean
- `cargo dupes check ...` — passes (7.2% exact, 2.2% near)

So `compile`, `lint` and `dupes` all pass; only `coverage` fails, on those 27 tests. Both red checks collapse to the single blocker described above, and both clear together once the three compiler bugs are fixed.

---

garretfick:
Correction, and main is now merged in (252b542, resolving a `codegen/Cargo.toml` conflict between the 0.242.0 version bump and the added `ironplc-project` dev-dependency).

I said in the status comment above that #1665 removes 3 of the 27. **That is wrong. It removes zero, and the count is still 27 with main merged.**

The three `NOT`-over-`DINT` fixtures were never among the failures. They pass on this branch today, because the current analyzer accepts `NOT` over a `DINT`. I saw "three tests involving that bug" and folded them into the blocker list without checking they were actually red.

What #1665 does for this PR is prevent three *new* failures: once it lands the analyzer rejects those programs, and #1665 updates the same three fixtures to `BOOL` in the same change, so the two cancel out.

| Fix | Effect on the 27 |
| --- | --- |
| #1665 — bit-string operators over `ANY_INT` | **0** |
| `VAR_IN_OUT` arity (P4018) | **−3** |
| single-element arrays `ARRAY[n..n]` (P2002) | **−5** |

With all three landed this PR sits at **19**, not 16, and every one of those 19 needs a decision rather than a bug fix — the P2026 constant-overflow semantics (10), `SHR(ABS(..))` (2), and 7 others including what looks like a fourth bug (`P4007 variable=devices`, a `VAR_GLOBAL` invisible inside a function).

#1665 is still worth landing on its own merits — it fixes silent data corruption — but it is not evidence that this PR's count is coming down. `VAR_IN_OUT` will be the first fix that actually moves it.

---

PR #1665, opened 2026-09-07
Hold the bit-string operators to ANY_BIT

garretfick:
`AND`, `OR`, `XOR` and `NOT` in their **operator** spelling accepted `ANY_INT` operands and compiled to the *logical* opcodes, silently returning the wrong value with no diagnostic.

Measured on `main` with `x : DINT := 10`, right operand `3`:

| | `x AND 3` | `x OR 3` | `x XOR 3` | `NOT x` |
| --- | --- | --- | --- | --- |
| **want** | 2 | 11 | 9 | -11 |
| **got** | **1** | **1** | **0** | **0** |

Each result is the truthiness of the operands, not a bit pattern. `BYTE` operands were correct throughout (2, 11, 9), so the defect was confined to operands outside `ANY_BIT`.

## Cause

Codegen selects the opcode by width and signedness, and `BOOL` and the signed integers are both `(W32, Signed)` — indistinguishable at that point, so an integer operand fell through to the boolean arm:

```rust
// codegen/src/compile_expr.rs
(OpWidth::W32, Signedness::Unsigned) => emitter.emit_bit_not_32(),  // BYTE/WORD/DWORD
(OpWidth::W64, Signedness::Unsigned) => emitter.emit_bit_not_64(),  // LWORD
_ => emitter.emit_bool_not(),                                        // BOOL *and* DINT
```

IEC 61131-3 defines all four over `ANY_BIT` (Table 28), so the operand is what is wrong, not the lowering.

## Fix

The operator-form table already carried the rows — `AND`, `OR`, `XOR` and `NOT`, each at `ANY_BIT`, with `FormOf::Not` and `Arity::Unary` already present. `rule_operator_operand_type_check` already knew how to check an operand against a row via `are_types_compatible`. It simply never asked for anything but `MOD`: it implemented only `visit_binary_expr`.

It now asks at the compare and unary nodes too, reporting P4049. **Codegen is unchanged** — the boolean arm is now reachable only for `BOOL`, which is what it is correct for.

This is the mirror of #1567, which fixed the *functional* spelling of the same family wrongly rejecting `WORD`, by declaring it `ANY_BIT`. Both spellings now agree in both directions.

## Left alone deliberately

Each matches the rule's existing conservative posture, and is called out in the module doc:

- The relational rows (`GT`, `EQ`, …) are `ANY_ELEMENTARY`; holding them to that is a separate decision, like `ADD`/`SUB` on `TIME` (#1621).
- `AND_THEN` / `OR_ELSE` have no table row — a CODESYS/TwinCAT short-circuit extension, not IEC operators.
- Operands whose type the predicate cannot judge (subrange, enumeration, structure) are skipped, as they are for `MOD` today.

`a > 0 AND a < 10` and `NOT (a > 0)` are unaffected: comparison results are `BOOL`. Both have explicit tests.

## Breaking change

`x AND 3` where `x` is a `DINT` no longer compiles. Today it compiles and returns the wrong number, so an error is strictly better than what users get now. `P4049.rst` gains a bit-string section showing the diagnostic and the fix.

## Note on the three fixture changes

`end_to_end_bool.rs` and `compile_bool.rs` had three fixtures asserting the miscompile — `NOT` over a `DINT` expecting `BOOL_NOT`. They passed before this change **and would have kept passing**, because codegen tests do not run the semantic rules. That gap is what #1663 closes; these fixtures are updated here so that PR rebases cleanly.

## Validation

`cd compiler && just` (compile incl. the `no_std` gate, coverage ≥85%, clippy, fmt, dupes) and `cd specs && just` both pass. 12 new rule tests cover accept and reject per operator, plus the comparison-operand and relational cases.

Follows the plan → prefactor → change workflow; the prefactor (sharing one operand check across the three node kinds) is its own commit and leaves the 1680-test analyzer suite unchanged. The plan is removed in the final commit.

---


compiler/analyzer/src/rule_operator_operand_type_check.rs
```diff
@@ -101,9 +167,22 @@ impl DiagnosticVisitor for RuleOperatorOperandTypeCheck<'_> {
...
+        let expected = form.operand_type();
+        for operand in operands {
+            self.check_operand(op, &expected, operand);
+        }
+    }
+
     /// Reports P4049 when `operand`'s resolved type is one the predicate can
     /// judge and it is not acceptable where `expected` is required.
-    fn check_operand(&mut self, op: &Operator, expected: &TypeName, operand: &Expr) {
+    fn check_operand(&mut self, op: &str, expected: &TypeName, operand: &Expr) {
```
garretfick: Why are we converting to a string here? That turns a compile time error into a runtime error.

garretfick (commented):
see comment

compiler/analyzer/src/rule_operator_operand_type_check.rs
garretfick: You're right — fixed in ff879e3.

That was the prefactor's doing. Unifying the three visitors meant one signature had to accept `Operator`, `CompareOp` and `UnaryOp`, and I reached for the widest type they all convert to. `&str` does accept all three, but it accepts every other string too: `check_operands` would have taken a label from the wrong node, or an empty one, and nothing would have caught it until someone read a diagnostic naming the wrong operator.

There was a second cost I hadn't clocked: `&node.op.to_string()` at the call site allocates a `String` for **every** binary, compare and unary node the rule visits, even though the label is only used when a diagnostic is built. On a clean compile that's an allocation per operator node, all discarded.

`&impl fmt::Display` fixes both. The callers pass `&node.op` — the real enum, so the compiler checks it again — and the render moves inside `check_operand`, onto the path that actually reports:

```rust
fn check_operands(&mut self, op: &impl fmt::Display, form: &OperatorFunctionForm, operands: &[&Expr])
```

No behaviour change: `d1 AND d2` on `DINT` still reports `operator=AND`, the analyzer suite passes (1692 tests), and the workspace tests, clippy and fmt are clean.

---

compiler/analyzer/src/rule_operator_operand_type_check.rs
```diff
@@ -101,9 +168,31 @@ impl DiagnosticVisitor for RuleOperatorOperandTypeCheck<'_> {
...
+        let expected = form.operand_type();
+        for operand in operands {
+            self.check_operand(op, &expected, operand);
+        }
+    }
+
     /// Reports P4049 when `operand`'s resolved type is one the predicate can
     /// judge and it is not acceptable where `expected` is required.
-    fn check_operand(&mut self, op: &Operator, expected: &TypeName, operand: &Expr) {
+    fn check_operand(&mut self, op: &impl fmt::Display, expected: &TypeName, operand: &Expr) {
```
garretfick: Why not just pass `Operator` here? This is being passed to create label context. Just convert to the string in the function.

garretfick: Following that through, the operator does not need passing at all — dropped in fa0ac5a.

`Operator` alone can't be the parameter, because three enums arrive here: `Operator` from `visit_binary_expr`, `CompareOp` from `visit_compare_expr`, `UnaryOp` from `visit_unary_expr`. That's what pushed it to a generic in the first place.

But `check_operands` already takes the row, and `OperatorFunctionForm` carries `name` next to `operand_type()`. For every operator this rule checks, the row name and the source spelling are the same string:

| operator | source | row |
| --- | --- | --- |
| `Operator::Mod` | `MOD` | `MOD` |
| `CompareOp::And` / `Or` / `Xor` | `AND` / `OR` / `XOR` | same |
| `UnaryOp::Not` | `NOT` | `NOT` |

So the parameter was carrying what the function already had:

```rust
fn check_operands(&mut self, form: &OperatorFunctionForm, operands: &[&Expr])
fn check_operand(&mut self, form: &OperatorFunctionForm, expected: &TypeName, operand: &Expr)
```

That removes the generic, the `fmt` import and a parameter from two signatures — and closes your earlier point at its source rather than by typing it more tightly: no caller supplies a label now, so the operator name and the expected type in a diagnostic come from one row and cannot disagree. Nothing previously stopped `CompareOp::And` being passed alongside the `MOD` row.

All five diagnostics verified byte-identical against the built compiler (`operator=AND`, `OR`, `XOR`, `NOT`, `MOD`); analyzer suite 1692 tests, workspace tests, clippy and fmt all clean.

One nuance, since it's the only behavioural difference and it's latent: this ties the label to the function-form name rather than the source spelling. Identical for everything checked today, but if the arithmetic rows are ever enabled (#1621), `a + b` would report `operator=ADD` rather than `operator=+`. I'd argue `ADD` is the better label — it matches how P4026 names the same family — but say the word if you'd rather it echoed what the user typed.

---

PR #1671, opened 2026-09-07
Reject program variables that hide global variables (P4050)

garretfick:
## Summary

Implements a new semantic rule that rejects program variables whose names match global variables in the merged library. This addresses a correctness issue where a program's local variable could silently overwrite a global variable's slot in the flat name table used by codegen, causing the global to become inaccessible.

## Changes

- **New rule `rule_program_var_hides_global`** (`P4050`): Rejects `VAR`, `VAR_INPUT`, `VAR_OUTPUT` and other program variable declarations that reuse a global's name (case-insensitive). Programs must use `VAR_EXTERNAL` to reference globals. The rule applies to globals declared in `CONFIGURATION`, `RESOURCE`, top-level, or compatibility libraries.

- **New intermediate `collect_global_var_decls`**: Extracts all `VAR_GLOBAL` declarations from a library in walk order. Reused by both the new rule and the existing `rule_var_decl_global_const_requires_external_const` rule (refactored to use it).

- **Documentation**: Added `P4050.rst` explaining the error, providing examples of both the problem and fixes (using `VAR_EXTERNAL` or renaming the local variable).

- **Updated explanation**: Rewrote `object-orientation.rst` to clarify that derived function blocks cannot redeclare inherited variables (not hiding, but rejection via `P4044`). Clarified that hiding does exist for method locals hiding instance variables via `THIS^`.

- **Updated scope documentation**: Added note that programs cannot reuse global names and reference `P4050`.

- **Updated `P4044` documentation**: Clarified that `EXTENDS` is Edition 3 syntax (not a vendor extension) and that the rule applies unconditionally.

- **Test updates**: Fixed test comment in `rule_assignment_aggregate_type_compat.rs` to reflect that only function blocks (not programs) can hide globals.

## Implementation Details

- The rule is asymmetric by design: programs reject name reuse, while functions, function blocks, and methods may still hide outer names. This follows the back end's flat name table limitation and is documented in ADR-0051.

- Case-insensitive comparison uses `Id` type, matching IEC 61131-3 identifier semantics.

- Diagnostics include both the program variable and the global declaration, making the fix obvious (`VAR_EXTERNAL` or rename).

- Comprehensive test coverage including configuration globals, resource globals, top-level globals, and case-insensitive matching.

## Related

Implements decision from ADR-0051 (included in this PR) on rejecting variable redeclarations where the compiler has no executable path.

https://claude.ai/code/session_019bizNZx7qVWoEEY6cYMHHA


docs/reference/compiler/problems/P4050.rst
```diff
@@ -0,0 +1,70 @@
...
+The comparison is case-insensitive, like all IEC 61131-3 identifiers, and
+applies to global variables declared in a ``CONFIGURATION``, in a
+``RESOURCE``, at the top level of a file, or in an activated library.
+
+Example
+-------
+
+The following code will generate error P4050:
+
+.. code-block::
```
garretfick, line 22: This should be a playground example. Same with other examples here.

docs/explanation/object-orientation.rst
```diff
@@ -60,15 +60,13 @@ provides single inheritance for function block types. An instance of a derived
...
-*shadows*) the inherited one: within the derived type's own body the name
-refers to the member declared there, not to the one it inherited. The
-inherited member is not removed — it still exists on every instance and is
-reached through the base type.
+A derived type inherits every variable of its base type, and it cannot
+declare a variable of its own with one of those names. The following is
+rejected with :doc:`P4044 </reference/compiler/problems/P4044>`, whatever
+the type of the second declaration:
 
 .. code-block::
```
garretfick, line 71: This should be a playground example

PR #1678, opened 2026-09-08
STRING_TO_UDINT selects its conversion by policy; document the policies

### LINKED ISSUE #1592: STRING_TO_UDINT silently returns 0 for the entire upper half of the UDINT range
Found triaging plans for #1464. **Ran; reproduces with two in-run controls.** Silent wrong answer on a conversion `docs/` marks Supported.

```iecst
PROGRAM main
VAR
  s_ok   : STRING[32] := '123';
  s_low  : STRING[32] := '2147483647';
  s_high : STRING[32] := '2147483648';
  s_max  : STRING[32] := '4294967295';
  ctrl : UDINT; low : UDINT; high : UDINT; maxv : UDINT;
END_VAR
  ctrl := STRING_TO_UDINT(s_ok);
  low  := STRING_TO_UDINT(s_low);
  high := STRING_TO_UDINT(s_high);
  maxv := STRING_TO_UDINT(s_max);
END_PROGRAM
```

```
$ ironplcc compile -o s2u.iplc s2u.st && ironplcvm run --scans 1 --dump-vars - s2u.iplc
ctrl: 123
low: 2147483647
high: 0
maxv: 0
```

| input | got | expected |
|---|---|---|
| `'123'` | 123 | 123 ✓ (control) |
| `'2147483647'` = `i32::MAX` | 2147483647 | ✓ (control — boundary) |
| `'2147483648'` | **0** | 2147483648 |
| `'4294967295'` = `UDINT` max | **0** | 4294967295 |

No diagnostic, no trap. Everything above `i32::MAX` — half the type's range — silently becomes zero.

## Cause

`compiler/codegen/src/compile_call.rs:1300-1307` dispatches `StringToNum` on `target.op_width` **only**, ignoring signedness. Every `STRING_TO_<unsigned>` therefore lands on `CONV_STR_TO_I32`, whose VM handler (`compiler/vm/src/vm.rs:1438-1440`) does `.parse::<i32>().ok().unwrap_or(0)` — the parse fails for anything above `i32::MAX` and the `unwrap_or(0)` swallows it.

The asymmetry is visible fifteen lines up: the `NumToString` arm at `compile_call.rs:1285-1286` *does* branch on signedness (`CONV_I32_TO_STR` vs `CONV_U32_TO_STR`). The outbound direction got a U32 variant; the inbound one never did. There is no `CONV_STR_TO_U32` builtin at all.

Round-trip control: `UDINT_TO_STRING(16#FFFFFFFF)` correctly produces a 10-character string; feeding that same string straight back through `STRING_TO_UDINT` yields `0`.

`STRING_TO_USINT` and `STRING_TO_UINT` are unaffected — their ranges fit inside `i32`.

## The design docs already warned about exactly this

`specs/design/complete-numeric-stdlib.md` records the hazard for the MIN/MAX/LIMIT family:

> a UDINT at or above 2^31 reinterprets as a negative i32 … That is what the U32 and U64 variants below exist to prevent

The string-conversion path repeats the mistake those variants were introduced to avoid.

## Docs

`docs/reference/standard-library/functions/type-conversions.rst:431` lists `STRING_TO_UDINT` — "String to 32-bit unsigned" — as **Supported**.

Separately, that page says only "The string must contain a valid numeric literal for the target type" and never states what happens when it does not. The parse-failure behaviour (return 0, no diagnostic, no trap) is a deliberate decision recorded at `compiler/container/src/builtin.rs:352,360` but never surfaced to users — which is what makes this bug invisible: the failure mode for a *valid* literal is identical to the documented-nowhere failure mode for an invalid one.

## Fix

Add a `CONV_STR_TO_U32` builtin and dispatch on signedness in the `StringToNum` arm, mirroring `NumToString`. Worth checking `STRING_TO_ULINT` at the same time.

garretfick:
## Summary

Slice 5 of 5 of the `STRING_TO_UDINT` steel thread for ADR-0049 (compile-time behavior policies). Merged so far: #1657 (prefactor), #1656 (policy enums and func_id block), #1669 (policies on `CompilerOptions` and their surfaces), #1677 (VM scanner and V4006 trap).

This slice closes the thread and **fixes #1592**: `STRING_TO_UDINT` now converts the whole UDINT range, and a non-convertible string traps or yields zero as the selected policy says.

- **Codegen.** `CodegenOptions` carries the two policies (derived from `CompilerOptions` by the `From` impl from #1657) and threads them to `CompileContext`. The `StringToNum` arm routes the 32-bit unsigned target to `builtin::str_to_num::func_id`, so the func_id names the target and both selected alternatives. Signed and sub-32-bit targets keep the single-encoding `CONV_STR_TO_I32` until their own PRs.
- **Tests.** An end-to-end matrix runs one source under every alternative, including the #1592 inputs (`'2147483648'`, `'4294967295'`) and the `UDINT_TO_STRING` round trip; the strict default traps V4006 with `string '12abc' is not convertible to UDINT`; the `codesys` / `twincat` presets give 12 for `'12abc'` and 0 for out of range; `rusty` gives 0 without a trap. A codegen conformance test shows the same source under two selections differs only in the func_id operand (ADR-0049 confirmation 1). A vm-cli test checks exit code 1 and the V4006 message from a compiled program.
- **Docs.** The type-conversions page gains a *String to Numeric Policies* section: each alternative, a results table per policy for `STRING_TO_UDINT`, the dialects' selections, and the note that for a literal that does not fit, CODESYS documents an undefined result and IronPLC's zero under those presets is its own choice. The CLI page documents the two `--policy-*` flags; the dialect explanation page gains a *Selecting Behavior Policies* section and a **Selects** line per dialect. The page says plainly that only `STRING_TO_UDINT` honors the policies today.
- **Records.** ADR-0049 flips to `accepted` with a postscript on what landed and the two details that differ from its table (three non-numeric alternatives; `rusty` selects `zero`). The design doc gains the codegen-owned requirements and an *Adding a policy* checklist.

## The series

1. #1657: prefactor (merged).
2. #1656: policy enums and the func_id block (merged).
3. #1669: policies on `CompilerOptions` and their surfaces (merged).
4. #1677: the VM scanner and the V4006 trap (merged).
5. **This PR**: `STRING_TO_UDINT` selects its builtin by policy; user docs; ADR-0049 to accepted.

The plan for this slice is the first commit and is removed by the last, per the development standards.

## Not in this series

The remaining `STRING_TO_*` targets (the rest of the tracking issue's PR list), the `flag` failure alternative, the non-trapping validator library function, and per-file / project-discovery policy selection.

## Test plan

- `cd compiler && just` (compile, coverage, lint, dupes) passes on this branch.
- Docs build with `-W` (fresh environment), `cd docs && just duplicates`, and `cd specs && just` pass.

https://claude.ai/code/session_01SPWFrGBWGGvdDDwzR2Ry88

---
_Generated by [Claude Code](https://claude.ai/code/session_01SPWFrGBWGGvdDDwzR2Ry88)_


compiler/codegen/tests/it/end_to_end_string_to_udint.rs
```diff
@@ -0,0 +1,244 @@
...
+
+#[rstest]
+#[case::codesys(Dialect::Codesys)]
+#[case::twincat(Dialect::TwinCat)]
+fn string_to_udint_when_codesys_family_dialect_then_prefix_and_zero(#[case] dialect: Dialect) {
+    // TwinCAT (observed) stops at the first invalid character, and documents
+    // 0 for a string that is not valid in the target type.
+    let options = CompilerOptions::from_dialect(dialect);
+    let (_c, bufs) = parse_and_run(&program("12abc"), &options);
+    assert_eq!(bufs.vars[1].as_i32() as u32, 12);
```
garretfick, line 170: These indices are problematic because it is difficult to know if the value is zero because of initial value or if because the value was converted correctly. There are two parts - the index and the initial value.

Issue #1683, opened 2026-09-11
Tracking: move the remaining STRING_TO_* targets onto the behavior-policy func_id block


garretfick:
[ADR-0049](https://github.com/ironplc/ironplc/blob/main/specs/adrs/0049-behavior-policies-selected-at-compile-time.md) landed the behavior-policy machinery and its first application, `STRING_TO_UDINT`, in #1656, #1657, #1669, #1677 and #1678 (fixing #1592). The other `STRING_TO_<numeric>` functions still compile to the single-encoding `CONV_STR_TO_I32` / `CONV_STR_TO_F32` builtins: they ignore the two `--policy-string-to-num-*` selections, return 0 on failure whatever the policy says, and the sub-32-bit targets **wrap** on out-of-range input (`STRING_TO_SINT('300')` gives 44) because the conversion is followed by a `TRUNC_*`.

This issue tracks moving every remaining target onto the `0x0480..=0x04FF` func_id block described in [behavior-policies.md](https://github.com/ironplc/ironplc/blob/main/specs/design/behavior-policies.md), one PR per width class. Each PR declares the target's six func_ids at the positions the block reserves, implements the six handlers by instantiating the existing scanner with the target's bounds, pins them in the wire-format test, range-checks inside the conversion so nothing wraps, adds end-to-end tests at min, max, one past each bound and the shared invalid inputs under all six policy combinations, extends the benchmark suite, and updates the type-conversions doc page and the func_id table in the instruction-set spec.

## PRs

- [x] **PR 2** (#1686, merged): `SINT`, `INT`, `DINT`, `USINT`, `UINT`, `BYTE`, `WORD`, `DWORD`. Bit-string types are aliases of the unsigned ids of the same width; the analyzer now registers `STRING_TO_BYTE` / `STRING_TO_WORD` / `STRING_TO_DWORD`. The `TRUNC_*` after the conversion is gone, and the target positions for the rest of the series are fixed in the design doc (U64 = 6, I64 = 7, F32 = 8, F64 = 9).
- [x] **PR 3** (#1696, merged): `LINT`, `ULINT`, `LWORD` at positions 6 and 7. The integer scanner needed only its two bounds entries; `str_to_num::convert` now returns a slot of the target's width, and the analyzer registers the three functions.
[...]

PR #1686, opened 2026-09-11
STRING_TO_* honors the behavior policies for every integer target up to 32 bits

garretfick:
## Summary

PR 2 of the series tracked by #1683: `SINT`, `INT`, `DINT`, `USINT`, `UINT`, `BYTE`, `WORD` and `DWORD` join `STRING_TO_UDINT` on the `STRING_TO_<numeric>` func_id block (ADR-0049). PR 1 (#1656, #1657, #1669, #1677, #1678) built the mechanism; this PR adds targets to it and redesigns nothing.

- **Encoding.** Five new `Target`s at the positions the block fixes for the whole series: even positions unsigned, the odd position after each its signed counterpart, in width order 32, 8, 16, 64 (32 first because U32 landed at 0), then the reals. `I32` = 1, `U8` = 2, `I8` = 3, `U16` = 4, `I16` = 5; positions 6–9 are reserved for PRs 3 and 4. Thirty `declare_builtins!` rows (0x0488–0x04AD), pinned in the wire-format test. Bit-string types alias the unsigned target of their width, so `STRING_TO_BYTE` and `STRING_TO_USINT` compile to the same func_id.
- **VM.** One integer scanner, parameterised by the target's bounds (prefactor commit). The magnitude accumulates in 64 bits with a branchless overflow flag, and the range check is two unsigned compares against the target's magnitude on each side of zero, so the 64-bit targets of PR 3 fit without wider arithmetic. Out of range is a failure at every width; `STRING_TO_SINT('300')` traps V4006 or yields zero, and never wraps to 44.
- **Codegen.** Every `W32` integer target maps to a block target; the `TRUNC_*` after the conversion is gone. `CONV_STR_TO_I32` is no longer emitted (its VM handler stays as a wire-format commitment; the spec says so).
- **Analyzer.** `STRING_TO_BYTE`, `STRING_TO_WORD` and `STRING_TO_DWORD` are registered.
- **Trap.** V4006 names the target the func_id encodes, so a bit-string function reports the unsigned integer type of its width (documented on the type-conversions and V4006 pages).
- **Tests.** End-to-end for all eight functions at min, max, one past each bound and the shared invalid inputs from PR 1 under all six policy combinations; instruction-level range tests per target in the VM; the conformance tests for `REQ-BP-codegen-001/002` and `REQ-BP-container-002` generalised to every target.
- **Docs and specs.** The type-conversions page gains the per-function ranges, a `STRING_TO_SINT` results table and the alias note; the design doc and instruction-set spec record the target positions and an *Adding a target* checklist; ADR-0049 gets a dated postscript.

## Benchmarks

New `st_string_to_num` group in `ironplc-benchmarks`: one hundred conversions of a valid literal per scan, one case per width, strict policies. Same machine, `main` versus this branch (the `main` numbers for SINT/INT/DINT go through the old `CONV_STR_TO_I32` path, `UDINT` through the PR 1 scanner):

| Case | literal | `main` | this PR |
|---|---|---|---|
| SINT | `'100'` | 4.54 µs | 4.12 µs |
| INT | `'30000'` | 4.69 µs | 4.41 µs |
| DINT | `'2000000000'` | 4.62 µs | 4.69 µs |
| UDINT | `'4000000000'` | 4.57 µs | 4.62 µs |

The first draft of the shared scanner cost UDINT about 10% on ten-digit literals (128-bit value compare and `Option`-chained checked arithmetic); the branchless accumulator and magnitude-based bounds brought it back to within noise.

**Note:** the benchmark binary currently panics at startup on `main` because `bench_diverse_opcodes` no longer compiles under the analyzer (bitwise operators on `DINT` and a `DWORD := ANY_INT` assignment, rejected since #1665). The numbers above were taken with that group patched out locally; the patch is not in this PR, and the fix is queued as a separate task.

## Test plan

- `cd compiler && just` (compile, coverage, lint, dupes) passes.
- `cd specs && just` passes; the plan is deleted in the last commit.
- Docs build with `-W` and `cd docs && just duplicates` pass.

Part of #1683.

https://claude.ai/code/session_01H2o7aPy8ZCL4eCe5vCc2vD

---
_Generated by [Claude Code](https://claude.ai/code/session_01H2o7aPy8ZCL4eCe5vCc2vD)_


compiler/codegen/src/compile_call.rs
```diff
@@ -1222,33 +1222,75 @@ pub(crate) fn compile_string_conversion(
...
+fn str_to_num_target(target: VarTypeInfo) -> Option<opcode::builtin::str_to_num::Target> {
+    use opcode::builtin::str_to_num::Target;
+    match (target.op_width, target.signedness, target.storage_bits) {
+        (OpWidth::W32, Signedness::Unsigned, 32) => Some(Target::U32),
+        (OpWidth::W32, Signedness::Signed, 32) => Some(Target::I32),
+        (OpWidth::W32, Signedness::Unsigned, 8) => Some(Target::U8),
+        (OpWidth::W32, Signedness::Signed, 8) => Some(Target::I8),
+        (OpWidth::W32, Signedness::Unsigned, 16) => Some(Target::U16),
+        (OpWidth::W32, Signedness::Signed, 16) => Some(Target::I16),
+        _ => None,
```
garretfick, line 1264: can this be a diagnostic error so that we don't do nothing? alternative the above code that checks needs to give an error if this returns nothing. Generally, we try to avoid `_` catch all.

docs/reference/standard-library/functions/type-conversions.rst
```diff
@@ -415,25 +415,34 @@ selects it at compile time with two behavior policies; see
      - Support
    * - ``STRING_TO_SINT``
      - String to 8-bit signed
-     - Supported
+     - Supported, honors the policies
```
garretfick: Why do we need to call out "honors the policy" or even "supported"? Let's just create an issue for the item that's not supported (LREAL) to simplify what's in the docs.

compiler/codegen/src/compile_call.rs
garretfick: Done in 1d5e1a1-ish (the "Review:" commit). The helper and its `_ => None` are gone; the arm now matches every `(width, signedness, value width)` explicitly. The six 32-bit integer shapes select the block func_id, `F32` selects `CONV_STR_TO_F32`, the 64-bit integers and `F64` return the todo diagnostic, and any other 32-bit target returns an internal error (the analyzer offers no such `STRING_TO_*`, so reaching it is a compiler bug). Nothing falls through and does nothing.

---

docs/reference/standard-library/functions/type-conversions.rst
garretfick: Dropped the support column: the String to Numeric table now lists only the functions that exist (name and description), and the `STRING_TO_LINT` / `STRING_TO_LREAL` "not yet supported" rows are gone. Those targets are already tracked as PRs 3 and 4 of #1683 (LINT, ULINT, LWORD; REAL, LREAL), so I did not open a second issue. The one sentence noting that `STRING_TO_REAL` returns zero rather than honoring the policies stays in the policies section, since that is a behavioral difference a reader hits.

---

Issue #1685, opened 2026-09-11
ADR-0012's accept-as-is promise has no harness, and dialect is not detected per file


garretfick:
Found while auditing ADR-0012 for #1586. Two gaps, both recorded in that ADR's new Implementation Status section; this issue tracks closing them.

ADR-0012 states the principle that **IronPLC must be able to parse any file that the vendor's own toolchain accepts, producing zero parse errors on syntactically valid vendor code.** It is a load-bearing decision — ADR-0036 depends on it for the rule that IronPLC defines no dialect of its own — but nothing measures it, and one of its two mechanisms was built differently.

## 1. Nothing measures "parse-clean on real projects"

Confirmation item 1 asks, for each vendor dialect: take 3+ open-source projects from that ecosystem and confirm zero parse errors on all files. There is no such test. The nearest thing is the single OSCAT case in `compiler/parser/src/tests/corpus.rs`, which is one vendor-neutral `.st` file.

Without a harness, "we parse what TwinCAT parses" is an aspiration. It is also the claim most likely to rot silently: every parser change can regress it, and nothing would notice.

Worth deciding as part of this: whether the corpus is vendored into the repo (licensing, size) or fetched in a dedicated CI job, and whether a regression fails the build or reports a count that must not increase. A ratchet is probably the realistic shape — real vendor projects will not be 100% clean on day one, and a hard gate that starts red gets disabled.

## 2. Dialect is not detected from the file extension

ADR-0012's "Dialect detection strategy" section says the parser determines the dialect from the file extension: `.scl` → Siemens, `.TcPOU` → Beckhoff, `.st`/`.iec` → standard.

What exists is a two-part split:

- The extension selects the **parser** — `FileType::from_path` in `compiler/sources/src/file_type.rs` routes `.TcPOU`/`.TcGVL`/`.TcDUT`/`.TcIO` to the XML wrapper and `.st`/`.iec` to plain ST.
- The **dialect** — which `--allow-*` extensions are enabled — comes from `--dialect` or individual flags chosen by the user.

So a `.TcPOU` file compiled without `--dialect twincat` parses as XML and then rejects the TwinCAT syntax inside it. For a user pointing the compiler at a TwinCAT project, that is the accept-as-is promise failing at the first file, and the fix is a flag they have to know to pass.

[...]

PR #1694, opened 2026-09-12
ADR-0012: Clarify dialect selection is explicit, not inferred

garretfick:
## Summary

Amend ADR-0012 to correct the record on how vendor dialects are selected. The original text described a mechanism where the dialect would be automatically detected from file extensions (`.scl` → Siemens, `.TcPOU` → Beckhoff, etc.). This mechanism was never implemented. The actual design that exists separates two concerns: the file *format* is detected from the extension, and the *dialect* is selected explicitly by the user via `--dialect` or `--allow-*` flags.

## Changes

- **Rename and rewrite the mechanism section** from "Dialect detection strategy" to "File format from the extension, dialect from the command line" to accurately describe the implemented design
- **Add detailed rationale** for why dialect selection must be explicit rather than inferred:
  - Preserves the ability to check vendor code against strict Edition 2 (a check no vendor toolchain offers)
  - Makes accepted syntax visible in the invocation, not reconstructed from a table
  - Avoids conflating file storage format with syntax validation rules
- **Update "No mixing" section** to reflect that a single compilation uses one dialect for all files, rather than per-file dialect selection
- **Revise consequences** to acknowledge the trade-off: users must pass `--dialect` on first use, but this is mitigated by help text on rejections and documentation leading with the flag
- **Clarify implementation status**: The dialect-from-extension mechanism was never built; what exists is the format/dialect split now documented. This closes one of two gaps in issue #1685
- **Add amendment note** at the end explaining the correction and why the decision outcome (accept-as-is, no mixing) remains unchanged

## Implementation Details

This is a documentation-only change to ADR-0012. No code changes are required; the amendment corrects the ADR to match the actual implementation that has been in place. The change is marked as an amendment per the ADR amendment rules in `specs/steering/development-standards.md`, preserving the original text and adding the correction as a dated amendment note.

https://claude.ai/code/session_01N6sRVf8NKy4wUf3wzyxFMZ

### DIFF
--- a/specs/adrs/0012-accept-vendor-dialect-files-as-is.md
+++ b/specs/adrs/0012-accept-vendor-dialect-files-as-is.md
@@ -3,6 +3,9 @@
 status: proposed
 date: 2026-02-27
 amended: 2026-09-11 (Implementation Status added; status unchanged)
+amended: 2026-09-11 (dialect selection recorded as an explicit user choice
+rather than an inference from the file extension; status unchanged — see the
+amendment note under *More Information*)
 
 > **Terminology note (added later):** This ADR predates
 > [`specs/steering/glossary.md`](../steering/glossary.md), which now draws a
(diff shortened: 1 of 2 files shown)


specs/adrs/0012-accept-vendor-dialect-files-as-is.md
```diff
@@ -39,9 +42,16 @@ The principle is: **IronPLC must be able to parse any file that the vendor's own
 
 ### No mixing of vendor dialects within a file
 
-A single file belongs to exactly one vendor dialect. IronPLC does not allow mixing Siemens and Beckhoff constructs (or any other combination of vendor-specific extensions) in the same file. The dialect is determined per-file from the file extension, and only that dialect's extensions are enabled for parsing that file.
+A single file belongs to exactly one vendor dialect. IronPLC does not allow mixing Siemens and Beckhoff constructs (or any other combination of vendor-specific extensions) in the same file. Only the selected dialect's extensions are enabled while parsing it.
 
-A workspace (project) may contain files from different vendors — for example, some `.scl` files and some `.TcPOU` files — and each file is parsed independently with its own dialect. But within any single file, only one vendor's extensions are valid.
+The dialect is selected by the user for a whole compilation (see
```
garretfick: This is not quite right. What we have now is the default. We would eventually allow specifying by file different dialects so that you can combine. But this would be an exception - something we expect most people to not take advantage of.

garretfick: You're right, and it reached further than this paragraph — fixed in b69d929.

Two ways the old text was wrong:

**Forward.** ADR-0049 already depends on per-file dialects: it cites this ADR for "Compatibility is per file" and picks compile-time policy encoding because "selection is per file, which per-file dialects and mixed compositions require." So per-file dialects are load-bearing in an accepted ADR, not an open question — my "whether per-file dialect selection should ever exist is left open" contradicted it.

**Today.** One compilation already carries more than one dialect configuration: bundled compatibility library bodies are parsed with `CompilerOptions::default()` (strict Ed 2) in `sources/src/libraries`, whatever the user declares.

What changed:

- Restored the workspace paragraph my first commit deleted — files from different vendors, each parsed against its own dialect. It was right about granularity; only the mechanism ("determined per-file from the file extension") was wrong.
- New *Granularity: one declaration per run, by default* subsection. Per-run is the default, per-file is the expected exception, and it reads as an exception: most projects target one platform and will never reach for it. It also records that finer granularity never changes *who* decides — a per-file dialect is declared too, never inferred from the extension — and that it doesn't weaken no-mixing, since each file is still parsed against exactly one dialect.
- Renamed the section to "dialect declared by the user" rather than "from the command line", since a per-file declaration need not be a CLI argument. Same reason for the glossary's canonical use: "per-file dialect detection" → "selection".
- Implementation Status now lists per-file declaration for user source as not built, and explicitly as an expected extension rather than a gap holding the status at `proposed`.

---

PR #1714, opened 2026-09-12
STRING_TO_REAL and STRING_TO_LREAL honor the behavior policies

garretfick:
## Summary

PR 4 of the series tracked by #1683, the last: `REAL` and `LREAL` join the `STRING_TO_<numeric>` func_id block (ADR-0049) at the positions PR 2 reserved, which completes the block: every `STRING_TO_<numeric>` function is now on it. The policies, presets and encoding are as before; what this PR decides is the real literal grammar and two edge cases the integer targets do not have, recorded in the design doc as `REQ-BP-vm-008` through `010` with conformance tests.

- **Grammar** (`REQ-BP-vm-008`). An optional sign, a mantissa (digits, digits with a decimal point and an optional fraction, or a decimal point with a fraction), and an optional `E`/`e` exponent with an optional sign, with `_` separators between digits. No based literal, typed prefix, or word: `inf`, `infinity` and `nan` are not literals. This is CODESYS's "floating-point number, also in exponential notation" (`'9.876'`, `'1.2E-34'`). The integer form (`'5'`) is accepted because every surveyed implementation converts it; the point-first and point-last forms (`'.5'`, `'1.'`) because the C `strtod` the surveyed runtimes build on accepts them, and because rejecting `'.5'` while `ignore-surrounding` skips the point to convert 5 would be a worse result than 0.5. For the integer targets nothing changes: CODESYS documents that "floating-point numbers are not convertible; the decimal separator is treated and truncated as a following character", which is what `reject` (failure) and `ignore-trailing` (the integer part) already do; the docs now say so.
- **Overflow to infinity is a failure** (`REQ-BP-vm-009`). A magnitude that rounds to infinity at the target's width (`'1e39'` for `REAL`, `'1e309'` for `LREAL`) does not fit the target, exactly as `'300'` does not fit `SINT`: V4006 under `trap`, 0.0 under `zero`. Underflow is not a failure: the value rounds to a subnormal or to zero, which is what the text names at that width.
- **Nothing produces a NaN or an infinity** (`REQ-BP-vm-010`). The words are failures under every alternative, `ignore-surrounding` skips them to a literal behind them (`'nan5'` converts 5), and `zero` yields positive 0.0.
- **Parsed at the target's width.** The literal is measured by the grammar and parsed with `core`'s `FromStr` for `f32` or `f64` directly, so the result is the correctly rounded value of the text at that width rather than a wide parse narrowed (a unit test pins a case where the two differ).
- **Encoding.** `Target::F32 = 8`, `Target::F64 = 9`, twelve rows at 0x04C0–0x04C5 and 0x04C8–0x04CD pinned in the wire-format test.
- **VM.** A separate `str_to_real` scanner behind the same trim/skip helpers and the same failure handling. The prefactor commit split `str_to_num::convert` into a per-target scan and one failure function so the second scanner drops into the scan half. `STRING_TO_REAL` no longer returns 0.0 on any failure whatever the policy says.
- **Codegen and analyzer.** `F32` selects the block func_id instead of `CONV_STR_TO_F32` (the VM keeps that handler as a wire-format commitment, like `CONV_STR_TO_I32`); `F64` selects it instead of the todo diagnostic. `STRING_TO_LREAL` is registered.
- **Tests.** End-to-end for both functions under all six policy combinations: the CODESYS examples, overflow at each width, underflow, the NaN and infinity words, the shared invalid inputs, the presets, and a round trip. The VM conformance tests pin the three requirements through the builtins.
- **Docs.** The type-conversions page gives the real grammar, the overflow and NaN rules, a `STRING_TO_REAL` results table and the integer/decimal-point note; the V4006 page adds the real example; the instruction-set spec and ADR-0049's postscript record the series as complete.

## Benchmarks

`st_string_to_num` gains `REAL` (`'3.14159'`) and `LREAL` (`'3.141592653589793'`) cases. A/B on the same host and day, `main` versus this branch, with the still-broken `bench_diverse_opcodes` group patched out locally (not in this PR); `main` cannot compile the `LREAL` case:

| Case | `main` | this PR |
|---|---|---|
| SINT | 4.07 µs | 4.38 µs |
| INT | 4.29 µs | 4.44 µs |
| DINT | 4.69 µs | 4.66 µs |
| UDINT | 4.53 µs | 4.70 µs |
| LINT | 5.70 µs | 5.83 µs |
| REAL | 4.83 µs (old direct parse) | 6.30 µs |
| LREAL | n/a | 7.95 µs |

The integer cases are within this host's run-to-run noise (its absolute numbers moved 25% between runs today). REAL costs about 15 ns more per conversion than the old `CONV_STR_TO_F32` arm. Measured by bypassing the pieces: roughly 9 ns is the measuring pass the grammar needs (separators, rejecting `inf`/`nan`, finding the prefix under `ignore-trailing`) and the rest is the block's func_id decode and policy dispatch that every target on the block pays, which is why UDINT matches `main`. A first draft was 6.82 µs; dropping a separate separator scan in favour of tracking separators while measuring brought it to 6.30.

## Test plan

- `cd compiler && just` (compile, coverage, lint, dupes) passes.
- `cd specs && just` passes; the plan is deleted in the last commit.
- Docs build with `-W` and `cd docs && just duplicates` pass.

Part of #1683; completes its PR list.

https://claude.ai/code/session_01H2o7aPy8ZCL4eCe5vCc2vD

---
_Generated by [Claude Code](https://claude.ai/code/session_01H2o7aPy8ZCL4eCe5vCc2vD)_


compiler/codegen/src/spec_conformance_behavior_policies.rs
```diff
@@ -163,23 +165,21 @@ END_PROGRAM"
...
+/// The load opcode of the slot width a `STRING_TO_<type_name>` result is
+/// compared at.
+fn load_opcode(type_name: &str) -> u8 {
     match type_name {
-        "LINT" | "ULINT" | "LWORD" => LoadWidth::I64,
-        _ => LoadWidth::I32,
+        "LINT" | "ULINT" | "LWORD" => opcode::LOAD_VAR_I64,
+        "REAL" => opcode::LOAD_VAR_F32,
+        "LREAL" => opcode::LOAD_VAR_F64,
+        _ => opcode::LOAD_VAR_I32,
```
garretfick: Is there some way that we can avoid the general match using `_`. Can we make the key an enumeration so that we know we have completely covered all possible options?

garretfick: Done in 447489f. The functions table already carries the block `Target` for each function, so the lookup is now keyed on that and matches every variant explicitly: the six 32-bit integer targets to `LOAD_VAR_I32`, the two 64-bit ones to `LOAD_VAR_I64`, `F32` and `F64` to their float loads. Adding a target without a width mapping is a compile error rather than a silent fall-through to the 32-bit width.

---

PR #1728, opened 2026-09-13
Enforce the POU call hierarchy: a function cannot hold a function block (P4054)

### LINKED ISSUE #1525: Redeclaring __SYSTEM_UP_TIME silently shadows it and reads zero forever
Found while triaging `system-time-global` (#1464). Reproduced with a control, not inferred.

## Reproduction

```iecst
VAR_GLOBAL
    __SYSTEM_UP_TIME : TIME;
END_VAR
PROGRAM main
VAR
    seen : TIME;
END_VAR
    seen := __SYSTEM_UP_TIME;
END_PROGRAM
```

```
$ ironplcc check --dialect rusty --allow-top-level-var-global uptime.st
$ echo $?
0                                    # no diagnostic

$ ironplcvm run --scans 5 --dump-vars=- uptime.iplc
__SYSTEM_UP_TIME: T#400ms
__SYSTEM_UP_LTIME: LTIME#400ms
seen: T#0ms                          # wrong
```

Control — same program with the redeclaration replaced by an unrelated global:

```
__SYSTEM_UP_TIME: T#400ms
unrelated: 0
seen: T#400ms                        # correct
```

Note the dump shows the VM slot holding the right value in *both* runs. The VM writes index 0 faithfully; the user's duplicate declaration creates a second variable of the same name that shadows it everywhere in their code.

## Cause

Two gaps compound:

1. **The reserved-name guard was never implemented.** No rule under `compiler/analyzer/src/rule_*.rs` mentions `__SYSTEM_UP`; the only analyzer touchpoints are the seeding sites (`stages.rs:115`, `rule_use_declared_symbolic_var.rs:72`, `xform_resolve_expr_types.rs:45`).

2. **`SymbolEnvironment::insert` does not detect duplicates at all** — `compiler/analyzer/src/symbol_environment.rs:175`:

```rust
// Check for duplicate global symbols
if let Some(_existing) = self.global_symbols.get(name) {
    // For now, allow redefinition (this might be needed for forward declarations)
    // TODO: Implement proper duplicate detection
}
```

Codegen then appends user globals after the synthetic pair with no dedup (`compile.rs:226`), and `assign_variables` does a map `insert` (`compile_setup.rs:39`), so the later duplicate overwrites the earlier index mapping. The user's reads bind to a slot at index ≥ 2 that the VM never writes.

## Scope

The `__` prefix is documented as a reserved compiler namespace — `specs/design/compatibility-libraries.md:102`: *"Names with a `__` prefix (e.g. `__SYSTEM_UP_TIME`, `__ISVALIDREF`) mark things that only the compiler can provide."* Nothing enforces it.

The narrow fix is a rule rejecting user declarations of the two reserved uptime names when `--allow-system-uptime-global` is on. The broader one is the `TODO` above — duplicate global detection would catch this and a class of neighbours, though it needs the forward-declaration question answered first.

Worth noting the reserved-name guard was a step in the original plan and is recorded nowhere else, which is why this is filed before that plan is deleted.

garretfick:
First of four, split out of the combined branch so each piece is reviewable. This one stands alone and is rebased on current `main`; the rest follow once it lands.

## Why

`rule_pou_hierarchy.rs` never enforced a call hierarchy. It emitted `P4013` for a repeated POU name, recorded the current POU kind in a field nothing read, and its one call-hierarchy test ended in a commented-out assertion. The duplicate check cannot fire either: the rule runs at `stages.rs:356`, long after `xform_toposort_declarations` at `:196` collapses declarations to one per name, so a repeat never reaches it.

So this passes `ironplcc check` on `main` today:

```
FUNCTION Delayed : BOOL
  VAR
    timer : TON;
  END_VAR
  timer(IN := TRUE, PT := T#1s);
  Delayed := timer.Q;
END_FUNCTION
```

## What changed

The file now holds the rule its name promised. IEC 61131-3 lets a program invoke functions and function blocks, a function block invoke functions and function blocks, and a function invoke only functions (Ed.2 §2.5.1, Ed.3 §6.6.1). The distinction is state: a function has none, and a function block instance is state.

**`P4054`** is reported for a function block instance a `FUNCTION` declares outside `VAR_IN_OUT`, and for every invocation of it and method call on it, so the declaration and each call site are both marked. `VAR_IN_OUT` is exempt because Edition 3 permits it — the instance stays the caller's.

The hierarchy's other direction, a function or function block reaching for a program, cannot be written: a program is not in the type environment, so naming one as a variable type is `P2008` and invoking one is `P4012`, from any POU. Nothing covered that, so `stages.rs` now pins it for a function, a function block and a program, along with `PROGRAM ... WITH` in a resource staying the one way to instantiate a program.

**This leaves `P4013` emitted nowhere.** It was already unreachable, so no diagnostic is lost in practice, but its docs page now describes something nothing produces. The next PR in the stack moves duplicate detection into the environments and makes it real. Worth knowing when weighing whether to land this one first.

## What is not here

The other three pieces stay on `claude/bold-brown-sfozgu-combined` until this merges:

| Next | Size |
|---|---|
| Environments report a repeated declaration name | ~700 |
| Remove `remove_shadowed_functions` | ~270 |
| Duplicate variable names, `P4014` (fixes #1525) | ~690 |

## Verification

- `cd compiler && just`, `cd docs && just compile` and `cd specs && just` pass.
- Against the built compiler: the example above → two `P4054`, one on the declaration and one on the call; a function block declaring `p : SomeProgram` and calling `p()` → `P2008` and `P4012`.
- Rule tests cover a function calling a function, a program and function block declaring and invoking instances, a declaration alone, a declaration plus invocation, repeated invocations, `VAR_TEMP`, `VAR_INPUT`, `VAR_IN_OUT` (accepted), a method call on an own instance, and an undeclared instance being `P4012`'s to report rather than this rule's.

https://claude.ai/code/session_01QVGMf972QJsY4bfnUDgqtF


compiler/analyzer/src/rule_pou_hierarchy.rs
```diff
@@ -55,271 +60,313 @@ pub fn apply(
...
+            stateful_instances,
+        });
+        let result = node.recurse_visit(self);
+        self.function = None;
+        result
     }
 
+    // A function block or program may declare and invoke function blocks,
+    // and a function cannot be nested in either, so there is nothing to
+    // find inside them.
```
garretfick: Then should we just remove this function? Yes, this stops early but we are not doing that generally.

compiler/analyzer/src/rule_pou_hierarchy.rs
```diff
@@ -55,271 +60,313 @@ pub fn apply(
...
     fn visit_function_block_declaration(
         &mut self,
-        node: &FunctionBlockDeclaration,
+        _node: &FunctionBlockDeclaration,
     ) -> Result<Self::Value, Infallible> {
```
garretfick: Is this still catching if a function block (or function) tries to call a program? That's also not allowed.

compiler/analyzer/src/rule_pou_hierarchy.rs
garretfick: Removed both overrides in b768bef. You're right that the codebase doesn't stop the walk early elsewhere, and they were also hiding why the rule is correct: the walk finds nothing in a function block or program body because the rule only records instances while it is inside a function, not because it refuses to look. The rule's tests are unchanged.

---

compiler/analyzer/src/rule_pou_hierarchy.rs
garretfick: Still prohibited, but by the pipeline rather than by this rule — and that was true before this PR too, since the rule on `main` never emitted anything but `P4013` for a duplicate name.

A program is not in the type environment, so nothing can name one as a variable type or invoke one:

```
PROGRAM Target
  VAR x : INT; END_VAR
  x := 1;
END_PROGRAM

FUNCTION_BLOCK Caller
  VAR p : Target; END_VAR
  p();
END_FUNCTION_BLOCK
```

```
error[P2008]: Cannot determine kind of type identifier (identifier=Target)
10 │     p : Target;
   │         ^^^^^^ Variable type
error[P4012]: Function block invocation is not a variable in scope (invocation=p)
12 │   p();
   │   ^^^ Function block invocation
```

A bare `Target();` with no declaration is the `P4012` alone. Same from a function, and same from a program — a program cannot invoke a program either; only `PROGRAM ... WITH` in a resource instantiates one.

Nothing tested that, which is a fair reason to doubt it, so b768bef adds pipeline tests through `analyze` for all three POU kinds naming a program as a variable type, for the bare invocation, and for the resource instantiation still being accepted. They live in `stages.rs` rather than in the rule's tests because the behaviour is the pipeline's, not this rule's. The module doc now says so instead of leaving it as an aside.

One thing I did not change: `P2008` reads "cannot determine kind of type identifier", which is accurate but unhelpful here, since `Target` *is* declared, just not as a type. Now that the symbol environment holds programs and configurations, the resolver could say "Target is a program and cannot be used as a type" instead. That's a new diagnostic and a docs page, so I left it out of this PR — say the word and I'll open an issue or fold it in.

---

Issue #1580, opened 2026-08-31
REF_TO a named array type fails codegen with P9999; only the inline array form works


garretfick:
Found triaging plans for #1464. **Ran; reproduces with a control.**

```iecst
TYPE ARR4 : ARRAY[0..3] OF INT; END_TYPE

PROGRAM main
VAR
  arr : ARR4;
  pt : REF_TO ARR4 := REF(arr);
  v : INT;
  i : INT := 1;
END_VAR
  arr[1] := 77;
  v := pt^[i];
END_PROGRAM
```

```
$ ironplcc compile --allow-ref-to --allow-ref-stack-variables -o rn.iplc refnamed.st
error[P9999]: Capability is not implemented (yet!)
   Not implemented at codegen/src/compile_array.rs#L202
```

**Control** — the identical program with the array type spelled inline compiles clean and runs correctly (`v: 77`):

```iecst
  arr : ARRAY[0..3] OF INT;
  pt : REF_TO ARRAY[0..3] OF INT := REF(arr);
```

So dereferenced array subscripting through a `REF_TO` works; it is only the *named type* spelling that fails.

## Cause

`register_ref_to_array_metadata` (`compiler/codegen/src/compile_array.rs:731`) is gated on the target being an inline array:

```rust
if let ReferenceTarget::Array(subranges) = &ref_init.target
```

A `ReferenceTarget::Type` — which is what `REF_TO ARR4` produces — registers no array metadata, so the later subscript lowering finds none and bails at line 202.

## Note on discovery

This is adjacent to, not a regression of, the `deref-array-subscript-local-program` work: that plan only ever described the inline form, and its own deliverable is verified working. The named-type case appears never to have been covered.

Since the two spellings are semantically identical in IEC 61131-3, a user hitting this gets a "not implemented" error for a program that differs from a working one only in where the array type is written.

PR #1735, opened 2026-09-19
Prefactor: resolve REF_TO targets through one path

garretfick:
## Summary

Prefactor for #1580 (related work, not a fix). `pt : REF_TO ARR4` (where `TYPE ARR4 : ARRAY[0..3] OF INT`) still fails codegen with P9999 after this change, exactly as before; the fix lands in a separate PR. This PR reshapes the code so that fix drops in at one place.

`ReferenceTarget` stays a distinct enum (`Named(TypeName) | Array(ArraySubranges)`) because `Named` may be any type, not only an array. Three places each hand-wrote the inline-vs-named branching; each commit below collapses one of them.

## Commits

1. **One resolver in the analyzer** — `TypeEnvironment::resolve_reference_target(declaring, target)` returns the `IntermediateType` a `REF_TO` points at. The body moved verbatim from `fold_reference_declaration`, including the `ParentTypeNotDeclared` diagnostic and the inline `Type`/`Alias` handling. Five new unit tests in `type_environment.rs`.
2. **Remove the dead parser conversion** — `array_specification()` only ever produced `SpecificationKind::Inline`, so the `Named` arm of the `match` in `ref_to_target` was dead. New `array_subranges()` rule; `array_specification()` wraps it and `ref_to_target` uses it directly.
3. **One reference registration path in codegen** — the `InitialValueAssignmentKind::Reference` arm was copied four times. New `compile_reference.rs` holds `register_reference_variable` (called from all four sites) and `register_ref_to_array_metadata`. The dimension/stride loops it duplicated from `register_array_variable` are now one shared `compute_dimensions`. The `ReferenceTarget::Array` gate stays, with a comment pointing at #1580.

## Behaviour

- Every existing test passes unchanged. No test file outside the new unit tests in commit 1 was added or edited.
- The one observable change: an oversized `REF_TO ARRAY` now reports the same not-implemented diagnostic as a plain array instead of overflowing the element count (noted in the commit 3 message).
- The hard-coded `is_string_element: false` in the reference path is left as is; it is a separate latent issue.
- The bug still reproduces: `TYPE ARR4 : ARRAY[0..3] OF INT; END_TYPE` with `pt : REF_TO ARR4 := REF(arr)` and `v := pt^[i]` fails with P9999 under `--allow-ref-to --allow-ref-stack-variables`, while `REF_TO ARRAY[0..3] OF INT` compiles.

## Not done

- The `Reference` arm of the array-element-type match in `xform_resolve_expr_types.rs` was left alone. It answers a different question (an element type *name* for the subscript resolver, returning `None` for a non-array named type rather than an error) and has no declaring type name to pass, so routing it through the new method would not be behaviour-preserving.
- `type_environment.rs` was already over the 1000-line module guideline before this change (1164 lines); the new method and tests take it to 1319. Splitting it is out of scope for a prefactor and would be its own change.

## Verification

- `cd compiler && just` passes (compile, coverage at 85%, clippy, fmt, dupes).
- `cd specs && just` passed while the plan file existed. The plan was `git rm`'d in the final commit.

https://claude.ai/code/session_01JXC6LtyzWX3RKXHRSRqEUQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01JXC6LtyzWX3RKXHRSRqEUQ)_


compiler/analyzer/src/type_environment.rs
```diff
@@ -1161,4 +1203,117 @@ mod tests {
...
+
+        assert!(result.is_err());
+        assert_eq!(
+            result.unwrap_err().code,
+            Problem::ParentTypeNotDeclared.code()
+        );
+    }
+
+    #[test]
+    fn resolve_reference_target_when_inline_array_element_undeclared_then_error() {
```
garretfick, line 1309: Isn't this the same as the test above?

garretfick: Same setup, different target kind, and the inline one was too weak to show the difference because it never asserted the problem code. Merged them into one `rstest` with two cases, each asserting its code: an undeclared named target is `ParentTypeNotDeclared` (P2011), an undeclared inline element type is `ArrayElementTypeNotDeclared` (P2013). Pushed in 6b3c4a5.

---

garretfick: Correction: the commit is 5b2d04a, not 6b3c4a5.

---

Issue #1734, opened 2026-09-17
Codegen P9999 from `Diagnostic::todo()` has no source location, and `check` never reports codegen problems


Yichi-Jin:
## Summary

A program that `ironplcc check` accepts can still fail `ironplcc compile` with P9999. At 15 codegen sites, that P9999 has no source location: it renders as `┌─ :1:1` with an empty file name and an empty source line, so neither a person nor a tool can tell which construct is not supported.

## Reproduction

`ironplcc` built from source at 98da8c99 (0.244.0), macOS 26.6 arm64.

```st
PROGRAM main
VAR
    level : REAL;
    alarm : BOOL;
END_VAR
    CASE level OF
        1: alarm := TRUE;
    END_CASE;
END_PROGRAM
```

```
$ ironplcc check case_real.st        # exit 0, no output
$ ironplcc compile case_real.st -o case_real.iplc
error[P9999]: Capability is not implemented (yet!)
  ┌─ :1:1
  │
1 │
  │ ^ Not implemented at codegen/src/compile_stmt.rs#L784
  │
  = Learn more: https://www.ironplc.com/reference/compiler/problems/P9999.html?version=0.244.0&channel=cli&file=codegen/src/compile_stmt.rs&line=784

Error: "Compile failed with 1 problem(s)"
```

For comparison, a P9999 site that passes a span points at the declaration (path shortened):

```st
TYPE Mode : (Idle, Fill, Drain); END_TYPE
PROGRAM main
VAR
    schedule : ARRAY[0..2] OF Mode := [Fill, Idle, Drain];
    step : INT;
END_VAR
    step := step + 1;
END_PROGRAM
```

```
$ ironplcc compile enum_array.st -o enum_array.iplc
error[P9999]: Capability is not implemented (yet!)
  ┌─ enum_array.st:4:5
  │
4 │     schedule : ARRAY[0..2] OF Mode := [Fill, Idle, Drain];
  │     ^^^^^^^^ Unsupported array element type
```

## Cause

[`Diagnostic::todo()`](https://github.com/ironplc/ironplc/blob/98da8c993dc948a211ebc1b34c498df00ea25d60/compiler/dsl/src/diagnostic.rs#L177-L184) labels the problem with `SourceSpan::default()`. These sites in `compiler/codegen/src` still call it (line numbers at 98da8c99):

- `compile_expr.rs`: L38 and L98–L99 (`op_type` and `storage_bits` when `resolved_type` is missing), L371
[...]

PR #1750, opened 2026-09-21
Diagnose a CASE selector that is not an integer or enumeration

garretfick:
## Summary

IEC 61131-3 defines the `CASE` selector as an expression of type `ANY_INT` or an enumeration. Nothing checked that, so `CASE level OF` with `level : REAL` passed `ironplcc check` and failed only in codegen, as a P9999 with no source location. This is the REAL selector part of #1734; the other two parts of that issue (spans on every `Diagnostic::todo()` site, and whether `check` should run codegen) are not addressed here.

Follows the codegen prefactor merged in #1743.

## Changes

**Prefactor (first commit).** The expression-to-representation lookup that `rule_constant_range` did privately moves onto `TypeEnvironment` as `representation_of_expr`, so both rules ask the same question the same way. No behaviour change.

**New rule `rule_case_selector_type`, P4053.** Reports against the selector expression when its resolved type has a representation in the type environment that is not an integer, a subrange of one, or an enumeration. (P4052 was taken by #1746 while this was in flight, so the code is P4053.)

- Rejected: `REAL`, `LREAL`, `BOOL`, the bit strings `BYTE`/`WORD`/`DWORD`/`LWORD`, strings, times, dates, and aggregates.
- Bit strings are rejected deliberately even though codegen compiled a `WORD` selector before. The standard does not list them, and a program that compiled but should not have is what the rule exists to catch. The bit-string *label* extension (`--allow-bit-string-case-labels`) is unaffected: it pairs a radix-prefixed label with an integer selector.
- Skipped: a selector whose type is unresolved, or resolves to a generic category that is not in the type environment (a bare literal, `CASE 1 OF`, is `ANY_INT`).

**Codegen.** With analysis in front of it, the float arm added in #1740 is a broken invariant rather than a missing capability, so `non_integer_case_selector` now builds `internal_error_at` (P9998) at the selector. The two `compile_not_implemented_location` cases that exercised it move to `compile_case.rs` and assert the internal error at the same span. Those tests resolve types without the semantic rules, which is the only way to reach the arm.

**Docs.** New `P4053.rst` with the `REAL` and `WORD` examples and their fixes. `case.rst` said the selector is an integer expression and omitted enumerated values as labels; it now says both and links P4053.

## Behaviour

```
$ ironplcc check case_real.st
error[P4053]: CASE selector is not an integer or enumeration type (actual=real)
  ┌─ case_real.st:6:10
  │
6 │     CASE level OF
  │          ^^^^^ CASE selector
```

This is a breaking change for any program that selected on a bit-string variable. Such a program compiled before but was never valid IEC 61131-3.

## Testing

- `cd compiler && just` passes (compile, coverage, clippy, fmt).
- Rule tests cover every integer type, a named subrange, an enumeration, an alias of an integer, an integer expression, and a bare literal (accepted); `REAL`, `LREAL`, `BOOL`, every bit string, `STRING`, `TIME`, `DATE`, an alias of `REAL`, and a real expression (rejected); plus the label span and the `actual=` context.
- Docs build with no warnings and the duplicate-content check passes.

https://claude.ai/code/session_01QBZDUEuPX6oktNBEkcUA2p


compiler/analyzer/src/rule_case_selector_type.rs
```diff
@@ -0,0 +1,346 @@
...
+    #[rstest]
+    #[case::real("REAL")]
+    #[case::lreal("LREAL")]
+    #[case::bool("BOOL")]
+    #[case::byte("BYTE")]
+    #[case::word("WORD")]
+    #[case::dword("DWORD")]
+    #[case::lword("LWORD")]
+    #[case::string("STRING")]
+    #[case::time("TIME")]
```
garretfick, line 203: We are missing LTIME and other date-time items.

garretfick: Added. The rejected list now has every elementary type that is not an integer: `LTIME`, `LDATE`, `TIME_OF_DAY`, `LTIME_OF_DAY`, `DATE_AND_TIME`, `LDATE_AND_TIME` and `WSTRING` alongside the ones already there. All pass with no rule change, since the rule accepts by representation (`Int`, `UInt`, `Subrange`, `Enumeration`) rather than by naming rejected types.

---

PR #1756, opened 2026-09-21
Environments report a repeated declaration name

### LINKED ISSUE #1525: Redeclaring __SYSTEM_UP_TIME silently shadows it and reads zero forever

garretfick:
Second of four, after #1728. Cut fresh from `main` and stands alone.

## Why

`FunctionEnvironment::insert` already returned `P4016` for a repeated function and `TypeEnvironment::insert_type` returned `P2007` for a repeated type. Neither ever fired: `xform_toposort_declarations` kept one declaration per name, so a repeat never reached an environment.

So every duplicate-name problem page documented behaviour the pipeline did not have. The example on the `P4013` page — two `FUNCTION_BLOCK MyFunction` declarations, commented `(* Error: Duplicate POU name *)` — compiled clean on `main`, as did its cross-kind example. Both now produce the error the page claims.

This also closes the gap #1728 opened: `P4013` went unemitted when the POU hierarchy rule was replaced, having been unreachable there too. It is reachable again here.

## What changed

The toposort keeps every declaration, and each environment reports the repeats for the kinds it holds. Ordering is unaffected — the dependency graph is keyed by name either way.

- `TypeEnvironment::insert_type` records a repeated data type, function block or interface (`P2007`, or `P4013` when a function block is involved).
- `SymbolEnvironment::insert` reports the pairs the type environment never sees, those involving a program or a configuration (`P4013`).
- `FunctionEnvironment::insert` keeps reporting `P4016`.
- The transform feeding the symbol and function environments checks a function against a symbol of the same name, since those live in different environments.

| Later declaration repeats | Code |
|---|---|
| function repeating a function | P4016 |
| data type repeating a data type | P2007 |
| anything else (function block, program, configuration, interface, or a cross-kind clash) | P4013 |

The primary label is on the later declaration and a secondary label names the first. Both passes run best-effort: a repeat is collected as a diagnostic and analysis continues on the first declaration, rather than the whole library reverting. The late-bound initializer pass, which keeps its own name table, keeps the first declaration silently so a repeat is reported once.

## What is not here

`remove_shadowed_functions` still drops a library `FUNCTION` that user source redeclares, so a user declaration named like an activated library's still wins silently. That is the next PR, which is why the `REQ-CL-analyzer-004` conformance tests are untouched and still pass here. The `P4014` variable work is the fourth.

| Next | Size |
|---|---|
| Remove `remove_shadowed_functions` | ~270 |
| Duplicate variable names, `P4014` (fixes #1525) | ~690 |

## Verification

- `cd compiler && just`, `cd docs && just compile` and `cd specs && just` pass. The whole workspace suite is green with no test changed outside the files above.
- Against the built compiler, each reported exactly once: duplicate function → `P4016`; duplicate data type → `P2007`; duplicate function block, duplicate program, and a `TYPE` clashing with a `FUNCTION_BLOCK` → `P4013`; a `FUNCTION` clashing with a `FUNCTION_BLOCK` → `P4013`.
- Both `playground::` examples on the `P4013` page were run against the built compiler and now produce the documented error.

https://claude.ai/code/session_01QVGMf972QJsY4bfnUDgqtF

---
_Generated by [Claude Code](https://claude.ai/code/session_01QVGMf972QJsY4bfnUDgqtF)_


compiler/analyzer/src/xform_resolve_type_decl_environment.rs
```diff
@@ -162,11 +171,11 @@ impl Fold<Diagnostic> for TypeEnvironment {
                 }
             }
             InitialValueAssignmentKind::String(string_initializer) => {
-                self.insert_type(&node.type_name, string::from(string_initializer))?;
+                self.insert_type(&node.type_name, string::from(string_initializer));
```
garretfick, line 174: Isn't this ignoring a possible error?

garretfick: No — `insert_type` returns `()`, so there is no result being dropped here. A repeat is recorded inside the environment and drained later, which is what makes the pass best-effort: analysis continues on the first declaration instead of the whole library reverting.

The chain is `insert_type` → `self.duplicates` → `take_duplicates()` at `xform_resolve_type_decl_environment.rs:37`, whose diagnostics `stages.rs` collects.

I checked rather than assuming, since this arm and the `EnumeratedValues` arm below it are the two that discard the call:

```
TYPE
  Name : STRING[10];
  Name : STRING[20];
END_TYPE
```
→ `error[P2007]: Type declaration name is duplicated (name=Name)`, and the enumeration equivalent likewise.

I also audited every caller: all production `insert_type` calls are in this file, under the `apply` that drains. The three in `intermediates/array.rs` and `intermediates/subrange.rs` are inside `#[cfg(test)]` modules with their own environments.

What I think prompted the question is real, though: two lines above, `insert_alias(...)?` propagates, so the bare `insert_type(...)` reads like a dropped result. The asymmetry is deliberate — an alias to a missing base type cannot continue, a duplicate name can — and `insert_type`'s doc comment says so, but nothing at the call site does. Happy to add a short comment on the discarding arms if you'd like it visible there; I left it out rather than add noise to seven call sites.

---

Issue #1541, opened 2026-08-30
Enum explicit values and the base-type suffix are accepted under the strict Edition 2 dialect with no diagnostic


garretfick:
Found while triaging `2026-07-19-twincat-explicit-enum-values` (#1464). Reproduced.

```iecst
TYPE E_AssertionType : (Type_UNDEFINED := 0, Type_ANY, Type_BOOL) BYTE; END_TYPE
PROGRAM main
VAR
  x : E_AssertionType;
  result : DINT;
END_VAR
  x := Type_BOOL;
  IF x = Type_BOOL THEN result := 42; END_IF;
END_PROGRAM
```

```
$ ironplcc check enum_ext.st
$ echo $?
0
```

No flags, no dialect — so the default, which `ironplcc compile --help` advertises as *"Strict IEC 61131-3:2003 (Edition 2). **No extensions.**"* Two non-Edition-2 constructs pass silently:

- the `) BYTE;` **base-type suffix**, which the plan itself quotes Beckhoff describing as an *"extension beyond IEC 61131-3"*;
- **explicit enumerator values**, which are Edition 3, not Edition 2.

`specs/steering/syntax-support-guide.md:22` requires: *"**Non-standard gating**: If not standard IEC 61131-3, gate behind `--allow-x` flag"*. Neither construct has a flag — `grep` for an enum-related `allow_` in `compiler/parser/src/options.rs` finds nothing, and `docs/explanation/enabling-dialects-and-features.rst` lists a flag for every sibling extension (paren string length, bit-string case labels, struct initializer expressions, mixed located vars) but none for this.

## The plan's stated reason is not true

The plan justified shipping ungated:

> the parser has no mechanism to gate a non-keyword grammar change by dialect or edition at all… exactly matching the precedent already established for `STRING(n)` parentheses.

`STRING(n)` is the counter-example, not the precedent. `compiler/parser/src/rule_token_no_paren_string_length.rs:9`:

> The grammar accepts the parenthesis form **unconditionally** … so this **token-stream check is what enforces the flag**.

That is exactly a mechanism for gating a non-keyword grammar change: parse permissively, reject post-lex against `CompilerOptions` with a `P40xx` code. It is also what ADR-0040 prescribes, and what `--allow-bit-string-case-labels` (P4041), `--allow-struct-initializer-expressions` (P4043) and `--allow-constant-initializer-expressions` (P4037) all do.

So the gap is a decision made on a false premise, and the machinery to close it already exists and has four users.

[...]

PR #1765, opened 2026-09-21
Gate explicit enum member values behind --allow-enum-explicit-values

garretfick:
Strict Edition 2 — the default dialect — accepted an explicit per-member enumeration value:

```iecst
TYPE E_ModeLanguage : (Deutsch := 1, English := 2); END_TYPE
```

Explicit enumerator values are an IEC 61131-3:2013 (Edition 3) addition, not Edition 2, so the strict Edition 2 dialect has to reject them.

Adds the `allow_enum_explicit_values` flag — enabled by `iec61131-3-ed3`, `rusty`, `codesys` and `twincat` — and a new analyzer rule reporting **P4055** for every enumeration member declared with an explicit value when the flag is off.

```
$ ironplcc check enum_ext.st
error[P4055]: Explicit enumeration member value requires choosing a dialect or --allow-enum-explicit-values flag
1 │ TYPE E_AssertionType : (Type_UNDEFINED := 0, Type_ANY, Type_BOOL) BYTE; END_TYPE
  │                         ^^^^^^^^^^^^^^ Enumeration member with an explicit value
```

## Why an analyzer rule

Per ADR-0040 the grammar stays maximal and option-free. `:=` is the same token in a dozen unrelated positions, so there is no distinguishing token to demote or reject pre-parse; the illegality is visible only in the parsed structure, which ADR-0040 rule 3 sends to a post-parse check.

The AST already carries it faithfully: `EnumeratedValue::explicit_value` holds the value the user wrote, and only the declaration grammar path sets it. A default (`:= GREEN`), a `CASE` label and an expression operand all leave it `None`, so a plain visit cannot mistake one of those for the extension — each has a test.

## Documentation

The `enumerated-types` reference page never documented explicit values at all, and claimed values must be unique when only names are checked. Both fixed, and its examples are runnable playgrounds. The base-type suffix is now described as the vendor extension it is.

## Out of scope

Issue #1541 also reports the base-type suffix (`(A, B) BYTE;`) as ungated. That is a vendor extension rather than Edition 3 syntax, so it needs its own flag with a different dialect set. It stays open on the issue.

Refs #1541

https://claude.ai/code/session_01418gGs5aXLk55PJveQPdhM


docs/reference/language/data-types/derived/enumerated-types.rst
```diff
@@ -40,19 +40,53 @@ Example
...
+values starting at zero, so ``Red`` is 0, ``Yellow`` is 1 and ``Green`` is 2.
+
+Explicit Values
+---------------
+
+A member can be given its own value instead of the one its position implies.
+Members that follow continue from the value before them, so ``Type_ANY`` below
+is 1 and ``Type_BOOL`` is 2:
+
+.. code-block::
```
garretfick: This should be a playground example.

garretfick: Done in f96331e. It's now a `.. playground::` wrapping a complete program, with `:allows: enum-explicit-values` so the example states the flag it needs (the playground's default `rusty` dialect already enables it, but it reads better stated).

I applied the same treatment to the `Base Type` example further down — it was the other static code block this PR added to the page, so leaving it would have made the page inconsistent. Both compile and run clean under `rusty`, `codesys`, `twincat` and `iec61131-3-ed3`; docs CI passes with no new jscpd clones.

---

Issue #1560, opened 2026-08-31
DATE, LDATE, DT and LDT literals past 2106-02-07 overflow u32 in the compiler


garretfick:
Found while correcting the LDATE/LTOD/LDT unit claims (#1551) — the range column of the amended ADR-0025 table would not have been true as written. Reproduced against `target/debug/ironplcc`.

`DateLiteral::seconds_since_epoch` and `DateAndTimeLiteral::seconds_since_epoch` return `u32` and compute the second count in `u32` before codegen widens it:

```rust
// compiler/dsl/src/time.rs:207
pub fn seconds_since_epoch(&self) -> u32 {
    const UNIX_EPOCH_JULIAN_DAY: i32 = 2_440_588; // 1970-01-01
    let julian_day = self.value.to_julian_day();
    let days = (julian_day - UNIX_EPOCH_JULIAN_DAY) as u32;
    days * 86_400
}
```

`u32::MAX / 86_400` is 49,710 days, so the last representable date is 1970-01-01 + 49,710 days = **2106-02-07**. Nothing rejects a later literal.

## Reproduction

```iecst
PROGRAM main
VAR
  d : LDATE := LDATE#2200-01-01;
END_VAR
END_PROGRAM
```

```
$ ironplcc compile --dialect iec61131-3-ed3 --output out.iplc ldate.st
thread 'main' (1637) panicked at dsl/src/time.rs:207:9:
attempt to multiply with overflow
   3: ironplc_dsl::time::DateLiteral::seconds_since_epoch
             at ./dsl/src/time.rs:207:9
   4: compile_constant
             at ./codegen/src/compile_expr.rs:454:28
```

The boundary is exact:

| Literal | Result |
|---|---|
| `LDATE#2106-02-07` | compiles |
| `LDATE#2106-02-08` | panics |

The same panic reaches `DateAndTimeLiteral::seconds_since_epoch` at `dsl/src/time.rs:252` via `LDT#2200-01-01-00:00:00`, and `DATE#2200-01-01` on the 32-bit type. `LTOD` is unaffected — `whole_milliseconds()` is bounded by 86,399,999 by construction.

Two distinct problems, worth separating:

## 1. Any out-of-range date literal panics instead of diagnosing

[...]

PR #1767, opened 2026-09-22
Validate date literals are within representable range

garretfick:
## Summary

Add semantic validation to ensure date and date-and-time literals name dates that can be represented as unsigned 32-bit second counts since the Unix epoch (1970-01-01). Dates outside the range 1970-01-01 to 2106-02-07 (or 2106-02-07-06:28:15 for date-and-time) are now reported as error P2038 rather than silently wrapping to incorrect dates.

## Changes

- **New semantic rule** (`rule_date_literal_range`): Validates all date and date-and-time literals against the representable range. Reports every out-of-range literal with context about the valid bounds.

- **Codegen safety**: Updated `compile_epoch_seconds` to validate seconds fit in u32 before emission, providing a safety net for callers that skip semantic analysis (e.g., tests, tools).

- **Type signature widening**: Changed `DateLiteral::seconds_since_epoch()` and `DateAndTimeLiteral::seconds_since_epoch()` return type from `u32` to `i64` to allow negative values (dates before epoch) and values beyond u32::MAX (dates after 2106-02-07) to reach validation logic rather than truncating silently.

- **Documentation**: Added problem code P2038 with detailed explanation of the date range, why it exists, and how to work around it. Updated DATE, LDATE, DATE_AND_TIME, and LDATE_AND_TIME type reference pages to document the range and link to P2038.

- **Tests**: Comprehensive test coverage including boundary conditions (epoch, last representable date), both literal types, all positions a literal can appear (initializer, assignment, comparison, global constant), and verification that all violations are reported.

## Implementation Details

- The 64-bit types (LDATE, LDATE_AND_TIME) are held to the same range as their 32-bit counterparts because they store the same second count (per ADR-0025 amendment). This is documented as issue #1560 for future resolution.

- The range bounds are expressed without type prefixes in error messages because DATE# and LDATE# parse to the same literal, as do DT# and LDT#.

- Shared helper `seconds_to_midnight()` ensures date and date-and-time literals agree on epoch location and day duration.

https://claude.ai/code/session_01Xp7WYkgFE6JLWfkamtHTkR


compiler/codegen/src/compile_expr.rs
```diff
@@ -251,6 +251,45 @@ fn compile_compare(
...
+            Label::span(span, "Date literal"),
+        )
+        .with_context("seconds since 1970-01-01", &seconds.to_string())
+    })?;
+    match op_type.0 {
+        OpWidth::W64 => {
+            let pool_index = ctx.add_i64_constant(i64::from(stored));
+            emitter.emit_load_const_i64(pool_index);
+        }
+        _ => {
```
garretfick: Can we have exhaustive matching instead of catch all?

garretfick: Done in 92c57ad — and making it exhaustive turned out to find something.

The catch-all was inherited from the two date arms it replaced, and it was wrong for two of the four widths. `r : REAL; r := D#2024-01-01;` compiled to:

```
LOAD_CONST_I32 pool:0
STORE_VAR_F32  var:0
```

The second count's bit pattern went into a float slot, to be read back as a number unrelated to the date. Silent, and it predates this PR.

A date is a count rather than a measurement, so no floating-point type holds one, and the analyzer already rejects the assignment that would ask for it (P4035 — confirmed with `ironplcc check`). Reaching codegen at a float width is therefore a broken invariant rather than a missing capability, so `F32 | F64` now returns an internal error at the literal, the same way the CASE selector's float arm became one in #1750. `W32` and `W64` spell out their own arms. Two `rstest` cases (REAL and LREAL) cover it through `try_parse_and_compile`, which reaches the arm because the codegen tests resolve types without the semantic rules.

One thing I left alone: the sibling `Duration` and `TimeOfDay` arms in the same `compile_constant` still use `_ =>`, and look like the same latent issue. I kept it out of this PR rather than widening the diff — happy to open an issue for it, or fold it in here if you would rather.

---

PR #1769, opened 2026-09-22
Gate the enum base-type suffix behind --allow-enum-base-type

### LINKED ISSUE #1541: Enum explicit values and the base-type suffix are accepted under the strict Edition 2 dialect with no diagnostic

garretfick:
Follow-on to #1765, which gated the explicit-values half of #1541. This gates the other half.

Strict Edition 2 — the default dialect — accepted the base-type suffix on an enumeration declaration:

```iecst
TYPE E_Small : (A, B) WORD; END_TYPE
```

The suffix names the elementary type the members are stored in, overriding the automatic count/value-based sizing in `intermediates/enumeration.rs`. IEC 61131-3 has no such form — an `enumerated_specification` (Annex B) is a list of member names or another enumerated type's name, and nothing may follow it — so it is a CODESYS/TwinCAT extension, listed as one in `beckhoff-twincat-dialect.md` §3.10 under "Enum with Underlying Type".

Adds the `allow_enum_base_type` flag and an analyzer rule reporting **P4056**:

```
$ ironplcc check base.st
error[P4056]: Enumeration base-type suffix requires choosing a dialect or --allow-enum-base-type flag (base_type=WORD)
1 │ TYPE E_Small : (A, B) WORD; END_TYPE
  │      ^^^^^^^ Enumeration declared with a base-type suffix
```

## Not an Edition 3 flag

The one thing worth a close look. `allow_enum_base_type` is enabled by `rusty`, `codesys` and `twincat` — **not** `iec61131-3-ed3`, which is where it differs from its sibling `--allow-enum-explicit-values`. Explicit values are standard as of Edition 3; the suffix is a vendor extension in every edition. The two travel together in vendor code, which is exactly why they are separate flags rather than one.

Verified: `iec61131-3-ed3` accepts explicit values and rejects the suffix.

## Why an analyzer rule

Same mechanism as #1765, deliberately: two flags over one language construct should read as a pair.

Per ADR-0040 the grammar stays maximal and option-free. The elementary type keyword that spells the suffix (`BYTE`, `WORD`, `INT`, …) is ordinary standard syntax everywhere else, so there is nothing to demote; a token rule would have to key on the `)`-then-type-keyword adjacency and would fire on malformed input that is no enum declaration at all. The parser already records the suffix faithfully in `EnumeratedSpecificationInit::underlying_type`, so the check is a visitor over the parsed tree.

## Known limitation: the caret

The label points at the declared type name, not at the suffix, because `ElementaryTypeName` is a fieldless enum carrying no `SourceSpan`. Giving the suffix its own span means threading one through the parser, the renderer, both analyzer transforms, the XML transform and codegen — larger than this change, so it is filed separately rather than folded in. The declaration is one line, so the caret is close, and the message names the base type it found.

## Validation

`cd compiler && just` (compile, coverage, clippy, fmt, `cargo dupes`), `cd docs && just ci` and `cd specs && just` all pass on the rebased branch. No `.st` or TwinCAT resource in the repo uses the suffix, so nothing needed a dialect bump.

Closes #1541

https://claude.ai/code/session_01418gGs5aXLk55PJveQPdhM

---
_Generated by [Claude Code](https://claude.ai/code/session_01418gGs5aXLk55PJveQPdhM)_


docs/reference/language/data-types/derived/enumerated-types.rst
```diff
@@ -98,16 +99,22 @@ A declaration can name the elementary type the members are stored in:
    END_PROGRAM
 
 Without it, IronPLC picks the smallest type that holds every member's value.
-The suffix is a CODESYS/TwinCAT extension beyond IEC 61131-3 rather than
-standard syntax; IronPLC currently accepts it under every dialect, including
-strict Edition 2.
+
+The suffix is a CODESYS/TwinCAT extension beyond IEC 61131-3, so the default
```
garretfick: We don't mention specific extensions here. We just refer to how to enable features.

docs/reference/compiler/ironplcc.rst
```diff
@@ -292,6 +292,18 @@ Options
    flag. Enabled by ``--dialect=iec61131-3-ed3``, ``--dialect=rusty``,
    ``--dialect=codesys``, and ``--dialect=twincat``.
 
+``--allow-enum-base-type``
+   Allow the base-type suffix on an enumeration declaration (e.g.
+   ``(A, B) WORD``), naming the elementary type the members are stored in.
+   IEC 61131-3 has no such form — nothing may follow an enumerated
+   specification — so the suffix is a CODESYS/TwinCAT extension. Without it
```
garretfick: Don't mention dialects here. Let the dialect define what it enables.

docs/reference/language/data-types/derived/enumerated-types.rst
garretfick: Done in e7069a8. Both sections now use the house includes instead of their own prose — `requires-edition3.rst` for Explicit Values, `requires-dialect-extension.rst` for Base Type — and the base-type heading picked up the `(Language Extension)` suffix that `array-types` and `subrange-types` use for the same situation. No dialect, vendor or flag names left in the prose; the problem codes stay under `Related Problem Codes`, where the page already points at P2003.

I applied it to the Explicit Values section too, not just the one you flagged. That section came in with #1765 and had the same defect sitting directly above this one, so fixing only the lower half would have left the page inconsistent.

---

docs/reference/compiler/ironplcc.rst
garretfick: Done in 8d8f921. Both enum flag entries now say only what syntax they gate, which edition standardized it (or that none did), and the problem code — the dialect mapping is left to the per-dialect `**Enables:**` lists.

You said you were still seeing dialects in the allow list, so I swept the rest rather than just this line:

- `ironplcc.rst` — both `--allow-enum-explicit-values` and `--allow-enum-base-type`
- `enabling-dialects-and-features.rst` — the same two flag entries (the `**Enables:**` lists are untouched; they're the canonical mapping and the build check enforces them)
- `P4055.rst` and `P4056.rst` — dropped "CODESYS, TwinCAT and RuSTy" and the `--dialect=iec61131-3-ed3` reference
- `main.rs` — the `--allow-enum-explicit-values` clap help, which is this same list seen from the terminal

The `--allow-enum-explicit-values` half of that came in with #1765. Same defect, adjacent lines, so I fixed it here instead of leaving it — happy to split it out if you'd rather this PR stayed narrow.

Worth knowing: `--allow-fb-inheritance` still ends with "Enabled by `--dialect=iec61131-3-ed3`, …" in both files, and its clap help does the same. I left it alone as unrelated to this PR — say the word and I'll clean it up separately.

`cd compiler && just`, `cd docs && just ci` and `cd specs && just` all pass.

---

